PR #7, opened 2026-03-27
feat: add P-256 key support, signing oracle endpoint, and cache api

marat-affinidi:
# Problem

  The VTA currently supports only Ed25519/X25519 key types and has no "sign without export" operation. This limits its usefulness as a key custody service — any consumer that needs ECDSA signatures (ES256) must manage
  P-256 keys outside the VTA, defeating the purpose of centralized key management.

  Common scenarios blocked today:

   - JWT signing with ES256 — the dominant algorithm in OAuth
    2.0, OpenID Connect, and most cloud IAM systems. Without P-256 support, consumers must store private keys locally, creating the exact security gap the VTA is designed to eliminate.
   - Token lifecycle management — consumers that authenticate against external services need a secure, scoped place to cache short-lived tokens. Without this, tokens end up in plaintext files or environment variables.
   - Agent and automation workflows — headless systems (CI runners, background services, SDK integrations) need to sign payloads without interactive key export. A signing oracle model ("send payload, get signature") keeps 
  key material in a single hardened boundary.

# Solution

  P-256 key support — New P256 variant in KeyType. Keys are derived via HMAC-SHA512 domain separation from the existing BIP-32 seed, producing cryptographically independent key material that avoids cross-curve reuse,
  Ed25519 clamping artifacts, and group-order bias.

  Signing oracle endpoint — POST /keys/{key_id}/sign accepts a base64url payload and algorithm (eddsa or es256), derives the key in memory, signs, and returns the signature. Key material is dropped immediately after
  signing and never leaves VTA.

  Token cache API — GET/PUT/DELETE /cache/{key} provides a TTL-scoped key-value store per authenticated caller. Designed for caching short-lived bearer tokens, PATs, or session artifacts without disk exposure.

- Add P256 variant to KeyType enum (vta-sdk)
- Add sign-request/sign-result protocol types with SignAlgorithm enum (vta-sdk)
- Add derive_p256() to Bip32Extension trait using p256 crate (vta-service)
- Add POST /keys/{key_id}/sign REST endpoint for signing payloads (vta-service)
- Add sign() method to VtaClient (vta-sdk)
- Add GET/PUT/DELETE /cache/{key} REST API for token caching (vta-service)
- Add cache keyspace to AppState
- Update create_key, get_key_secret, keys_cli to handle P-256 keys
- Add P-256 derivation and sign/verify tests

### DIFF
--- a/docs/bip32_paths.md
+++ b/docs/bip32_paths.md
@@ -43,6 +43,22 @@ keyspace under the key `path_counter:{base_path}`. Every key allocation:
 All key types within a context (signing, key-agreement, pre-rotation) share
 **one counter**, so indices are unique and never reused.
 
+### P-256 Domain-Separated Derivation
+
+P-256 keys share the same BIP-32 path namespace as Ed25519/X25519 but use
+**HMAC-SHA512 domain separation** to produce independent key material:
+
+1. Derive the BIP-32 path normally (SLIP-0010 for Ed25519).
+2. Compute `HMAC-SHA512(key="p256-key-derivation", data=signing_key_bytes || chain_code)`.
+3. Take the first 32 bytes as the P-256 scalar (reduced mod n automatically).
+
+This ensures:
+- **No cross-curve key reuse** — Ed25519 and P-256 keys at the same path are
+  cryptographically independent.
+- **No Ed25519 clamping artifacts** — the HMAC output is uniformly distributed.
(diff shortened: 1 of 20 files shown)


marat-affinidi:
Superseded by new PR from cleanly rebased branch (feature/p256-signing-oracle-rebase).

PR #33, opened 2026-04-24
feat: did:key VTA identity, rate limiter fix, PNM URL persistence

marat-affinidi:
## Summary

Adds `did:key`-based VTA identity support for localhost/development VTAs that don't need a `did:webvh` hosted on a public domain.

### 1. `fix(vta): add ConnectInfo<SocketAddr> to axum::serve for rate limiter`
The `axum::serve()` call was missing `.into_make_service_with_connect_info::<SocketAddr>()`, causing the `ConnectInfo<SocketAddr>` extractor (used by the rate-limiter middleware) to panic at runtime. Now wired correctly.

### 2. `feat(vta): add create_did_key VTA identity option`
New `create_vta_did_key()` path in TOML-based setup that generates a `did:key` identity from the VTA's BIP-32 seed. Stores Ed25519 (signing) and X25519 (key-agreement) key records. `init_auth()` branches on `did:key:` prefix to use `secrets_from_did_key()` which handles multibase fragment IDs and curve-converted X25519.

### 3. `fix(cli): add P256 support to keys create command`
`pnm keys create --key-type p256` was failing with "unknown key type" even though the VTA server fully supports P256. Root cause: when P256 was added to the server (v0.2.0) and to `cmd_key_import` (v0.3.0), the `cmd_key_create` match arm was only refactored for builder pattern — P256 was never added. Fix adds the missing match arm and updates help text + README for both pnm-cli and cnm-cli.

### 4. `fix: DIDComm fragment-ID mismatch for did:key VTAs`
**Addresses review findings (🔴 merge-blocker + 🟡 + 🟢):**

**🔴 DIDComm fragment-ID mismatch:** `did:key` uses multibase fragment IDs (`{did}#{ed_pub_mb}`) while `did:webvh` uses `#key-0/#key-1`. Two consumer sites hardcoded the webvh convention:
- `server.rs` DIDComm service init → `get_secret("#key-0")` returned None for did:key → TDKProfile got zero secrets
- `status.rs` trust-ping → independently derived X25519 instead of curve-converting from Ed25519

Fix introduces `AuthInit` struct to carry verification-method IDs from `init_auth()` through to consumers, and branches `send_trust_ping()` on DID method.

**🟢 Restored operator-facing log prose** on SIGNING KEY MISMATCH / KEY-AGREEMENT KEY MISMATCH messages.

### 5. `fix(pnm): persist VTA URL for did:key DIDs`
The upstream URL refactor (derive from DID document) breaks `did:key` VTAs — `did:key` has no service endpoints, so `resolve_vta_url()` fails. Fix re-adds `VtaConfig.url` as an optional field used only for DIDs that cannot advertise endpoints:
- Interactive setup prompts for URL when VTA DID is `did:key:`
- Non-interactive requires `--vta-url` flag for `did:key:`
- Client construction uses persisted URL as fallback before DID resolution
- For `did:webvh`, the field stays None (URL derived at runtime)

## Testing
- `cargo test -p vta-service` — 34 tests pass
- `cargo test -p pnm-cli` — 36 tests pass
- Manual: `pnm keys create --key-type p256` works
- Manual: VTA starts with `did:key` identity, `pnm health` reaches it via persisted URL

### DIFF
--- a/cnm-cli/README.md
+++ b/cnm-cli/README.md
@@ -263,7 +263,7 @@ cnm config update --community-vta-name "My VTA" --public-url "https://vta.exampl
 | Command                                                                    | Description     |
 | -------------------------------------------------------------------------- | --------------- |
 | `keys list [--status active\|revoked] [--limit N] [--offset N]`            | List keys       |
-| `keys create --key-type ed25519\|x25519 [--context-id ID] [--label LABEL]` | Create a key    |
+| `keys create --key-type ed25519\|x25519\|p256 [--context-id ID] [--label LABEL]` | Create a key    |
 | `keys get <key_id>`                                                        | Get a key by ID |
 | `keys revoke <key_id>`                                                     | Revoke a key    |
 | `keys rename <key_id> <new_key_id>`                                        | Rename a key    |
--- a/cnm-cli/src/main.rs
+++ b/cnm-cli/src/main.rs
@@ -503,7 +503,7 @@ enum AuthCredentialCommands {
 enum KeyCommands {
     /// Create a new key
     Create {
-        /// Key type: ed25519 or x25519
+        /// Key type: ed25519, x25519, or p256
         #[arg(long)]
         key_type: String,
         /// BIP-32 derivation path (auto-derived from context if omitted)
--- a/pnm-cli/README.md
+++ b/pnm-cli/README.md
@@ -185,7 +185,7 @@ url = "http://localhost:3000"
 | Command                                                                    | Description     |
 | -------------------------------------------------------------------------- | --------------- |
 | `keys list [--status active\|revoked] [--limit N] [--offset N]`            | List keys       |
-| `keys create --key-type ed25519\|x25519 [--context-id ID] [--label LABEL]` | Create a key (BIP-32 derived) |
+| `keys create --key-type ed25519\|x25519\|p256 [--context-id ID] [--label LABEL]` | Create a key (BIP-32 derived) |
 | `keys import --key-type TYPE --private-key KEY [--label L] [--context-id ID]` | Import an external private key |
 | `keys get <key_id>`                                                        | Get a key by ID |
 | `keys revoke <key_id>`                                                     | Revoke a key               |
--- a/pnm-cli/src/bootstrap.rs
+++ b/pnm-cli/src/bootstrap.rs
@@ -427,6 +427,7 @@ pub async fn run_connect(
         crate::config::VtaConfig {
             name: slug.clone(),
             vta_did: Some(credential.vta_did.clone()),
+            url: None,
         },
     );
     if pnm_config.default_vta.is_none() {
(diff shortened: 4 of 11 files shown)


stormer78 (changes requested):
## Review summary

Four of the five commits look good to merge as-is: the `ConnectInfo` rate-limiter fix, `--vta-url` persistence, clippy 1.95 cleanups, and P256 in `keys create`. One issue in the `create_did_key` VTA-identity commit is a merge-blocker IMO.

### 🔴 `did:key` VTA breaks DIDComm — fragment-ID mismatch

`create_vta_did_key()` in `vta-service/src/setup/from_toml.rs:727` saves key records at `{did}#key-0` and `{did}#key-1`, but a `did:key` resolves to verification methods named `{did}#{ed_pub_mb}` and `{did}#{x_pub_mb}`. `init_auth` correctly inserts secrets into the resolver under the right (multibase-fragment) IDs, but two other call sites still look up by `#key-0` / `#key-1`:

- `vta-service/src/server.rs:366-373` — DIDComm service init: both `sr.get_secret(...)` calls return `None` for a `did:key` VTA, so the `TDKProfile` is built with **zero secrets**. DIDComm fails silently as soon as a mediator is wired up.
- `vta-service/src/status.rs:324-344` — `send_trust_ping` does the same lookup and additionally derives X25519 from an independent BIP-32 path. The actual X25519 in a resolved `did:key` document is **curve-converted from Ed25519**, not independently derived, so the trust-ping uses a key no peer can encrypt to.

Also: the X25519 KeyRecord at `{did}#key-1` holds a public key that the resolved DID Document never advertises — two sources of truth that disagree.

The PR's end-to-end test (`pnm keys list`) only exercises REST, which is why this slipped through.

**Suggested fix:** branch on DID method in both `server.rs:362` and `status.rs:324` (the way `init_auth` already does) and use `#{ed_pub_mb}` / `#{x_pub_mb}` IDs for `did:key`. Drop the `#key-1` X25519 record for `did:key` — it's incorrect and unused. Please add a DIDComm-touching test (trust-ping against a local mediator, or a self-encrypt round-trip) for a `did:key` VTA so this can't regress. There's already a correct-shape helper at `keys::derive_and_store_did_key()` (`keys/mod.rs:87`) — used by the existing `vta create-did-key` CLI — that may be reusable.

### 🟡 One-shot `pnm setup` still drops the URL

`pnm-cli/src/setup.rs:182` (the phase-1+phase-2 one-shot branch when the operator answers the VTA-DID prompt at phase-1) still hard-codes `None` for `vta_url`, so URL persistence only works through `pnm setup continue`. Cheap fix: prompt for URL inline, or route the one-shot flow through the same code path as `continue`.

### 🟢 Smaller observations

- In `init_auth`, the "SIGNING KEY MISMATCH" / "KEY-AGREEMENT KEY MISMATCH" log messages lost their operator-facing prose ("DIDComm signing/verification will fail. This likely means the DID was created with different code or seed."). The structured fields survive, but the prose was load-bearing for an operator paging in cold — worth restoring.
- `signing_path.parse::<DerivationPath>()` then `root.derive(&dp)` in the new `did:key` branch is equivalent to `root.derive_ed25519(&signing_path)` used in the `did:webvh` branch — minor stylistic inconsistency, not worth churning.

The other four commits are good to land once §1 is resolved.

PR #111, opened 2026-05-15
feat: support REST-only registered servers and isolate daemon auth cache

carlos-affinidi:
## Summary

This PR fixes two related issues in the registered WebVH server flow:

1. VTA only recognized `DIDCommMessaging` and the legacy `WebVHHostingService` type, so valid `webvh-daemon` deployments advertising `WebVHHosting` could be rejected or fail REST fallback.
2. Adding daemon REST auth exposed a boundary problem: WebVH auth state was living on `WebvhServerRecord`, which is reused by public API/SDK/backup surfaces.

### 1. Hosting-type mismatch blocked registered `webvh-daemon` flows

The registered-server flow assumed server DIDs would expose either:

- `DIDCommMessaging`, or
- `WebVHHostingService`

Current daemon deployments can advertise `WebVHHosting` instead. That meant:

- `vta webvh add-server` / `pnm webvh add-server` could reject valid hosting DIDs
- transport resolution could fail for REST-only servers
- registered `webvh-daemon` deployments were not usable unless they also exposed DIDComm or the legacy service type

### 2. VTA did not drive the daemon REST auth flow itself

For REST-backed WebVH publication, the VTA needs to authenticate to the daemon using its configured `vta_did`, then reuse or refresh bearer tokens for subsequent operations. That flow was missing, which blocked publish/update/delete paths against registered REST-only servers.

### 3. WebVH auth cache leaked across public surfaces

`WebvhServerRecord` mixed operator-visible metadata with daemon auth cache (`access_token`, `access_expires_at`, `refresh_token`).

That same type is reused for:

- persisted `server:{id}` records
- REST/DIDComm `list webvh servers` responses
- SDK payloads
- backup export/import

So storing live daemon tokens on that public type would expose bearer state on multiple surfaces.

### Hosting-type alignment and transport resolution

- Accept `DIDCommMessaging`, `WebVHHosting`, and `WebVHHostingService` when validating registered WebVH servers
- Keep `WebVHHostingService` as a compatibility alias
- Update CLI help text and validation messaging to reflect the supported service types
- Centralize transport selection so VTA:
  - prefers DIDComm when available
  - falls back to REST for `WebVHHosting`
  - also falls back to REST for legacy `WebVHHostingService`

### Daemon REST auth support

- Add WebVH REST auth support in `WebvhClient`
- Implement the daemon challenge/response flow:
  - request challenge from `/api/auth/challenge`
  - sign the response as the configured `vta_did`
  - submit the signed message to `/api/auth/`
- Add access-token reuse and refresh handling
- Fall back to full re-authentication when refresh is unavailable or rejected
- Add response parsing compatibility for daemon camelCase fields while keeping existing snake_case compatibility

### Auth-state boundary fix

- Make `WebvhServerRecord` metadata-only
- Move daemon REST auth cache into a service-private record stored separately under `server-auth:{id}`
- Keep server metadata under `server:{id}`
- Remove auth cache when a WebVH server is deleted

This fixes the leak at the type/storage boundary instead of relying on ad hoc response redaction.

### Publish-path integration

Thread the authenticated REST transport through registered-server WebVH operations, including:

- create
- register existing DID with server
- update
- rotate
- delete
- protocol-management paths that publish via `update_did_webvh`

### Backup and compatibility behavior

- Exclude `server-auth:{id}` records from backup export/import
- Clear stale WebVH auth cache on restore
- Preserve read compatibility with legacy records/backups that still contain embedded auth fields by accepting them on deserialize and omitting them on serialize

## Result

This closes the functional blocker for registered REST-only `webvh-daemon` flows and fixes the auth-state leak by separating private daemon token cache from public WebVH server metadata.

## Validation

Added/updated targeted tests covering:

- supported server service types
- DIDComm-vs-REST transport resolution
- access-token freshness and auth-state persistence behavior
- metadata-only `list webvh servers` serialization
- server auth-cache deletion on remove
- backup export/import behavior excluding auth cache
- compatibility with legacy embedded auth fields
- daemon camelCase response parsing

## Notes

For REST-authenticated daemon deployments, the daemon ACL must contain the VTA DID, since the VTA authenticates as its configured `vta_did`.

Current REST auth still assumes the active VTA signing key is seed-derived; imported active VTA identity keys remain a follow-up.

### DIFF
--- a/pnm-cli/src/cli.rs
+++ b/pnm-cli/src/cli.rs
@@ -512,7 +512,7 @@ pub(crate) enum WebvhCommands {
         /// Server identifier
         #[arg(long)]
         id: String,
-        /// Server DID (must resolve to a DID document with a WebVHHostingService endpoint)
+        /// Server DID (must resolve to a DID document with a DIDCommMessaging, WebVHHosting, or WebVHHostingService endpoint)
         #[arg(long)]
         did: String,
         /// Human-readable label
--- a/vta-sdk/src/webvh.rs
+++ b/vta-sdk/src/webvh.rs
@@ -7,12 +7,6 @@ pub struct WebvhServerRecord {
     pub did: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub label: Option<String>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub access_token: Option<String>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub access_expires_at: Option<u64>,
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub refresh_token: Option<String>,
     pub created_at: DateTime<Utc>,
     pub updated_at: DateTime<Utc>,
 }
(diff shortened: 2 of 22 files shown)


stormer78:
This has been included in PR #113 

Closing this

PR #336, opened 2026-06-09
feat: DIDComm status endpoint + PNM health mediator discovery

marat-affinidi:
### 1. `GET /services/didcomm` status endpoint (VTA)
Returns `{enabled, mediator_did, websocket_status}` for DIDComm configuration discovery. Allows PNM and other clients to discover mediator connection status without requiring DID document resolution — works for all DID methods (did:key, did:web, did:webvh).

### 2. PNM health fallback to `/services/didcomm`
When VTA DID document has no DIDCommMessaging service (e.g. did:key), PNM health now falls back to `GET /services/didcomm` to discover the mediator DID. Enables mediator health checks for all DID methods. Fallback errors are surfaced in diagnostics output (not silently swallowed).

### 3. Fix `GET /services` (list all) 500 for did:key VTAs
`list_services()` unconditionally read the webvh on-chain DID document, causing a 500 `list_failed` error for did:key VTAs (which have no webvh record). Now short-circuits for non-webvh DIDs and returns service state from AppConfig:
- DIDComm: enabled flag + `messaging.mediator_did`
- REST: enabled flag + `public_url`  
- WebAuthn: enabled flag (no URL from config)

### 4. SDK: preserve full 409 Conflict body
`handle_response` now returns `VtaError::Conflict(raw_json_text)` for all 409 responses, preserving structured details (e.g. `mediator_did`) for callers. `enable_didcomm` simplified to use `handle_response` directly — no more hand-rolled request path. Body-read errors are propagated with `?` instead of `unwrap_or_default`.

### 5. Consolidate `EnableDidcommConflictBody`
Single public definition in `vta-sdk::protocol`, removed duplicate from `vta-cli-common`. CLI imports from SDK.

### 6. Feature-gate `didcomm_websocket_status` on `AppState`
Field gated behind `#[cfg(feature = "didcomm")]` — it is only written on the DIDComm path and read by `GET /services/didcomm`. Added doc comment noting it is only meaningful on the axum server path. The `enabled` field in the status response is also gated: always `false` on no-didcomm builds.

### Drive-by fixes (additive, required for CI)
- **`step_up.rs` feature gate**: `STEP_UP_APPROVE_REQUEST_TYPE` const and the `buffer_outbound` block gated behind `#[cfg(feature = "didcomm")]`. Fixes Feature combos CI job (`aws-secrets,keyring,rest` and `azure-secrets,keyring,rest`). Warning-free under `--no-default-features`.
- **e2e test fields**: Added `step_up_approver: None` and `step_up_require: None` to `UpdateAclRequest` in `client_didcomm.rs` (required after rebase onto main which includes #329 and #326).

## Testing
- E2E verified with deployed VTA (AWS) + local docker-compose harness
- `pnm services list` confirmed working for did:key VTAs
- `pnm services didcomm enable` (already enabled) returns friendly 409 message with mediator DID
- Unit test added: `returns_config_state_for_did_key_vta`
- All 6 existing list/list_drain unit tests pass
- All feature combos compile warning-free: default, `aws-secrets,keyring,rest`, `azure-secrets,keyring,rest`, `tee,rest,didcomm`
- SDK integration tests pass (14/14 including `enable_didcomm_409_maps_to_conflict`)
- clippy clean, fmt clean

### DIFF
--- a/cnm-cli/src/auth.rs
+++ b/cnm-cli/src/auth.rs
@@ -120,5 +120,5 @@ pub async fn connect(
     url_override: Option<&str>,
     keyring_key: &str,
 ) -> Result<vta_sdk::client::VtaClient, Box<dyn std::error::Error>> {
-    store().connect(keyring_key, url_override).await
+    store().connect(keyring_key, url_override, None).await
 }
--- a/docs/02-vta/runtime-service-management.md
+++ b/docs/02-vta/runtime-service-management.md
@@ -65,6 +65,26 @@ nature (DIDComm isn't running yet at first-enable).
 | Show in-flight drain entries | `pnm services didcomm drain list` |
 | Per-mediator traffic + sender attribution | `pnm services report [--since <rfc3339>] [--until <rfc3339>] [--format json|table]` |
 
+#### DIDComm status endpoint (`GET /services/didcomm`)
+
+`pnm services list` reports the advertised service array from the VTA's
+DID document. For DID methods with no resolvable service block (e.g.
+`did:key`), the mediator can't be discovered that way, so the VTA also
+exposes `GET /services/didcomm`, returning `{ enabled, mediator_did,
+websocket_status }` straight from runtime config. `pnm health` falls back
+to it to discover the mediator DID for those methods. Like every operation
+here, it is **super-admin** gated — parity with `pnm services list`, which
+exposes the same `mediator_did`.
+
+> **Caveat — `websocket_status` is server-path-only.** The live mediator
+> connection state (`connected` / `connecting` / `disconnected`) is tracked
+> only on the axum server path (`run()`). TEE / Nitro-Enclave front-ends
+> build their `AppState` via `build_app_state()`, which never wires the
+> listener event logger, so `websocket_status` there always reports
+> `disconnected` even when the mediator connection is healthy. Treat the
+> field as authoritative only on non-TEE deployments; use `enabled` +
+> `mediator_did` for configuration discovery on all deployments.
+
 ### Mutate REST
 
 | Task | Command |
--- a/pnm-cli/src/auth.rs
+++ b/pnm-cli/src/auth.rs
@@ -194,13 +194,19 @@ pub async fn show_token(keyring_key: &str) -> Result<(), Box<dyn std::error::Err
 
 /// Connect to the VTA using the preferred transport (DIDComm or REST).
 ///
-/// If `url_override` is provided, always uses REST.
-/// Otherwise resolves the VTA DID and prefers DIDComm when available.
+/// Transport selection priority:
+/// 1. `mediator_did_hint` from config → DIDComm (pinned).
+/// 2. VTA DID doc DIDCommMessaging service → DIDComm (resolved).
+/// 3. `url_override` + REST discovery → DIDComm if available.
+/// 4. `url_override` → REST-only fallback.
 pub async fn connect(
     url_override: Option<&str>,
+    mediator_did_hint: Option<&str>,
     keyring_key: &str,
 ) -> Result<vta_sdk::client::VtaClient, Box<dyn std::error::Error>> {
-    store().connect(keyring_key, url_override).await
+    store()
+        .connect(keyring_key, url_override, mediator_did_hint)
+        .await
 }
 
 #[cfg(test)]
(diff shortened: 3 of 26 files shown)


stormer78:
Nice work — this closes a real 500 and a genuine `did:key` discovery gap, and the test coverage is solid. A few suggestions before merge, roughly in priority order:

**1. `enabled` means two different things across the two endpoints.** In `list.rs` the short-circuit uses `didcomm_enabled && mediator_did.is_some()`, but `get_didcomm_status_handler` uses `config.services.didcomm` alone. A VTA with `services.didcomm = true` and an empty `mediator_did` would report `enabled: false` from `/services` and `enabled: true` from `/services/didcomm`. Worth picking one definition and sharing it.

**2. SDK conflict path can silently defeat the friendly CLI message.** In `enable_didcomm`, the full JSON body is preserved as `Conflict(text)` only when `body.error == "didcomm_already_enabled" && body.mediator_did.is_some()`. When DIDComm is already enabled but `mediator_did` is `None` (empty messaging config), it falls through to `from_http`, which passes just the error string — not JSON — so the CLI's `serde_json::from_str::<EnableDidcommConflictBody>` fails to parse and prints the raw error instead of "DIDComm already enabled." Dropping the `&& mediator_did.is_some()` clause makes both paths consistent.

**3. `EnableDidcommConflictBody` is defined three times** (inline in the SDK, in `vta-cli-common`, and re-parsed in the test). Could live once in the SDK — or better, surface `mediator_did` as a typed field on the conflict so the CLI doesn't re-parse JSON at all.

**4. `didcomm_websocket_status` on `AppState` isn't feature-gated**, but it's semantically a DIDComm concept (written only on the `didcomm` path, read only on the `rest` path). On a no-`didcomm` build it'd be a permanently-`Disconnected` field. Suggest gating it `#[cfg(feature = "didcomm")]` to match `didcomm_bridge`. It's also only live on the `run()` path — `build_app_state()` (TEE front-ends) never wires the event logger — worth a comment noting the field is meaningful only on the axum server path.

**5. The SDK rewrite drops the `rpc()` helper** in favor of a hand-rolled request just to extract one field. A one-line comment on why `rpc()` couldn't carry the conflict body would help keep the client surface uniform.

**6. Two undocumented drive-bys** — the `#[cfg(feature = "didcomm")]` gate in `step_up.rs` and the `step_up_*` e2e field additions. Worth a line in the PR description (and a quick check that `step_up.rs` builds warning-free under `--no-default-features`, given the `let Some(mediator_did) = …` binding).

None of these are blockers — #1 and #2 would be the most valuable to address and should both be quick. Happy to discuss any of them.

stormer78:
Pushed two follow-up commits from a review pass (both build + clippy clean, tests added):

**`b64fe19c` — hardening**
1. **Super-admin gate on `GET /services/didcomm`.** Changed the extractor from `AuthClaims` (any authenticated role) → `SuperAdminAuth`, for parity with the sibling `GET /services` (`list_services`), which exposes the same `mediator_did`. Without this, a reader-role caller could discover the mediator DID here that it cannot obtain from the list endpoint. The PNM health discovery path authenticates as the super-admin, so the documented use case is unaffected. Added `didcomm_status_forbids_non_super_admin` (asserts 403) and retargeted the "disabled" test to a super-admin token.
2. **Bounded the raw error body in `VtaClient::extract_error_message`.** The new `unknown error: {text}` fallback could embed a body up to the 1 MB request cap (e.g. an upstream proxy's HTML error page) into the error string that propagates to CLI output and logs. Now truncated to 256 chars + `…`. Added `oversized_error_body_is_truncated`. The 409 Conflict path is untouched (it intentionally preserves the full JSON for caller parsing).

**`34458131` — docs**
Documented the new `GET /services/didcomm` endpoint in the operator guide's Inspect section (it was undocumented), including the caveat that `websocket_status` is only meaningful on the axum server path — TEE/enclave front-ends build `AppState` via `build_app_state()`, which never wires the listener event logger, so the field always reports `disconnected` there regardless of real mediator connectivity. Mirrors the doc comment already on the `AppState` field.

One open question for you: #1 makes the endpoint super-admin-only. If you intended `pnm services didcomm` discovery to work for context-scoped (non-super) admins, say the word and I'll switch it to `AdminAuth` instead.

stormer78 (approved):
LGTM

PR #339, opened 2026-06-09
fix(vta-service): mediator wizard ships a usable WS endpoint and secret bundle

geoffturk:
## Summary

  Two follow-on fixes to the `vta setup` mediator-DID flow so a freshly-minted mediator boots with the right WebSocket endpoint and a complete secret bundle. Both interactive and `--from <toml>` setup paths are kept in sync.

  ### 1. `WS_URL` defaults now end in `/ws`

  The wizard's derived `WS_URL` did the scheme swap (`https` -> `wss`) but stopped there. The Affinidi mediator's WebSocket endpoint is canonically at `{base}/ws` alongside HTTP DIDComm at `{base}/`, so emitting `wss://host/mediator/v1` produces a DID doc that nginx 404s on every WS upgrade.

  - `default_ws_url_from_http` trims any trailing slash and appends `/ws`.
  - TOML setup applies the same `/ws` suffix when deriving `WS_URL` from `messaging.url`, so both paths agree.

  Confirmed against a live deployment: with the suffix the upgrade reaches the mediator and gets a 401 "Missing credentials" — same path, just no token — which is the expected protocol response.

  ### 2. Mediator wizard secret-bundle gaps

  The interactive mediator-DID flow could not actually render the `didcomm-mediator` builtin: the template started requiring `WS_URL` (commit 666dbd1) and neither setup path supplied it. The offline export also kept a copy of the pre-#337 label-as-kid
  defect that the SDK fixed for the online path. Together these meant a freshly-minted mediator could ship with no operating secrets the mediator-setup guard would accept.

  - Interactive setup prompts for `WS_URL`, defaulting to the HTTP URL with the scheme swapped.
  - TOML setup derives `WS_URL` from `messaging.url` the same way; non-http(s) URLs error early with guidance.
  - `select_secret_kid` moves from a private `client::secrets` fn into `vta_sdk::did_secrets` as `pub`, so the online and offline bundle builders share one implementation and cannot drift again.
  - `vta_service::operations::export::build_did_secrets_bundle` now calls `select_secret_kid` instead of its own `label.contains('#') || label.starts_with("did:")` check. Non-VM-id secrets are excluded with a `debug!` line; same contract as the SDK path.
  - `save_entity_key_records` writes the VM id as both `key_id` and `label`, instead of the freeform "<label> signing key" / "<label> key-agreement key". Belt-and-braces against #337-style regressions.

  ## Test plan

  - [x] `cargo test -p vta-service --lib` (677 pass)
  - [x] `cargo test -p vta-sdk --lib --features client` (152 pass, incl. the 5 `select_secret_kid` regression tests at their new location)
  - [x] `cargo clippy -p vta-sdk -p vta-service` clean
  - [x] `cargo fmt --check` clean
  - [x] Live-deployment check: WS upgrade reaches mediator and returns 401 "Missing credentials" instead of nginx 404.
  - [x] Re-run `vta setup` end-to-end against a clean store and confirm the rendered `didcomm-mediator` document carries the `/ws` WebSocket service and the saved secrets all have VM-id kids.

### DIFF
--- a/vta-sdk/src/client/secrets.rs
+++ b/vta-sdk/src/client/secrets.rs
@@ -1,6 +1,7 @@
 //! Convenience methods for paginating + bundling key secrets via [`VtaClient`].
 
 use super::VtaClient;
+use crate::did_secrets::select_secret_kid;
 use crate::error::VtaError;
 
 impl VtaClient {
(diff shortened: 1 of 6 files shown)


stormer78 (approved):
LGTM

PR #559, opened 2026-06-22
feat: serve VTA did:webvh log via canonical did.jsonl paths

yaroslava-kurash:
### Summary
Add public did:webvh log serving for the VTA at canonical resolver paths, with correct handling for both root and pathful DIDs.

### What changed
- Added public handler for `/.well-known/did.jsonl` for root-style VTA did:webvh.
- Added canonical path handler for pathful VTA did:webvh (`/<path>/did.jsonl`).
- Wired unauth route fallback to canonical path handler.
- Kept strict in-handler validation so only canonical `did.jsonl` paths return content.
- Use text/jsonl as the Content-Type for all `did.jsonl` responses, aligning with the `did:webvh` spec's SHOULD requirement
- Added tests for:
  - root DID returns from `/.well-known/did.jsonl`
  - pathful DID returns 404 on well-known path
  - pathful DID returns log on canonical path

### Why
did:webvh resolution rules require:
- `/.well-known/did.jsonl` only when no DID path segments exist
- `/<path>/did.jsonl` when DID includes path segments

This change aligns VTA behavior with spec-compliant resolver expectations.


stormer78:
Pushed a follow-up commit (ec132a2) hardening the edges of this change. Happy-path behaviour is unchanged; summary of what/why:

**Structure**
- Moved the VTA's own self-hosted `did.jsonl` handlers out of `did_webvh.rs` into a dedicated `routes::self_hosted_did` module. "Serve my own DID" is a distinct concern from generic did:webvh management, and this gives future self-hosted methods (`did:web`) an obvious home.
- Kept it under the existing `webvh` feature rather than a new flag: whether the VTA self-hosts is a **runtime** distinction (serverless vs server-managed, flippable in place via the promote op), so it isn't a compile-time axis.

**Security hardening**
- **Opaque 404s.** Every failure now returns a bare 404 with an empty body instead of distinct messages (`"VTA has no configured DID"`, `"...is not a did:webvh identity"`, `"...resolves to a pathful endpoint"`). Those let an unauthenticated caller fingerprint the VTA's DID configuration. Genuine storage errors still surface as 500 (logged) for operational visibility.
- **Catch-all no longer changes the global 404 surface.** Its non-match path returns a bare 404 byte-identical to axum's default fallback, so mounting the root wildcard doesn't turn every unrelated unknown GET into a JSON error body. A cheap suffix pre-filter keeps the config lock off that hot path.
- **`nosniff`** on the 200, matching the VTC `did_log` route (served at the router root, outside any security-headers middleware).
- Documented the **sole-root-wildcard invariant** at the route registration and handler. (Confirmed safe: the request path is only ever compared for equality against the path derived from the *configured* DID — never used as a store key — so no traversal.)

**Consistency**
- Aligned the VTC's `/.well-known/did.jsonl` to `text/jsonl` as well (did:webvh v1.0 SHOULDs it — DID-to-HTTPS Transformation §6), so VTA and VTC serve the same resource identically.

**Tests** — added opaque-404 cases (no DID / non-webvh DID / non-canonical path), bare-404 parity for unknown GETs, and a guard that the catch-all doesn't shadow a real authed route. All green; clippy clean (no new warnings).

**Versions** — `vta-service` 0.10.1→0.10.2, `vtc-service` 0.10.0→0.10.1 (both `publish = true`; per-PR bump convention — not a double bump, the PR hadn't touched versions).

**Out of scope (noted for a separate fix):** `/attestation/did-log` (TEE) advertises `application/jsonl` but actually returns a bare `String` (i.e. `text/plain`) — a pre-existing annotation/impl mismatch, left alone to keep this focused.

PR #562, opened 2026-06-23
feat(vta-sdk): public DI-signed did:key REST auth

stormer78:
## What

Adds a public, ergonomic way for any VTA consumer to authenticate a `did:key` over **plain REST** — the canonical DI-signed (`eddsa-jcs-2022`) `auth/authenticate/0.1` path — and obtain a bearer token / ready client.

- `provision_client::authenticate_did_key_rest(base_url, did, key, vta_did) -> Result<TokenResult, VtaError>`
- `VtaClient::connect_with_did_key(base_url, did, key, vta_did) -> Result<VtaClient, VtaError>` (`feature = "provision-client"`)

## Status: superseded — closing

Closing this. Not every VTA is reachable over REST, so a consumer shouldn't depend on a REST auth path — per the workspace guidance, DIDComm is the canonical transport and REST is only a fallback. The DIDComm `did:key` auth capability already exists in the SDK (`VtaClient::connect_didcomm` / `agent_session::AgentSession`), so this REST addition isn't the right approach and isn't needed. Consumers should authenticate over DIDComm instead.

### DIFF
--- a/vta-sdk/src/client/mod.rs
+++ b/vta-sdk/src/client/mod.rs
@@ -342,6 +342,37 @@ impl VtaClient {
         }
     }
 
+    /// Authenticate a `did:key` over plain REST (the canonical DI-signed
+    /// `eddsa-jcs-2022` path) and return a ready, token-bearing REST client.
+    ///
+    /// No DIDComm / mediator is required — this is the single call an external
+    /// consumer needs to go from a `did:key` + its private key to an
+    /// authenticated [`VtaClient`]. `client_did` must be a `did:key` whose seed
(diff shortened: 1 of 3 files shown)


stormer78:
Superseding this. Not every VTA is reachable over REST, so a consumer must not depend on a REST auth path — DIDComm is the canonical transport (REST is fallback). The DIDComm did:key auth already exists in the SDK (VtaClient::connect_didcomm / agent_session::AgentSession); the right change is to teach vta-mcp a scoped did:key DIDComm mode instead. Closing in favour of that.

PR #587, opened 2026-06-25
feat(vta-sdk): optional P-256 slots in didcomm-mediator template

marat-affinidi:
## What

Adds **opt-in P-256 (ES256 / ECDH-ES) verification methods** to the `didcomm-mediator` DID template, so a mediator can advertise P-256 identity + key-agreement keys alongside the mandatory Ed25519 + X25519 pair. Needed for counterparties that can't speak Curve25519 (JOSE-first stacks, hardware-backed wallets, ES256 issuers).

## Why this shape

The template renderer is pure `{TOKEN}` substitution with **no conditionals**, so optional keys can't be expressed as `{{#if}}` blocks. Instead this adds **null array-element pruning**: after substitution, an array element that resolved to JSON `null` is dropped. Combined with `optionalVars` that default to `null`, a template can list optional members as whole-string tokens that simply vanish when the caller doesn't supply them.

## Changes

- **`render.rs`** — prune `null` array elements after substitution (purely additive; no built-in template relies on literal `null` array members — verified).
- **`templates/didcomm-mediator.json`** — optional `VM_P256_SIGNING` / `VM_P256_KA` verification-method slots + `AUTH_P256` / `ASSERTION_P256` / `KEYAGREEMENT_P256` relationship references, all defaulting to `null`.
- **tests** — P-256 render test, a focused null-pruning unit test, and the existing default test now asserts the slots prune cleanly (document byte-identical to before when no P-256 vars are supplied).

## Validation

`cargo test -p vta-sdk did_templates` → **43 passed**. Backward compatible: without P-256 vars the rendered `didcomm-mediator` document is unchanged.

## Related

Consumer side (mediator-setup `--key-suite p256`) is in a follow-up PR on `affinidi/affinidi-tdk-rs`, which depends on a release of this `vta-sdk` change.

### DIFF
--- a/vta-sdk/src/did_templates/render.rs
+++ b/vta-sdk/src/did_templates/render.rs
@@ -114,7 +114,26 @@ fn substitute_value(value: &Value, vars: &HashMap<String, Value>) -> Value {
     match value {
         Value::String(s) => substitute_string(s, vars),
         Value::Array(items) => {
-            Value::Array(items.iter().map(|v| substitute_value(v, vars)).collect())
+            // Substitute each element, then prune any that resolved to JSON
+            // `null`. This is the mechanism for *optional* array members: a
+            // template writes a whole-string token (e.g. `"{VM_P256_SIGNING}"`)
+            // as an array element and declares it in `optionalVars` with a
+            // `null` default. When the caller supplies a value the element
+            // renders; when they don't, the `null` default makes it disappear
+            // rather than leaving a literal `null` in the document (which would
+            // be invalid in a `verificationMethod` / `authentication` array).
+            //
+            // Only whole-array-element nulls are pruned — a `null` nested
+            // inside an object value is left untouched. No built-in template
+            // relies on a literal `null` array member, so this is purely
+            // additive.
+            Value::Array(
+                items
+                    .iter()
+                    .map(|v| substitute_value(v, vars))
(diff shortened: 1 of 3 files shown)


stormer78:
**Minor (non-blocking) — partial P-256 supply can produce a silently-misconfigured document**

The 5 P-256 tokens (`VM_P256_SIGNING`, `VM_P256_KA`, `AUTH_P256`, `ASSERTION_P256`, `KEYAGREEMENT_P256`) are mutually independent in the template. If a caller supplies a VM object but omits its matching relationship reference (e.g. provides `VM_P256_SIGNING` but not `AUTH_P256`), the renderer emits a `verificationMethod` entry that no `authentication`/`assertionMethod`/`keyAgreement` array references — an advertised-but-unusable key, with no render-time error.

This is consumer-side responsibility (the follow-up `mediator-setup --key-suite p256` is expected to supply all 5 together), and it's not a security issue, so leaving the renderer general and enforcing coherence in the consumer is the right call. Flagging so the follow-up PR makes the 5 vars an all-or-nothing set.

Otherwise this looks clean — additive, well-tested, and the null-pruning is a sensibly general mechanism. The "no built-in template relies on a literal `null` array member" claim checks out: across all 8 templates every nullable token sits in object-field position, never as a whole array element.

PR #603, opened 2026-06-30
feat: vta-service preload self DID into resolver cache from local did.jsonl

yaroslava-kurash:
This PR improves VTA startup/runtime reliability for self-resolution of did:webvh in private or constrained network environments.

What changed
1. Added self-DID resolver preload during auth/resolver initialization.
2. The VTA now seeds DIDCacheClient with its own DID document from locally stored did.jsonl (WEBVH keyspace), instead of depending on network fetch of its public DID URL.
3. Added feature-gated wiring for webvh keyspace into init_auth.
4. Added unit tests for:
   1. successful preload and cache-hit resolution
   2. malformed local log as safe no-op (no poisoned cache)

Why
In some deployments (for example EC2/VPC-internal), the VTA may be unable to resolve its own public domain from inside its runtime environment. Preloading self DID from local did.jsonl removes that dependency and prevents avoidable resolution failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### vta-service — reliability: preload VTA self DID into resolver cache
+
+`vta-service` now preloads its own `did:webvh` DID document into the
+`DIDCacheClient` during auth/resolver initialization, using the locally stored
+`did.jsonl` log (`WEBVH` keyspace) as the source of truth.
+
+This avoids self-resolution network round-trips (and related startup/runtime
+failures) when a VTA cannot reach its own public domain from inside private
+network environments.
+
+Behavior is best-effort and non-fatal: if local log state is missing or
+malformed, the service logs a warning and falls back to normal resolver
+behavior.
+
 ### vti-common — security: keyspace values bound to their location (AAD); breaking on-disk format
 
 AES-256-GCM keyspace encryption now authenticates every value against its
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10163,7 +10163,7 @@ dependencies = [
 
 [[package]]
 name = "vta-service"
-version = "0.10.17"
+version = "0.10.18"
 dependencies = [
  "aes 0.9.1",
  "aes-gcm",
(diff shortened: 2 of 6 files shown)


stormer78:
Thanks for this — solid feature and the tests (cache-hit + malformed no-op) are exactly right. I pushed a commit (`db728d1`) addressing the review points so it's merge/publish-ready:

**Must-fix**
- **Version bump `0.10.17 → 0.10.18`.** `vta-service` source changed but the version didn't. `publish.yml` runs on every push to `main` and **silently skips any crate whose version already exists**, so without the bump this feature would merge green but never actually ship to crates.io (that's what happened to an earlier fix). Bumped + refreshed `Cargo.lock`.
- **Non-webvh build was a CI failure.** `preload_self_did_document`'s body is entirely `#[cfg(feature = "webvh")]`, so in a default (no-webvh) build its params were unused — and CI runs `cargo clippy --workspace -- -D warnings` on default features (`webvh` isn't default). Added a non-webvh no-op (`let _ = (…)`) that consumes the params. Verified: default clippy is now clean.

**Should-consider**
- **Staleness documented.** The preload runs once at init, so runtime `pnm services {…}` mutations (which republish `did.jsonl`) aren't tracked by the seeded entry. I documented why that's bounded/safe: per the runtime-service-management invariant `verificationMethod` stays byte-identical, so the keys DIDComm pack/unpack needs are never stale — only advertised `service` entries could drift in the VTA's *self*-view. Left re-seeding-on-mutation as a noted follow-up (the service-mgmt ops already hold the fresh log) rather than expanding this PR's scope.

**Nits**
- Tightened the doc comment to say the preload is **`did:webvh` only** (the earlier wording implied `did:web` too, but there's no local log to seed it from).
- Simplified `init_auth` to a single `webvh_ks: Option<&KeyspaceHandle>` param instead of the dual `#[cfg]` `webvh_ks` / `_webvh_ks` split (same type either way; the caller still passes `None` when webvh is off).

**Verification:** `cargo clippy -- -D warnings` clean on **both** default and `--features webvh`; both preload tests pass.

(Pushed directly since maintainer edits are enabled — shout if you'd rather I'd left it as suggestions.)

PR #616, opened 2026-07-02
feat: keep self-DID resolver cache in sync after DID-log mutations

yaroslava-kurash:
## Summary
This PR closes the self-DID resolver cache consistency gap by making post-mutation refresh explicit across runtime service-management and did_webvh mutation paths. It also hardens failure behavior (evict stale entries) and improves tests.

## What changed
- Added protocol-level refresh orchestration for the VTA self DID after service mutations.
- Hooked refresh into:
  - shared service lifecycle publish path (REST, TSP, WebAuthn enable/update/disable)
  - DIDComm service mutations (enable/update/disable)
- Added did_webvh refresh after local DID-log persistence (create and update paths).
- Fail-closed behavior:
  - on did-log read/parse/deserialize failures, cache entry is evicted best-effort.
- Updated startup/preload comments to match current behavior.
- Added changelog entry (Unreleased) summarizing this work.

## Testing
- Added and updated unit tests for:
  - protocol helper refresh from persisted did_log
  - protocol helper eviction when did_log is missing
  - did_webvh refresh seeding from did_log
  - did_webvh eviction on malformed did_log

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,18 @@ sync after runtime DID-log mutations, including protocol `services {…}`
 operations and did-webvh create/update paths.
 
 Highlights:
-- Added explicit protocol-layer post-mutation refresh for service-management
-  mutations (REST/WebAuthn/TSP shared lifecycle + DIDComm enable/update/disable).
-- Hardened refresh behavior to fail closed: on did-log read/parse/decode
-  failures, stale cache entries are evicted best-effort.
+- Centralized the post-mutation refresh at the DID-log write site: every runtime
+  mutation (did-webvh create/update and all protocol `services {…}` ops, which
+  funnel through `update_did_webvh`) reseeds the shared resolver cache once, from
+  the freshly-built log, right after it is persisted.
+- Fail-safe refresh: on did-log read/parse/decode failure the last-known-good
+  cache entry is kept (never evicted). For the VTA's own DID `verificationMethod`
+  stays byte-identical across service mutations, so a stale-but-present self-doc
+  still carries the exact keys pack/unpack needs — strictly safer than dropping
+  the entry, which would strand a serverless / network-unreachable `did:webvh`.
 - Kept startup preload + listener resolver reuse behavior aligned with runtime
   refresh semantics.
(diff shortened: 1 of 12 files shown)


stormer78:
Heads up @yaroslava-kurash — I pushed a follow-up refactor commit (6377fb7) after a review pass. It touches the resolver-refresh behavior this PR introduces, so flagging the reasoning here:

**1. Fail-safe instead of fail-closed on refresh error.** The helper previously *evicted* the cache entry when the did-log failed to read/parse/deserialize. For the VTA's own DID that's risky: `verificationMethod` stays byte-identical across service mutations (per the runtime-service-management invariant), so a stale-but-present self-doc still carries the exact keys DIDComm pack/unpack needs. Evicting would strand a **serverless / network-unreachable `did:webvh`** with no self-resolution source at all — turning a transient store glitch into a self-resolution outage. Changed both failure branches to keep the last-known-good entry (never evict, never poison).

**2. Removed a double refresh.** Every service-management op funnels through `update_did_webvh`, which already reseeds the cache from the freshly-built log at the persist site (`did_webvh::update::orchestrator`). The added `refresh_self_did_resolver_after_service_mutation` then re-read the *same* log back from fjall and reseeded the identical document a second time, under `PROTOCOL_LOCK`. Dropped that helper and its five call sites (enable/disable/update_didcomm + the REST/TSP/WebAuthn `publish_patch`); the orchestrator refresh is now the single source of truth, with a comment there so it isn't re-scattered later.

**3. Cleanup.** Removing the helper also dropped the duplicated `sample_did_log` fixture and duplicate eviction test. The remaining did_webvh test was rewritten to assert the new preserve-on-failure semantics.

Net −251 lines. Verified: `cargo fmt --check` + clippy clean, and 223 `operations::{protocol,did_webvh}` tests pass — including `update_refreshes_resolver_with_latest_document`, which proves the service-mutation → refresh path still works end-to-end through the centralized orchestrator refresh (so no coverage was lost by removing the protocol-layer helper).

Happy to split this into a separate PR instead if you'd rather keep authorship clean — just say the word.

PR #620, opened 2026-07-03
fix(vta-service): release setup store lock before admin/staff seeding

yaroslava-kurash:
## Summary
- Fixes a setup locking bug where non-interactive setup could fail with `FjallError: Locked` after DID creation, during admin or staff provisioning.

## Root cause
- In setup apply flow, the store is intentionally closed before later steps re-open it.
- One keyspace handle remained alive, so Fjall lock was still held and reopen failed on the same data directory.

## Changes
- Setup lock release fix in from_toml.rs


stormer78 (approved):
LGTM

PR #647, opened 2026-07-13
feat(pnm): add --transport rest to recover DIDComm-locked VTAs

abqdr-aff:
When a VTA advertises DIDComm but its mediator is unreachable, pnm's auto transport selection always picks DIDComm with no REST fallback, locking the operator out (services list / disable / enable all hang). Add a global --transport <auto|rest> flag that forces REST, skipping DIDComm even when it is advertised — e.g. `pnm --transport rest services didcomm disable` recovers a VTA that force-enabled DIDComm against a mediator it can't reach.

- vta-sdk (0.19.1): SessionStore::connect_with_transport + TransportChoice { Auto, Rest }; connect() is unchanged and delegates with Auto.
- pnm-cli (0.10.7): global --transport flag threaded into connect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,20 +2,39 @@
 
 ## Unreleased
 
-### pnm-cli (0.10.7) — `--transport rest` recovery flag
+### pnm-cli (0.10.7) / cnm-cli (0.10.7) — `--transport rest` recovery flag
 
-Adds a global `--transport <auto|rest>` flag. `rest` forces the REST transport,
-skipping DIDComm even when the VTA advertises it — the recovery path when a
-VTA's mediator is unreachable (auto-selection picks DIDComm and hangs with no
-fallback, locking the operator out). Example: `pnm --transport rest services
-didcomm disable` recovers a VTA that force-enabled DIDComm against a mediator it
-can't reach.
+Adds a global `--transport <auto|rest>` flag to both CLIs. `rest` forces the
+REST transport, skipping DIDComm even when the VTA advertises it and even when
+the local config pins a `mediator_did` — the recovery path when a VTA's mediator
+is unreachable and auto-selection would keep dialling it. Example: `pnm
+--transport rest services didcomm disable` recovers a VTA that enabled DIDComm
+against a mediator it can't reach.
 
-### vta-sdk (0.19.1) — force-REST connect path
+`pnm` also reconciles a pinned `mediator_did` after a successful `services
+didcomm enable|update|disable` (repoint on enable/update, clear on disable). The
+pin is priority 1 of transport selection and never re-reads the DID document, so
+a stale one would keep forcing DIDComm at a mediator that is gone.
+
+Docs: `docs/02-vta/runtime-service-management.md` gains a "Recovery: the mediator
+is unreachable" section.
+
+### vta-sdk (0.19.1) — force-REST connect path + bounded DIDComm connect
 
 - `SessionStore::connect_with_transport` + `TransportChoice { Auto, Rest }`:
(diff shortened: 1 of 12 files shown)


stormer78:
Reviewed this and pushed a follow-up commit (c9a0fcb) onto the branch rather than leaving a list of change requests — the flag itself is right, it just needed the surrounding pieces to be usable. Summary of what I added and why.

### The hang is now bounded — this is the root cause

`--transport rest` recovers a locked-out operator, but nothing told them the flag exists. `SessionStore::connect`'s DIDComm path had no deadline, and the mediator client owns a reconnect/backoff loop, so an unreachable mediator retried indefinitely and the CLI just sat there. (`pnm health` already caps its DIDComm probes at 15–30s for exactly this reason; the connect path didn't.)

Every auto-selected DIDComm connect now goes through a bounded helper — 30s default, `VTA_DIDCOMM_CONNECT_TIMEOUT_SECS` to raise it on slow links — and the timeout error names the recovery command verbatim, per CLAUDE.md's "operator errors should suggest the fix". Without this, the flag is only findable by people who already know it's there.

Worth noting the flag remains the *right* shape: CLAUDE.md forbids silently downgrading past what a peer advertises, so an automatic DIDComm→REST fallback would have been the wrong fix. Explicit opt-in plus a discoverable error is the correct pair.

### Forced REST no longer guesses a URL

The original branch fell through to `resolve_vta_url`, which synthesizes a base URL from the DID's own domain when the DID document has no `#vta-rest` service. For a hosted `did:webvh` that domain is the **DID host**, not the VTA — so you'd authenticate against the wrong origin and get a failure no operator could diagnose.

Forced REST now resolves `--url`, else the `#vta-rest` advertisement, else errors and asks for `--url`. The lenient `resolve_vta_url` behaviour is unchanged for the auto path.

### `TransportChoice` is `#[non_exhaustive]`

`vta-sdk` is published, and TSP is landing as a variant (TSP > DIDComm > REST). Without this, adding it later is a breaking change for any downstream exhaustive `match`.

### pnm reconciles the pinned `mediator_did`

This was the gap that left the recovery incomplete. A `mediator_did` in the PNM config is priority 1 of transport selection and **never re-reads the DID document** — so an operator who successfully ran `--transport rest services didcomm disable` would still be dialling the dead mediator on every subsequent command, needing the flag forever.

After a successful `services didcomm …`: `disable` clears the pin, `enable`/`update` repoint it. Unpinned VTAs never acquire a pin as a side effect. A config-write failure warns loudly instead of silently leaving a stale pin.

### cnm gets the same flag (0.10.7)

The SDK's new timeout error tells operators to use `--transport rest`, so that advice has to work in both CLIs. cnm has no `services` surface, but it hits the same hang against a dead mediator.

### Docs + tests

- `docs/02-vta/runtime-service-management.md`: new "Recovery: the mediator is unreachable" walkthrough (including why `--drain-ttl 0` is accepted over REST but not over DIDComm).
- Tests: hint reconciliation (4 cases), timeout env parsing incl. the `0` → default guard, and assertions that both new error messages keep naming `--transport rest` and `--url`. `cargo fmt --check`, clippy, and the pnm/cnm/vta-sdk suites are clean.

### Unrelated CI failure

`vta-sdk`'s `trust_tasks::tests::every_uri_in_canonical_namespace` fails on this branch — but it also fails on `main` (3b54316). The `task-consent` Trust Task URI added in #646 was never added to the canonical-namespace allow-list, so `https://trusttasks.org/spec/task-consent/decision/1.0` trips the census test. Not from this PR; needs its own fix.

stormer78 (approved):
LGTM

PR #652, opened 2026-07-13
feat: add automatic mediator ACL provisioning for DIDComm connections

yaroslava-kurash:
## Problem

Mediators configured with **`ExplicitAllow`** require DID-level ACL entries before a DID can communicate through the mediator.

VTA and PNM currently do not create these ACLs automatically. As a result, DIDComm communication fails when connecting through mediators that enforce explicit allow policies, because the required ACLs are not present.

## Solution

Adopt the deployment model where VTA/PNM use **`ExplicitDeny`** and automatically create DID-level ACLs whenever a DIDComm connection is established.

This removes the need for manual ACL setup and ensures VTA, PNM, and other SDK consumers can operate correctly in environments where mediator ACL enforcement is enabled.

#### New `acl-setup` feature

The feature provides automatic ACL provisioning for DIDComm connections while keeping the functionality opt-in for SDK consumers.

#### `acl_setup.rs` (new)

Introduces shared ACL provisioning logic:

* Hashes the connecting DID using SHA-256.
* Creates an allow-all `MediatorAcl`.
* Calls `atm.trust_tasks().acl_set()` to provision the ACL.
* Performs the operation in a fire-and-forget Tokio task so connection establishment is not blocked.

#### `didcomm_session.rs`

In `connect_with_secrets()`, ACL provisioning is triggered automatically after the WebSocket transport is enabled:
When `acl-setup` is enabled, every DIDComm connection automatically provisions the corresponding mediator ACL.

### PNM / CNM CLI

Enabled the SDK's `acl-setup` feature in the CLI feature configuration.

As a result:

* DIDComm connections automatically provision mediator ACLs.
* No manual ACL configuration is required.
* Existing CLI workflows remain unchanged.

### vta-service

Enabled and integrated the SDK's shared `acl-setup` functionality.
VTA now provisions the required mediator ACL after establishing its DIDComm listener connection during startup.

ACL provisioning is now **opt-in via config**: a `setup_acl` boolean field in the `[messaging]` section of `config.toml` (and in the `vta setup` wizard / `--from <toml>` schema) controls whether the VTA provisions its per-DID ACL on connect. Defaults to `false`. Existing configs are unaffected.

## Result

With these changes:

* DID-level ACLs are provisioned automatically whenever DIDComm connections are established  (PNM/CNM always; VTA when `setup_acl = true`).
* VTA and PNM can operate correctly with mediators enforcing ACL policies.
* Manual ACL configuration is no longer required.
* ACL provisioning logic is centralized in vta-sdk and reused by VTA and CLI consumers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,51 @@
 
 ## Unreleased
 
+### vta-service (0.11.0) — automatic ACL provisioning on startup
+
+* Enabled the SDK's `acl-setup` feature and integrated automatic
+  mediator ACL provisioning into VTA startup.
+* VTA now provisions the required DID-level mediator ACL immediately
+  after establishing its DIDComm listener connection, eliminating the
+  need for manual ACL setup.
+* Reuses the shared ACL provisioning implementation provided by
+  `vta-sdk`.
+* Allows VTA deployments to operate correctly with mediators with
+  stricter ACL enforcement policies.
+* ACL provisioning is performed transparently during startup and does
+  not alter existing DIDComm workflows beyond automatically ensuring
+  the required mediator access rules are present.
+
+
+### cnm-cli (0.11.0) / pnm-cli (0.11.0) — automatic ACL setup on DIDComm connect
+
+* Enabled the SDK's `acl-setup` feature by default in the CLIs.
+* DIDComm connections now automatically provision mediator ACLs during
+  connection establishment.
+* Improves interoperability with mediators enforcing DID-level ACL
+  policies by removing the manual ACL setup requirement.
+* Keeps CLIs workflows unchanged while ensuring ACL provisioning is
+  performed transparently in the background.
+
+### vta-sdk (0.19.3) — automatic mediator ACL provisioning for DIDComm connections
+
+- Added an optional `acl-setup` feature that automatically provisions
+  DID-level mediator ACLs when a DIDComm connection is established.
+  The implementation hashes the client DID (SHA-256), creates an
+  allow-all `MediatorAcl`, and submits it via
+  `atm.trust_tasks().acl_set()` in a non-blocking background task.
+- `connect_with_secrets()` now invokes ACL provisioning after DIDComm
+  transport initialization when the `acl-setup` feature is enabled.
+  Existing behavior is unchanged when the feature is not enabled.
+- Introduced a shared `acl_setup` module containing reusable ACL
+  provisioning logic for SDK consumers.
+- New feature dependencies:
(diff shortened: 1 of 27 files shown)


stormer78:
Pushed two follow-up commits (`3dbc9f4`, `ed60e5b`) to tidy this up before merge — no behaviour change to the feature itself, and `setup_acl` semantics are untouched.

### `3dbc9f4` — make ACL provisioning non-blocking and TSP-aware
- **Collapsed the double-spawn.** `didcomm_session.rs` wrapped `set_client_acl_on_connection` in a `tokio::spawn`, but that fn already spawned internally for the `acl_set` round-trip — a spawn-inside-a-spawn. The VTA path (`server.rs`) called it bare, so it ran `ATMProfile::new().await` inline on the startup path. Now `set_client_acl_on_connection` owns a **single** background task covering both the profile build and the mediator round-trip and returns immediately, so both call sites are identical and neither blocks.
- **`server.rs` robustness.** Replaced `config.vta_did.as_ref().unwrap()` / `app_state.atm.as_ref().unwrap()` with the already-bound `vta_did` and an `if let Some(atm)` guard (warn + skip). `setup_acl = true` can no longer panic startup if the ATM is absent.
- **Documented the DIDComm/TSP coverage.** The mediator ACL is keyed on `sha256(did)` (protocol-agnostic), and on the VTA both transports multiplex over the DID's single mediator socket — so one provisioning call authorises the account for both. I verified the hash against the mediator's account-key convention (`sha256::digest(did)` in affinidi-messaging-sdk) and the `acl_set` signature (takes a pre-hashed hash → no double-hash bug).
- Kept the `acl_set` error at `debug` (not `warn`): on an `ExplicitAllow` mediator the response can't return until this very ACL grants `receive_forwarded`, so an `Err`/timeout there does **not** mean the request was dropped.

### `ed60e5b` — correct the TSP-coverage note
The first comment overstated the gap. The SDK **does** have a dedicated client-side TSP session (`session::TspPingSession`, `atm.tsp().connect_websocket`), but it's a transient `pnm health` liveness probe on an ephemeral DID and intentionally persists no mediator ACL. Clarified that:
- the **general** client request transport (`VtaClient` / `TransportChoice`) is still DIDComm-or-REST only, so every *persistent* ACL-needing client connect is covered via `didcomm_session`;
- the `TODO(tsp-client)` now points at a future *persistent* TSP request transport, not a non-existent one.

### Verification
- `cargo check` clean for `vta-sdk --features acl-setup`, `vta-service` (default), **and `vta-service --features tsp`**.
- `cargo fmt --check` clean. Both commits DCO-signed.

No version/CHANGELOG re-bump — these are internal refinements to the same unreleased feature.

PR #734, opened 2026-07-22
fix(vta-service): update the webvh update-key test call sites for the 5-arg signature

stormer78:
**`main` does not currently compile its `vta-service` lib-test target**, so every branch cut after #731 inherits a red `Test` job. This unblocks it.

### What happened

#731 gave `load_active_update_key` a `seed_store` + `base_path` so it can re-derive a superseded signing key. It updated the real caller in `orchestrator.rs` and the two tests it added — but **four older unit-test call sites in `update/mod.rs` kept the 3-arg form**:

```
error[E0061]: this function takes 5 arguments but 3 arguments were supplied
error[E0277]: the trait bound `str: vti_secrets::SeedStore` is not satisfied
```

(the `SeedStore` error is the positional fallout — `scid` lands where `seed_store` is expected).

### Why it got through

The breakage exists **only in the test target**. `cargo check` and `cargo clippy --workspace` — without `--all-targets` — compile the lib fine, so the fast gates stay green. Only the full `cargo test --workspace` that CI runs surfaces it.

### The fix

`MockSeedStore` already exists in that same test module (#731 added it for its own tests), so this just passes it plus the module's conventional `base_path` (`m/26'/0'/0'`).

All four tests exercise the `webvh_keys` fast path and never reach the re-derive fallback, so **the mock's seed is never consulted** and no assertion changes — this is purely making the call sites match the signature.

### Verification

`cargo test -p vta-service --lib did_webvh`: **98 passed, 0 failed**.
`cargo test --workspace`: **3639 passed, 0 failed**.
`cargo fmt --all --check` clean.

Kept deliberately separate from the VTC migration work in #733, which is blocked on this — an unrelated fix buried inside a migration PR would be harder to review and harder to revert.

### DIFF
--- a/vta-service/src/operations/did_webvh/update/mod.rs
+++ b/vta-service/src/operations/did_webvh/update/mod.rs
@@ -350,9 +350,16 @@ mod tests {
         .await
         .unwrap();
 
-        let handle = load_active_update_key(&ks, scid, &[Multibase::from(pub_mb.clone())])
-            .await
-            .expect("found via webvh_keys");
+        let seed_store = MockSeedStore(Mutex::new(Some(vec![0x42u8; 32])));
+        let handle = load_active_update_key(
+            &ks,
+            &seed_store,
+            "m/26'/0'/0'",
+            scid,
+            &[Multibase::from(pub_mb.clone())],
+        )
+        .await
+        .expect("found via webvh_keys");
         assert_eq!(handle.hash, hash);
         assert_eq!(handle.version_id, "1-zV");
     }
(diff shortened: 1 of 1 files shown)


stormer78:
Superseded by #735, which landed the same fix while this was in flight. Verified against current main (f819ab58): `cargo test -p vta-service --lib did_webvh` → 98 passed, 0 failed. Closing as redundant.

Only cosmetic difference: #735 seeds its `MockSeedStore` with `None` where this used `Some(vec![0x42; 32])` — equivalent here, since all four of these tests take the `webvh_keys` fast path and never consult the seed.

Issue #749, opened 2026-07-23
TSP: land the pnm health ping correlation, then pick up the upstream delivery fix


stormer78:
Tracking the VTI-side work for the TSP delivery fault diagnosed on 2026-07-23.
The transport bugs themselves are upstream in
affinidi/affinidi-tdk-rs#646; this issue covers what has to happen here.

## Background

`pnm health` reported `✗ TSP ping failed: TSP websocket closed before reply`
against a TSP-enabled VTA that was demonstrably receiving and dispatching the
ping (`TSP trust-task dispatched … status=200 OK`, twice).

Root cause was upstream and total, not intermittent: the mediator never marked a
raw-TSP websocket live, so flush-on-connect was the only delivery such a socket
ever got — every reply that arrived after the socket came up was stored and left
in the inbox. Reproduced locally at 0/10 frames delivered with no VTA involved;
10/10 with affinidi/affinidi-tdk-rs#646 applied.

## Work here

- [ ] **`TspPingSession::ping` must correlate its reply.** It accepted the first
      frame that unpacked and parsed as JSON. The mediator inbox is durable, so
      every reply an earlier probe failed to collect is still queued and gets
      flushed on connect — meaning the probe could report a healthy round trip,
      at an invented latency, off a pong from a previous run. This is what made
      the fault look intermittent: the warm-up ping "succeeded" off the backlog
      and the measured ping then failed. Correlate on the Trust-Task `threadId`
      (responders set it to the request id via `doc.respond_with`), with the
      echoed `nonce` as fallback.

      Written and verified locally, with regression tests
      (`tests/e2e/tests/tsp_ping_correlation.rs`); not yet committed because it
      sits on top of in-flight `TspSession::send_document` / `receive_next` work
      in the same file.

- [ ] **Bump `vta-sdk`** — it's a `publish = true` crate and this changes its
      source.

- [ ] **Bump the `affinidi-tdk` pin** once affinidi/affinidi-tdk-rs#646 is
      merged and the crates publish (mediator 0.17.7, sdk 0.18.62,
      test-mediator 0.2.40).

- [ ] **Redeploy the mediator** at `webvh.storm.ws:mediator` on 0.17.7. The fix
      is server-side: no client change makes `pnm health` green without it.

- [ ] **Re-run the live probe** — `pnm health` from a `--features tsp` build
      against the TSP VTA — and confirm a real pong. Note that until the
[...]

PR #750, opened 2026-07-23
fix(vta-sdk): correlate the TSP ping reply instead of taking the first frame

stormer78:
`TspPingSession::ping` accepted the first frame that unpacked and parsed as JSON.
The mediator inbox is durable, so every reply an earlier probe failed to collect
is still queued and gets flushed onto the socket the instant we connect — the
probe could report a healthy round trip, at an invented latency, off a pong from
a previous run.

That is what disguised the TSP delivery outage as intermittent. `pnm health`
sends a warm-up ping then a measured one; the warm-up "succeeded" off the
backlog and the measured one failed, so the transport looked flaky when it was
in fact totally broken (0 of 10 frames delivered in a local reproduction). A
probe a stale frame can satisfy cannot be trusted to report a broken transport.

## The fix

Correlate on the Trust-Task `threadId` — a responder sets it to our request id
via `doc.respond_with` — with the echoed `nonce` as a fallback for a responder
that replies with a fresh document rather than a threaded `#response`.
Uncorrelated frames are skipped and we keep waiting within the caller's budget,
so a backlog delays the probe rather than faking it.

## Tests

`tests/e2e/tests/tsp_ping_correlation.rs`:

- **`a_stale_inbox_frame_does_not_satisfy_a_ping`** — plants a reply for the
  prober *before* it connects, then pings a peer that never answers. Without the
  fix it reports success in 9ms off the planted frame; with it, the ping times
  out. Runs against any mediator, since it only needs flush-on-connect.
- **`a_correlated_reply_satisfies_the_ping`** — the other half, so "reject
  everything" can't pass. A live round trip, so it needs mediator 0.17.7.

## Also included

Only what those tests need from the TSP session work in flight:

- `TspSession::send_document` — send an arbitrary Trust-Task document over TSP,
  the generalisation of `announce`.
- `TspSession::receive_next` reports a dropped connection as `Err`, not
  `Ok(None)`. Conflating "socket closed" with "nothing arrived yet" made a dead
  inbox look like a quiet one, so a polling loop spun instead of reconnecting.

Deliberately **not** included: the `vta-mobile-core`, `vta-service` inbound
concurrency, and `didcomm_session` work from the same investigation.

## Upstream

The transport faults themselves were in affinidi-tdk-rs and are fixed and
published — see affinidi/affinidi-tdk-rs#646 (the mediator never marked raw-TSP
websockets live, so anything arriving after connect was stored and never pushed;
and the SDK dropped packed frames that landed between polls). The mediator has
been redeployed on 0.17.7 and `pnm health` now gets a real TSP pong.

Lockfile moves to the published fixes: mediator 0.17.7, sdk 0.18.62,
test-mediator 0.2.40. **Note:** updating sdk and test-mediator alone does *not*
move `affinidi-messaging-mediator` — it is transitive through test-mediator and
needs its own `cargo update`. Without that the e2e round trip still fails, with
the entire server-side fix absent.

vta-sdk 0.19.21. Refs #749.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -341,9 +341,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-mediator"
-version = "0.17.5"
+version = "0.17.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ff1c3bcbd1c3bc5ef993c8d181cb010ead46d130aa31c4340821190946cc5800"
+checksum = "cac110108e4019d850ad9a75c65e5598ab8406a1f6868137165c2512c73b6c04"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-common",
@@ -356,6 +356,7 @@ dependencies = [
  "affinidi-rate-limit",
  "affinidi-secrets-resolver",
  "affinidi-task-utils",
+ "affinidi-tsp",
  "ahash 0.8.12",
  "async-convert",
  "async-trait",
@@ -400,7 +401,7 @@ dependencies = [
  "trust-tasks-rs",
  "url",
  "uuid",
- "vta-sdk 0.19.20 (registry+https://github.com/rust-lang/crates.io-index)",
+ "vta-sdk 0.19.20",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


stormer78:
The failing `Test` job is **pre-existing on main, not from this PR**.

`vta-service/tests/delegated_consent_e2e.rs::an_approval_from_a_non_context_admin_confers_no_execution` fails identically (same assertion, same line 648, `left: 400 right: 200`) on `main` at `02e26ffe` — the commit this branch is based on — in [run 29929410962](https://github.com/OpenVTC/verifiable-trust-infrastructure/actions/runs/29929410962), which predates this PR. The commit before it passed.

This PR touches no files under `vta-service/`:

```
$ git diff --stat origin/main..HEAD -- vta-service/
(no output)
```

Cause and fix tracked in #751: #748's fail-fast on unsatisfiable elevations replaced the behaviour that test asserts. Once that lands, this branch needs a rebase to go green — nothing to change here.

PR #752, opened 2026-07-23
fix(vta-service): update the consent e2e test to #748's fail-fast contract

### LINKED ISSUE #751: main is red: delegated_consent_e2e test encodes behaviour #748 deliberately changed
`main` has been failing CI since #748 (`02e26ffe`). The run for `02e26ffe`
failed; the commit before it (`d9abc9b8`) passed. Every PR branched from `main`
inherits the failure — e.g. #750.

```
thread 'an_approval_from_a_non_context_admin_confers_no_execution'
  panicked at vta-service/tests/delegated_consent_e2e.rs:648:5:
assertion `left == right` failed: a set member's decision is accepted
  left: 400
 right: 200
```

## Cause

Not a flake, and not a regression in behaviour — the test asserts the old
behaviour that #748 intentionally replaced.

#748 added a fail-fast to `trust_tasks/policy_gate.rs::consent_gate`: when the
requester can't self-authorize the subject context and fewer than
`min_approvals` members of the approver set can *confer* it, the request is
rejected immediately rather than minting a pending that can never execute. From
its commit message:

> Reject now, before minting a doomed pending, with a message that names the
> context and the gap.

`an_approval_from_a_non_context_admin_confers_no_execution` constructs precisely
that scenario on purpose — the approver administers `some-other-ctx` while the
subject context is `default` — and then asserts the ceremony proceeds:

```rust
let (status, _granted) = post(&router, &deleg_token, &decision).await;
assert_eq!(status, StatusCode::OK, "a set member's decision is accepted");
```

Under fail-fast the original `webvh-update` is rejected up front, so no pending
exists, so the decision references a challenge that was never minted → 400.

## Fix

Update the test to the new contract. Its *intent* — "an approval from a
non-context admin confers no execution" — is not only preserved but
strengthened: the task can never execute, and now it's refused at submission
instead of after a pointless ceremony. The assertions should become:

- the initial `webvh-update` is rejected immediately, with the error naming the
  subject context and the conferral gap (the actionable message #748 added);
- no pending/challenge is minted, so there is nothing to approve.

The current tail of the test (re-submit must not execute) becomes unreachable
and should be replaced by the fail-fast assertion.

Worth a second pair of eyes on whether any *other* case in this file quietly
depends on a doomed pending being mintable — the file has 5 tests and only this
one fails, which suggests not, but the fail-fast changed a shared path.

stormer78:
Fixes #751 — `main` has been red since #748 (`02e26ffe`), which blocks every PR
branched from it.

```
thread 'an_approval_from_a_non_context_admin_confers_no_execution'
  panicked at vta-service/tests/delegated_consent_e2e.rs:648:5:
assertion `left == right` failed: a set member's decision is accepted
  left: 400
 right: 200
```

## Why it broke

Not a flake, and not a behaviour regression — the test asserts what #748
deliberately replaced.

#748 added a fail-fast to `policy_gate::consent_gate`: when the requester can't
self-authorize the subject context and fewer than `min_approvals` members of the
approver set can *confer* it, reject at submission rather than mint a pending
that can never execute. The test builds precisely that scenario on purpose — the
sole approver administers `some-other-ctx` while the subject context is
`default` — and then asserted the ceremony proceeds and the decision returns
200. With fail-fast the original `webvh-update` is refused up front, so no
pending exists, so the decision references a challenge that was never minted →
400.

## What changed

The guarantee the test protects is unchanged and strictly stronger: an approval
from a non-context-admin confers no execution. What moved is *when* the system
says so. Renamed to `an_unsatisfiable_elevation_is_refused_before_any_consent_ceremony`
and re-pointed at the new contract:

- the `webvh-update` is refused at submission;
- the refusal is **actionable** — it says the elevation is unsatisfiable rather
  than merely forbidden, and names both the context (`` `default` ``) and the
  approver set (`` `operators` ``) that cannot confer it. A bare "forbidden" is
  exactly what sent this into the production loop #748 was fixing, so it's worth
  a test;
- **no challenge and no `consentRequests` are minted** — there is nothing for an
  approver to sign. This is the part that used to happen and shouldn't;
- nothing committed to the log.

## Checks

All five tests in the file pass. The other four were checked for the same
dependency on a mintable-but-doomed pending; none has it. Full `vta-service`
suite green, `cargo fmt --check` clean, and no new clippy warnings (the
pre-existing unused-import warnings in `did_peer.rs` / `did_webvh.rs` are
untouched and not in this file).

Patch bump 0.12.9 → 0.12.10.

Unblocks #750, which is otherwise green and inherits only this failure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,7 +400,7 @@ dependencies = [
  "trust-tasks-rs",
  "url",
  "uuid",
- "vta-sdk 0.19.19",
+ "vta-sdk 0.19.20 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -10141,39 +10141,6 @@ dependencies = [
 
 [[package]]
 name = "vta-sdk"
-version = "0.19.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aa4dadface2af16f9146338c6b3e6771381f43d29e2e7829d08ac4f66c4ab66f"
-dependencies = [
- "affinidi-crypto",
- "affinidi-did-resolver-cache-sdk",
- "affinidi-messaging-core",
- "affinidi-messaging-delivery",
- "affinidi-messaging-didcomm",
- "affinidi-openid4vci",
- "affinidi-openid4vp",
- "affinidi-tdk",
- "base64 0.22.1",
- "chrono",
- "curve25519-dalek",
- "didwebvh-rs",
- "ed25519-dalek",
- "futures-util",
- "getrandom 0.4.3",
- "multibase",
- "reqwest 0.13.4",
- "serde",
- "serde_json",
- "sha2 0.11.0",
- "thiserror 2.0.19",
- "tokio",
- "tracing",
- "url",
- "uuid",
-]
-
-[[package]]
-name = "vta-sdk"
 version = "0.19.20"
 dependencies = [
  "affinidi-bbs",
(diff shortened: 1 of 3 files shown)


stormer78:
Pushed `9719d67a` to fix the `lockfile self-pins` failure.

It was **already failing on main**, not introduced here — this branch touches no vta-sdk file (`git diff --stat HEAD~1 -- vta-sdk/` is empty). The workspace is on vta-sdk 0.19.20 while `Cargo.lock` still pinned the registry copy — pulled back in transitively through the test-mediator dev-dep chain — at 0.19.19.

```
STALE vta-sdk: workspace 0.19.20 but Cargo.lock pins registry copy at 0.19.19
```

Applied the one-line refresh the guard prints:

```
cargo update -p 'https://github.com/rust-lang/crates.io-index#vta-sdk@0.19.19' --precise 0.19.20
```

Both guards now pass locally:

```
ok   vta-sdk: workspace 0.19.20 == locked registry copy 0.19.20
ok   vta-service: 0.12.9 -> 0.12.10
```

Worth noting this is the failure mode the guard's own header describes: it bites at release time, not build time — `cargo publish --locked` verifies a dependent against the stale pinned copy, so the dependent builds against old vta-sdk source and fails, while the workspace path-dep build stays green throughout.

PR #771, opened 2026-07-23
fix: auth check

ivan-branets-affinidi:


stormer78 (commented):
## Review — approve

Correct, well-scoped, well-tested fix for a real authentication-bypass. I built and ran the full test suite for the touched code locally — all green: `vti-common` unit 7/7, `vta-service` messaging unit 4/4, `auth_flow` 6/6, `vault_unseal_authcrypt` 1/1, `vtc-service` `auth_forged_plaintext` 2/2.

### The vulnerability
`ATM::unpack` will parse a **plaintext** or **anoncrypt** DIDComm envelope and return an attacker-controlled `from` header with `authenticated: false`. Several handlers unpacked a body and then trusted `msg.from` as the proven signer. Because `/auth/challenge` hands out the challenge + `session_id` publicly, a remote unauthenticated attacker could POST a plaintext envelope with `from: <admin DID>` echoing the public challenge and be minted an admin JWT. The same shape defeated the vault "sealed" secret sender cross-check (forge `from == caller`) and the refresh sender-binding.

### Why the fix is right
`vti_common::auth::bind_authcrypt_sender(&msg, &metadata)` folds three checks into one and returns the *cryptographically-bound* sender:
1. `metadata.encrypted && metadata.authenticated` → rejects plaintext/anoncrypt (`NotAuthcrypt`).
2. `metadata.encrypted_from_kid` present (`NoSenderKey`).
3. **`base_did(from) == base_did(kid)`** (`Mismatch`) — the load-bearing check. Even an attacker who authcrypts with *their own* key (so both flags are true) can't claim a victim's `from`. A naive "just require encryption" fix would miss this.

Applied consistently at every authorization-sensitive direct-`unpack` site: VTA + VTC `authenticate` and `refresh`, and VTA `vault::unseal_secret`.

### Things I specifically checked and cleared
- **No regression to mediator-less REST auth.** VTA `authenticate`/`refresh` try their DI-signed Trust-Task path first (`try_authenticate_trust_task` / `try_refresh_trust_task`) and return early; VTC `authenticate` tries the SIOP `id_token` path (`authenticate_siop`) first. The new authcrypt gate only applies to genuine DIDComm envelopes, so the REST login/refresh flows are untouched.
- **Completeness.** Grepped every `.unpack(` and `msg.from` trust site. The remaining ones are safe: TSP unseal uses the intrinsically-bound `sender_vid`; the SDK client `open_from_vta` decrypts its own reply (not an auth surface); VTC `join_request_manifest_handler` is an explicitly-unauthenticated public discovery read; messaging-path handlers receive the already-verified sender because `handle_didcomm` overwrites `msg.from` with the transport-bound sender before dispatch.
- **`messaging/service.rs` refactor is behavior-preserving.** The old inline three-way (`!encrypted` / `auth_sender.is_none()` / dispatch) maps exactly onto `InboundGate::{NotEncrypted, Unauthenticated, Authenticated}`, and `plaintext_from` is still used for anoncrypt reply addressing. The extraction just makes the policy unit-testable.
- **Refresh `signer_did` change** (`Option` → `Some(sender_base)`): for legitimate authcrypt clients `from` is always present, so the bound value is identical to before; only forged/plaintext senders now fail.

The tests are notably good: each regression pin asserts the 401 is *attributable to the guard* (`err.contains("authenticated (authcrypt) DIDComm envelope")`) rather than the "ATM not configured" short-circuit, and `auth_flow` additionally asserts the challenge session stays `ChallengeSent` — so a future refactor can't make them pass without exercising the fix.

### Minor, non-blocking notes
- `build_offline_atm` is duplicated in `vta-service` and `vtc-service` `test_support.rs`. Acceptable (test-only, separate crates), just noted.
- **VTC `/v1/auth/refresh` has no REST-only path.** VTC `authenticate` gained REST parity via SIOP, and VTA `refresh` has the Trust-Task path — but VTC `refresh` goes straight to `atm.unpack`, so a REST-only client that logged in via SIOP can't refresh without DIDComm. Pre-existing, not introduced here; filing a tracking issue separately.
- The `Mismatch` error message returned to the caller names the authenticated DID — that's the attacker's own key, so no information leak.

### Process nit (not code)
The PR title/body is bare (`fix: auth check`, empty body). For a security fix of this severity, the body should describe the bypass and its impact so the audit trail isn't only in code comments.

stormer78 (approved):
LGTM

PR #794, opened 2026-07-25
fix(setup): make the wizard's scripted prompts feature-proof, and run them in CI

stormer78:
Fixes the `config-seed` failure surfaced while cleaning up #793. The setup wizard's two golden tests — `interactive_matches_equivalent_toml` and `advanced_existing_keys_mode_maps_through` — failed under `--features config-seed`, and had been failing since that feature landed.

```
$ cargo test -p vta-service --bin vta --features config-seed setup::interactive
panicked at interactive.rs:1102: expected Index answer for prompt: VTA DID
panicked at interactive.rs:1102: expected Index answer for prompt: DIDComm messaging
test result: FAILED. 0 passed; 2 failed
```

**CI never noticed** because the `Feature combos` job only runs `cargo check` — nothing in CI ever *executed* a wizard test under a non-default secrets-backend set.

## Cause

`ScriptedPrompter` replayed answers **positionally**, ignoring both the prompt and the option list it was handed. The secrets-backend menu is assembled feature by feature (`aws-secrets`, `config-seed`, `keyring`, …), so:

| build | menu | `Answer::Index(0)` selects |
|---|---|---|
| default | `["OS keyring", "Plaintext"]` | OS keyring ✓ |
| `+config-seed` | `["Config file", "OS keyring", "Plaintext"]` | **Config file** ✗ |

Picking the wrong backend meant the wizard never asked for a keyring service name, so the next scripted answer (`text("golden-keyring")`) was eaten by the wrong prompt — and the panic surfaced several prompts *downstream* of the actual mistake, which is why it read as unrelated.

A positional script cannot survive a menu whose length is a build-time property. There was already a `#[cfg(feature = "tee")]` answer in the script working around the same class by hand.

## Fix

New `Answer::Label(&'static str)` picks a `select` option by label, resolved against the `items` the harness already received and previously discarded:

```rust
Answer::Label(want) => items.iter().position(|it| *it == want).ok_or_else(|| …)
```

The two backend answers now say `Answer::Label("OS keyring")` instead of guessing its position. This is immune to options being added, removed, or reordered — and when a label genuinely disappears, it fails at the *right* prompt with the available options listed, rather than derailing the script.

Existing `Answer::Index` uses for fixed menus (log format, messaging kind, VTA DID kind, advanced mode) are left alone — those lists don't vary by feature.

## Verified across seven feature sets

All green; previously **only the default set passed**:

| feature set | before | after |
|---|---|---|
| *(default)* | ok | ok |
| `config-seed` | **FAILED** | ok |
| `config-seed,keyring` | **FAILED** | ok |
| `aws-secrets,keyring` | **FAILED** | ok |
| `vault-secrets,keyring` | **FAILED** | ok |
| `k8s-secrets,keyring` | **FAILED** | ok |
| `config-seed,vault-secrets,k8s-secrets,keyring` | **FAILED** | ok |

Plus the full suites unchanged: `vta-service` lib 690 passed, bin `vta` 53 passed, clippy and fmt clean.

## CI

`Feature combos` gained a step that **runs** these tests under `config-seed` and under `config-seed,vault-secrets,k8s-secrets`. The first is the exact command that reproduced the original failure.

Every other step in that job only compiles. The wizard is the one place a feature flag changes *behaviour* rather than just what builds, so `cargo check` was never going to catch this — which is precisely how it stayed broken.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,6 +248,18 @@ jobs:
       - name: vta-sdk client + sealed-transfer + provision-integration
         run: cargo check -p vta-sdk --features client,sealed-transfer,provision-integration
 
+      # Every step above only *compiles*. The setup wizard's scripted-prompt
+      # tests are the one place where a feature flag changes behaviour rather
+      # than just what compiles: each secrets backend adds an option to the
+      # backend menu, which shifts the answer script. `cargo check` cannot see
+      # that, so these two tests sat broken under `config-seed` — the only combo
+      # that both adds a backend and is cheap to build — until someone ran it by
+      # hand. Run them, don't just build them.
+      - name: vta-service setup wizard under extra secrets backends
+        run: |
+          cargo test -p vta-service --bin vta --features config-seed setup::interactive
+          cargo test -p vta-service --bin vta --features config-seed,vault-secrets,k8s-secrets setup::interactive
+
   enclave:
     # The Nitro enclave binary uses a TEE-only feature combo that the
     # default `check` job never touches. Linux-only because vsock-store
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,46 @@
 
 ## Unreleased
 
+### vta-service 0.12.37 — make the wizard's scripted prompts feature-proof, and actually run them in CI
+
+The setup wizard's two golden tests (`interactive_matches_equivalent_toml`,
+`advanced_existing_keys_mode_maps_through`) failed under `--features config-seed`
+and had been failing since that feature landed. CI never noticed because the
+`Feature combos` job only runs `cargo check`, so nothing in CI *executed* a
+wizard test under any non-default backend set.
+
+**Cause.** `ScriptedPrompter` replayed answers **positionally**, ignoring the
+prompt and its option list. The secrets-backend menu is assembled feature by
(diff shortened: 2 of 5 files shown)


stormer78:
Rebased onto `main` after #795 merged. The conflict was **not** just the changelog:

- **Version collision** — #795 also took `vta-service` 0.12.37, so this branch moves to **0.12.38**.
- **#795 reintroduced the defect.** Two of the three tests it added reach for the same `Answer::Index(0), // secrets backend = keyring` idiom this PR exists to remove, so they were broken under `config-seed` from the moment they landed:

```
a_pre_created_empty_data_dir_is_never_asked_about
  → expected Index or Label answer for prompt: VTA DID
existing_store_offers_reuse_and_delete
  → expected Index or Label answer for prompt: DIDComm messaging
```

Both now use `Answer::Label("OS keyring")`. All four backend selects across the five wizard tests are label-based; no positional backend answers remain.

`interactive.rs` itself auto-merged cleanly — #795 edited `gather_inputs` and added tests, this PR changed the `ScriptedPrompter` harness.

**The new CI step is what surfaced this.** Merging this PR unmodified would have turned `Feature combos` red against the new `main`, which is exactly the job it was added to do.

Re-verified post-merge across all seven feature sets (5 tests passed each), plus `vta-service` 690 lib + 64 bin and `vti-common` 342. Clippy and fmt clean.

PR #816, opened 2026-07-27
fix(vta-service): Nitro enclave attestation + did:webvh persistence

marat-affinidi:
## Summary

Fixes required to run the VTA signing service inside an AWS Nitro Enclave, plus
`did:webvh` log-persistence correctness fixes so a service keeps its identity
across restarts.

## Changes

**Enclave attestation**
- Fix the NSM ioctl request encoding: the length field passed to the Nitro
  Security Module ioctl must be a `u64`, not a `u32`. The wrong width corrupted
  the request struct on aarch64 and caused attestation calls to fail.

**did:webvh persistence**
- Persist the auto-generated `did:webvh` log to the webvh keyspace so the DID
  document history survives restarts.
- Persist the auto-generated `did:webvh` record and backfill it on restore, so a
  restored instance keeps its identity.
- Resolve the service's own DID via the network-mode resolver during the first
  enable handshake, instead of a resolver path that isn't available at that point.

**Deploy / build**
- Exclude signing keys and enclave runtime state from the Docker build context.
- Restore per-crate `COPY` layering + the full workspace manifest list for better
  build-cache reuse.

## Testing

Built and run inside the enclave: attestation succeeds and the DID persists
across restarts.

---
Opened as a **draft** for review/visibility.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,9 @@ docs/
 *.md
 !README.md
 .claude/
+# Keep enclave signing keys and Nitro runtime state out of the build context.
+# The Nitro build uses selective per-crate COPYs, but excluding these avoids
+# shipping private keys / large EIF artifacts to the Docker daemon.
+signing/
+.deploy-nitro/
+*.txt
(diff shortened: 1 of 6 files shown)


marat-affinidi:
Superseded by #819. `main` has since refactored the `tee` module out of `vta-service/src/tee/` into the standalone `vta-tee` crate, so this branch's changes no longer apply cleanly. #819 re-lands the NSM ioctl `u64` fix on top of current `main`. The did:webvh persistence changes will be re-derived against the new `operations/did_webvh` layout in a separate follow-up.

PR #819, opened 2026-07-27
fix(vta-tee): NSM ioctl len field must be u64, not u32

marat-affinidi:
## Summary

Fixes the AWS Nitro Security Module (NSM) ioctl request encoding used by the
enclave attestation path.

## The bug

The NSM ioctl takes `struct nsm_iovec { __u64 addr; __u64 len; }`. The Rust
binding declared `len` as `u32`:

```rust
#[repr(C)]
struct NsmIoBuffer {
    addr: u64,
    len: u32,   // 4 bytes of trailing padding under #[repr(C)]
}
```

Under `#[repr(C)]` the `u32` is followed by 4 bytes of **uninitialized**
padding to satisfy 8-byte alignment. The kernel reads the full 8 bytes as the
`__u64` length, so the padding becomes the high bits — inflating the length to
a garbage value and failing the ioctl with `EMSGSIZE` (os error 90).

## The fix

Declare `len` as `u64` (matching the kernel struct) and widen the two length
assignments accordingly. The struct size is unchanged (`0x20`), so the
`NSM_IOCTL_REQUEST` request code is unaffected.

## Testing

- `cargo check -p vta-tee` passes.
- Verified live on an AWS Nitro Enclave: attestation succeeds and
  `/attestation/status` reports `tee_type: nitro, detected: true`.

---
Opened as a **draft** for review.


stormer78 (commented):
## Review — correct fix, two follow-ups

Reviewed in a dedicated worktree against `origin/main` (6bb8b34d). **Synced:** I merged `main` into this branch and pushed (`417ff6df` → `31ee3e69`) so CI validates against the current base. `cargo check -p vta-tee --all-features`, `cargo fmt --check` and `cargo metadata --locked` are all clean post-merge.

### The diagnosis is independently corroborated by the ioctl number itself

`NSM_IOCTL_REQUEST = 0xC020_0A00` decomposes as dir=`0b11` (READ|WRITE), **size=`0x020` = 32**, type=`0x0A`, nr=`0x00`. `struct nsm_message` is two `nsm_iovec`, so the kernel is being told each one is 16 bytes — i.e. `{ __u64 addr; __u64 len; }`. That is exactly the shape this PR moves to.

It also explains why the bug presented the way it did rather than as an obvious size rejection: with `len: u32` the Rust struct *still* measured 32 bytes, because `#[repr(C)]` pads 4 bytes after `len` to keep the following `addr` 8-byte aligned. The ioctl size check passed; the kernel then read those 4 uninitialised padding bytes as the high half of a 64-bit length. EMSGSIZE (90) is precisely what you'd expect.

Worth calling out a **second latent bug this also fixes**: `msg.response.len` is read back after the call (line 251). With a `u32` field, that read only picked up the low half of the 64-bit length the kernel writes — so the `truncate()` was reading a differently-shaped field than the one the kernel wrote.

### 1. A size assertion is worth adding, with an honest caveat

Something like this next to the constant:

```rust
// The `size` field of _IOWR(0x0A, 0, struct nsm_message) is 0x20 = 32.
// Keep the Rust mirror in lockstep with it.
const _: () = assert!(std::mem::size_of::<NsmMessage>() == 32);
```

Caveat, and please encode it in the comment rather than trusting the assert alone: **this cannot catch a `u32` regression**, because padding restores the same 32 bytes. It's intent documentation tying the struct to the ioctl number, not a real guard. The actual guard against this class of bug is the explanatory comment you already added — which is the right call and reads well.

### 2. Missing changelog entry

`vta-tee` goes 0.1.0 → 0.1.1 with no root `CHANGELOG.md` entry. `scripts/check-changelogs.sh` **passes vacuously** here: it greps for the version string, and `0.1.1` already appears at `CHANGELOG.md:915-922` for `vta-audit` / `vta-backup` / `vta-vault` / `vta-webvh`. There is no `vta-tee 0.1.1` heading anywhere in the file.

The convention in that script's own header is "if a publishable crate's VERSION changed in this PR, the root CHANGELOG.md must MENTION THE NEW VERSION" — and sibling repos pin these crates loosely, so an attestation-path fix is exactly the kind of thing that needs to be findable. Please add a short `### vta-tee 0.1.1 — NSM ioctl `len` must be u64` entry.

(Separately: it's probably worth a follow-up to tighten that guard so a shared patch number in an unrelated crate can't satisfy it. Not this PR's job.)

Issue #709, opened 2026-07-20
VTC Trust Tasks: 20 bound tasks are absent from the published manifest


stormer78:
Split out of #537's manifest-drift follow-up. Pinned by the census added alongside this issue (`vtc-service/tests/trust_task_manifest.rs`), which currently carries them in an `UNPUBLISHED_OK` exception table so the backlog cannot grow silently.

## The gap

`trust-tasks/index.json` self-describes as the source of truth for trusttasks.org publication. Whole feature families shipped after it was last reconciled and were never added. The router enforces these task URIs; the manifest publishes none of them:

| Family | Tasks |
|---|---|
| Backup | `backup/export/1.0`, `backup/import/1.0` |
| Admin invites | `admin/invites/manage/1.0`, `admin/invites/revoke/1.0` |
| Invitations | `invitations/issue/1.0`, `invitations/revoke/1.0` |
| Members | `members/purge/1.0`, `members/removed/1.0`, `members/request-vmc/1.0`, `members/self-remove-receipt/1.0`, `spec/members/request-vmc/1.0`, `spec/members/vmc/1.0` |
| Auth | `auth/admin-session/1.0`, `auth/recognise/challenge/1.0` |
| Other | `directory/query/1.0`, `ceremonies/list/1.0`, `recognition/check/1.0`, `relationships/graph/1.0`, `spec/join-requests/submit-receipt/1.0` |

Plus five task directories on disk (`trust-tasks/credential-exchange/{issue,offer,present,query,request}/1.0`) that are in neither the manifest nor any binding — decide whether those are published or deleted.

## Work

Each entry needs a manifest row **and** `spec.md` + `schema.json` on disk — the census test `every_manifest_entry_has_files_on_disk` enforces the second half, so a manifest-only addition fails.

As each family is published, delete its row from `UNPUBLISHED_OK`. When the table is empty, this issue is done.

## Note on urgency

Lower than it first appears: no CI workflow currently reads `trust-tasks/index.json`, so the "CI publishes Draft entries on every merge to main" claim in the manifest's own `description` field is aspirational. Nothing is actually being published today. That is worth either building or removing from the description — and it is the reason this is a backlog item rather than a live consumer-facing defect.

Depends on the URI-shape question in the companion issue: if the authority/path shape changes, every one of these should be published in the corrected form rather than added twice.

stormer78:
Reviewed. This is done — resolved by a route the issue didn't anticipate.

[...]

PR #824, opened 2026-07-27
fix(vta-service): resolve self-DID via network-mode resolver in first-enable handshake

marat-affinidi:
## Summary

The transient "first-enable" handshake that trust-pings a newly configured mediator built a **default** TDK config, which resolves DIDs **locally**. In a serverless / network-isolated deployment the agent has no direct network egress, so resolving the mediator DID — and the VTA's own serverless `did:webvh` during mediator authentication — failed with a network error.

## Fix

- Route the first-enable handshake through the VTA's shared **network-mode** resolver (bridged to the parent-side resolver sidecar in an isolated deployment), mirroring the resolver wiring already used by the messaging service and the auth ATM.
- Re-seed the self-DID into the resolver cache immediately before the handshake: webvh cache entries expire after the cache TTL, and a serverless `did:webvh` can only be resolved from cache.

## Changes

- `vta-service/src/routes/protocol.rs` — build the handshake TDK config with the network-mode resolver + re-seed self-DID cache entry (feature-gated on `webvh`).
- `vta-service/src/server.rs` — make `preload_self_did_document` `pub(crate)` so the handshake path can reuse it.

## Testing

`cargo check -p vta-service` (default features: `setup`/`webvh` + `didcomm`) passes.


stormer78 (commented):
## Review — right mechanism, one silent-degradation path and a landing-order dependency

Reviewed in a dedicated worktree against `origin/main` (6bb8b34d). **Synced:** merged `main` into this branch and pushed (`6daf2f33` → `19aa5afb`). Post-merge `cargo check -p vta-service --no-default-features --features tee,rest,didcomm`, `cargo fmt --check` and `cargo metadata --locked` are clean.

### Mechanism checks out

`let mut reseed = resolver.clone(); preload_self_did_document(&mut reseed, …)` is the established house idiom, not a new pattern — `operations::did_webvh::refresh_resolver_doc_from_log` (mod.rs:277-278) does exactly `let mut cache = did_resolver.clone(); cache.add_did_document(did, doc).await;`. So the clone shares the cache and the reseed lands where the handshake will read it. Good.

Routing the transient handshake's TDK through the app resolver also mirrors `messaging::service::build_messaging` (service.rs:73-80) — same shape, including `with_load_environment(false)`.

### 1. `.build().ok()` silently degrades back to the bug being fixed

```rust
let tdk_config = TDKConfig::builder()
    .with_did_resolver(resolver.clone())
    .with_load_environment(false)
    .build()
    .ok();
```

If `build()` ever errors, `tdk_config` is `None` — and `transient_prove` (`transient_handshake.rs:129-133`) then falls back to `TDKConfig::builder().build()`, which is the default local resolver. That's precisely the enclave-egress failure this PR exists to fix, reached with no log line at all: the operator sees the *original* symptom, from the *fixed* build.

`messaging/service.rs:82-83` handles the identical call as `.map_err(|e| format!("build TDK config: {e}"))?`. Suggest matching that — propagate it (the function already returns `Result<_, HandshakeError>`), or at minimum `warn!` before falling through, so a degraded handshake is distinguishable from an un-patched one.

### 2. This is inert in TEE until #827 lands

The reseed reads the log through `preload_self_did_document` → `webvh_store::get_did_log(webvh_ks, …)`. On a TEE VTA with a self-issued `did:webvh`, the `webvh` keyspace is empty — that's the whole subject of #827. Until that merges, this block takes the `"no local did.jsonl found for resolver preload"` warn path and does nothing in the deployment it targets.

Not a defect in this diff — the code is right and is a no-op elsewhere — but **please land #827 first**, and it's worth a line in the PR body noting the dependency so it isn't merged alone and reported as "didn't fix it".

### 3. Nit: the `#[cfg(feature = "webvh")]` gate is redundant

`AppState::webvh_ks` is unconditional (`server.rs:110`), and `preload_self_did_document` already has a `#[cfg(not(feature = "webvh"))]` no-op arm that consumes its params for `-D warnings` cleanliness. Dropping the gate here removes one more place the two cfg arms can drift out of sync.

### 4. Conflicts with #827 on both the version and the changelog anchor

Both PRs bump `vta-service` and both insert immediately above `### VTC Trust Tasks — the unpublished-manifest backlog is closed (#709)`. #827 already claims 0.12.44, i.e. it assumes this one lands first — but per (2) the useful order is the reverse. Whichever goes second needs a rebase plus a version renumber; worth agreeing the order before either merges rather than discovering it in a conflict.

stormer78:
Re-synced with `main` after #827 merged. This one needed a conflict resolution, not just a merge.

**#827 took `vta-service` 0.12.44**, so this PR's 0.12.43 became a *version downgrade* rather than a bump — `check-version-bumps.sh` would have failed it. Resolved as:

- `vta-service` **0.12.45** (Cargo.toml + Cargo.lock)
- changelog heading renumbered to match; both entries kept, this one directly above #827's 0.12.44 entry (they collided on the same anchor, as flagged earlier)

Verified after resolution: the code diff is unchanged (`routes/protocol.rs` +28/-1, `server.rs` visibility change), `cargo check -p vta-service --no-default-features --features tee,rest,didcomm` clean, `cargo fmt --check` clean, both release guards pass.

Also worth noting for whoever merges: **the earlier red `Test` job was the known `join_didcomm` flake**, not this diff — it panicked in `vtc-service/tests/join_didcomm.rs:274` (`didcomm_join_round_trips_..._vmc_delivery`, 23.5s), a vtc-service mediator-timing test, while this PR only touches vta-service. I reran it and it passed.

The one open review point from my earlier comment still stands and is unaddressed: `.build().ok()` silently degrades to `tdk_config: None`, which `transient_prove` turns back into a default local resolver — the exact failure this PR fixes, reached with no log line. `messaging/service.rs` handles the identical call as `.map_err(|e| format!("build TDK config: {e}"))?`. Now that #827 has landed the reseed is live, so this is the last thing between here and merge for me.

stormer78:
Applied the fix myself rather than leaving it as a request — pushed as `6fa042b7`.

`.build().ok()` is now a match that propagates the failure as a `Connect`-stage `HandshakeError`:

```rust
let tdk_config = match TDKConfig::builder()
    .with_did_resolver(resolver.clone())
    .with_load_environment(false)
    .build()
{
    Ok(cfg) => Some(cfg),
    Err(e) => {
        let cause = format!("build TDK config: {e}");
        tracing::warn!(mediator = %mediator_did, error = %e,
            "first-enable handshake aborted: could not build a resolver-backed TDK config");
        return Err(HandshakeError::Failed { stage: HandshakeStage::Connect, cause });
    }
};
```

`Connect` because that is the stage `transient_prove` already assigns to this identical call (`transient_handshake.rs:129-133`), so the failure reads the same whichever side builds the config.

Fail-closed is the right call here rather than falling through: the only fallback `run_transient_handshake` has for `None` is a default, locally-resolving TDK — precisely the enclave-egress failure this PR exists to prevent. The caller already maps `Err` to an HTTP error, so `services didcomm enable` now refuses with the cause instead of proceeding on a resolver that cannot reach the network.

Also updated the function's doc comment (it claimed `Err` meant "the handshake actually ran and failed", which is no longer the whole story) and extended the changelog entry.

Verified: `cargo check` and `cargo clippy -- -D warnings` with `tee,rest,didcomm` both clean, `cargo fmt --check` clean, changelog guard passes. Diff is +38/-5 across two files on top of the sync merge.

That closes the last open point from my review. From my side this is ready once CI is green.

PR #826, opened 2026-07-27
perf(deploy/nitro): per-crate COPY layering + full workspace manifest list

marat-affinidi:
## Summary

The Nitro enclave image (`Dockerfile.nitro`) uses a manifest-first dependency-cache warm-up: copy every crate's `Cargo.toml` + a stub source, run a dummy `cargo build` to compile & cache all external dependencies, then copy the real sources. This only works if the stub phase lists **every** workspace member — otherwise `cargo build` fails workspace resolution, the `|| true` swallows the error, and dependencies are rebuilt from scratch on every image build.

The stub list had drifted to only **9 of the 26** workspace members, and the real-source copy covered the same subset.

## Fix

- List **all** workspace members for both the manifest/stub phase and the real-source copy, with the stub target kind matched to each crate's manifest (bin / lib / both; plus `vta-mobile-core`'s `uniffi-bindgen` bin).
- Keep `deploy/nitro/config.toml` baked only in Stage 2 (runtime). Never `COPY . .` in Stage 1, so editing config re-layers just the tiny Alpine stage instead of invalidating the Rust dependency cache.
- `tests/e2e` (`autobins = false`, integration tests only) has no compile target until its real sources are copied, so it gets a stub `src/lib.rs` in the warm-up phase purely to satisfy workspace resolution. **This was the actual bug** — without a target, `cargo metadata`/`cargo build` fails with `no targets specified in the manifest`.

## `.dockerignore`

Exclude `signing/` (enclave signing cert + private key), `.deploy-nitro/` (EIF artifacts, logs, pidfiles), and stray `*.txt` from the build context — avoids shipping private keys and large EIF artifacts to the Docker daemon.

## Testing

Reconstructed the Stage-1 stub layout (root manifest + `Cargo.lock` + all 26 member manifests + stub sources) in a scratch dir and confirmed `cargo metadata` (full dependency resolution) exits 0 — i.e. the warm-up `cargo build` resolves the workspace instead of silently failing. Verified the member list is an exact 1:1 match with `[workspace].members`.


stormer78 (commented):
## Review — this is a build-breakage fix, not a perf change. Please retitle.

Reviewed in a dedicated worktree against `origin/main` (6bb8b34d). **Synced:** merged `main` into this branch and pushed (`40bfe68a` → `e766f1dd`).

### `Dockerfile.nitro` on `main` cannot build at all today

Stage 1 on `main` copies 9 member manifests plus the root, but the root `[workspace].members` lists **26** after the `vta-service` decomposition (#780–#791). I replayed `main`'s exact Stage-1 COPY list into a scratch tree and ran `cargo metadata --locked --offline`:

```
error: failed to load manifest for workspace member `.../tests/e2e`
referenced by workspace at `.../Cargo.toml`
```

The warm-up build's `|| true` hides it, but the *second* `cargo build --release --package vta-enclave` has no `|| true` — so the enclave image build fails outright on `main`, it isn't merely running with a cold cache. Replaying this PR's COPY + dummy-source block the same way resolves cleanly.

`perf(deploy/nitro)` will get this triaged as an optional speedup. It's `fix(deploy/nitro)`: unbreak the Nitro image build after the subsystem extraction. Worth saying so in the body too, since anyone who tried to build an EIF since #791 hit this.

### The manifest list and target kinds are correct

I cross-checked all 26 members against their actual `src/lib.rs` / `src/main.rs` / explicit `[lib]` / `[[bin]]` sections. Every classification in your comment block matches, including `vta-mobile-core`'s `[[bin]] name = "uniffi-bindgen", path = "src/bin/uniffi-bindgen.rs"` and `vtc-service`'s lib+bin.

`tests/e2e` deserves credit — it has no `src/` at all in the repo, and a member with zero targets makes cargo error during manifest load, before it ever gets to building `vta-enclave`. The stub really is required, and the comment explaining why is the sort of thing that stops someone "tidying" it away later.

### Points

**1. `|| true` is still there.** The new comment says the swallow is what made this invisible — and then keeps it. Consider dropping it, or at least making the failure audible:

```dockerfile
RUN cargo build --release --package vta-enclave \
    --no-default-features --features ${FEATURES} \
    || echo "WARN: dependency warm-up failed — the cache is cold, the real build below will be slow"
```

Otherwise the next crate extraction reproduces exactly this bug, silently, and the only symptom is a slow build until someone runs the real one.

**2. `.dockerignore`'s `*.txt` is a no-op as written.** Docker ignore patterns use `filepath.Match` semantics — `*` doesn't cross `/` — so `*.txt` matches only root-level `.txt` files, and the repo has none (`find . -name '*.txt'` outside `target/`/`.git/` is empty). If the intent was nested Nitro artifacts like `pcr.txt`, that needs `**/*.txt`; if `signing/` and `.deploy-nitro/` already cover it, I'd drop the line rather than leave a rule that reads like it does more than it does. The two directory entries above it are both correct and worth keeping.

**3. Docs regression in the header.** The Architecture block was deleted:

```
#   Client → parent TCP:8443 → vsock:5100 → enclave socat → VTA :8100
#   VTA DIDComm → localhost:4443 → enclave socat → vsock:5200 → parent → mediator
#   VTA HTTPS   → localhost:4444 → enclave socat → vsock:5300 → parent → internet
```

That's the only place in this file mapping the vsock ports to what they carry, and #832 is concurrently expanding that map to seven channels. I'd keep it (updated to 5400/5500/5600/5700) rather than remove it. The new CACHING NOTE is genuinely useful — it just doesn't need to replace the port map.

**4. Minor:** `tests/e2e/src/lib.rs` survives `COPY tests/ tests/` because the repo has no such file to overwrite it, so the built image's workspace carries a phantom `vti-e2e-tests` lib target. Harmless under `--package vta-enclave`; worth one line next to the existing comment so it doesn't read as an accident.

**5. Expectation-setting, no action:** all 26 manifests are in one layer, so any crate version bump anywhere now invalidates the dependency cache. Inherent to the manifests-first approach and still far better than the status quo — just worth knowing why the cache misses after a release PR.

PR #827, opened 2026-07-27
fix(vta-service): persist auto-generated serverless did:webvh record + log

marat-affinidi:
## Summary

In TEE mode the VTA mints its own `did:webvh` on first boot from the KMS-bootstrapped seed (`tee::did_autogen::maybe_generate_vta_did`) and stores the DID + `did.jsonl` log **only** under the `KEYS`/`BOOTSTRAP` keyspaces. That path never goes through the `create_did_webvh` flow, so the `webvh` keyspace is left empty.

But three read paths resolve the VTA's own DID from the `webvh` keyspace via `webvh_store`:
- `list_services` — reads the record (`did:{did}`) **and** log (`log:{did}`);
- `server::preload_self_did_document` — reads the log to seed the resolver cache;
- `/.well-known/did.jsonl` — reads the log.

So on a serverless TEE VTA, `GET /services` 500s with `VtaDidRecordMissing` and the VTA's own DID is unresolvable from cache — no client can authcrypt to it.

## Fix

Add a small `tee_webvh` bridge in `vta-service` that idempotently backfills the `webvh` keyspace with the record + log after auto-generation, and call it from the enclave startup right after `maybe_generate_vta_did`.

- **Idempotent**: no-op for non-`did:webvh` identities and once the entries exist, so it also repairs an already-generated DID on a rebuild + restore — without wiping state (which would rotate the DID).
- **Record fields** mirror the production serverless builder in `operations::did_webvh::create_did_webvh` (`server_id = "serverless"`, next `#key-{n}` fragment `2`, one pre-rotation key). Any drift self-heals on the next did update/rotate, which re-scans the log and persists corrected counts.
- **Log** is copied from the `KEYS`-keyspace `did.jsonl` the auto-gen already wrote.

Kept in `vta-service` (which owns `webvh_store`) rather than the low-level TEE crate, so that crate doesn't take on a `vta-webvh` dependency.

## Testing

- `cargo check -p vta-service --features tee` passes (compiles the new `#[cfg(feature = "tee")]` module + the enclave feature graph).
- `cargo test -p vta-service --features tee tee_webvh::` — new unit test for the record builder (SCID = 3rd colon-segment, fragment/pre-rotation/log counts) passes.


stormer78 (commented):
## Review — the right fix, but the backfill guard is too broad

Reviewed in a dedicated worktree against `origin/main` (6bb8b34d). **Synced:** merged `main` into this branch and pushed (`bfd8b87f` → `f2b443e4`). Post-merge `cargo check -p vta-service --no-default-features --features tee,rest,didcomm` is clean, `tee_webvh::tests::record_scid_is_third_colon_segment` passes, `cargo fmt --check` and `cargo metadata --locked` clean.

### Blocking: a configured *hosted* did:webvh gets mislabelled `serverless`

`backfill_serverless_webvh_identity` gates only on `vta_did.starts_with("did:webvh:")`, but the enclave calls it with whatever `config.vta_did` holds — not with "the DID autogen just minted".

`deploy/nitro/config.toml:43` documents `vta_did = "did:webvh:..."` as a supported setting, and `did_autogen::maybe_generate_vta_did` returns early when it's set (`did_autogen.rs:40-42`). So on a TEE VTA with an explicitly-configured **hosted** did:webvh:

- `tee:did_log` is absent (autogen never ran) → the log branch is correctly skipped;
- the record branch is independent of it and **still fires** → writes `server_id: "serverless"` for a server-managed DID.

That field is load-bearing downstream. `operations::did_webvh::update::orchestrator.rs:594` keys publication off exactly it:

```rust
if mode == Mode::Execute && record.server_id != "serverless" { /* publish to host */ }
```

so a false `"serverless"` means every subsequent `did update` / `services … enable` **silently stops publishing to the host** while the local log keeps advancing — the hosted `did.jsonl` diverges with no error. `orchestrator.rs:170,280` additionally refuse agent-name ops for serverless DIDs. And because the log branch was skipped, you're left with a record but no log, so `list_services` still fails, just with a different error than before.

Suggested fix — gate on the autogen marker rather than the DID method: read `tee:vta_did` from the `KEYS` keyspace (the key `did_autogen.rs:218` writes) and only backfill when it equals `vta_did`. Belt and braces, make the two writes all-or-nothing so a partial backfill can't leave a record with no log.

### `TEE_DID_LOG_STORE_KEY`: the stated rationale doesn't hold

> Mirrored here (rather than imported) so this bridge doesn't pull `vta-tee` into `vta-service`'s dependency graph.

`tee_webvh` is itself `#[cfg(feature = "tee")]`, and under that same feature `lib.rs:79` already does `pub use vta_tee as tee;`. `vta-tee` is unconditionally in the graph wherever this module compiles — nothing is being avoided.

Meanwhile `"tee:did_log"` is now a bare literal in **four** places: written twice in `vta-tee/src/did_autogen.rs` (224, 231), read in `vta-service/src/routes/attestation.rs:92`, and re-declared here. `"tee:vta_did"` is a private const in `did_autogen.rs` that this PR wants too (see above). Suggest exporting both as `pub const` from `vta_tee::did_autogen` and using them at all call sites — otherwise the next key rename breaks the attestation route and this bridge silently.

### Verified correct

I checked the record construction field-by-field against the production serverless builder at `operations/did_webvh/mod.rs:1234-1246` — `server_id`, empty `mnemonic`, `log_entry_count: 1`, `next_fragment_id: 2`, `portable` all match. `pre_rotation_count: 1` matches `derive_pre_rotation_keys(…, 1)` at `did_autogen.rs:130`. `context_id: "vta"` matches the context autogen creates. SCID-as-third-colon-segment is right.

Also confirmed `#sealed-transfer-0` in the autogen DID document doesn't perturb `next_fragment_id: 2` — it isn't a `#key-{n}` fragment, so the "create mints #key-0 and #key-1, next is 2" reasoning holds.

Call-site ordering is correct: the backfill runs before `vta_service::server::run`, so `preload_self_did_document` at server init sees the log on the same boot. And the two independent `is_none()` guards mean a half-populated keyspace repairs rather than being skipped wholesale — nice.

### Nit: test coverage stops at the pure function

Only `build_serverless_webvh_record` is tested. `server.rs:2124` already has a `temp_webvh_ks()` tempdir-store helper you can reuse — an idempotency test (call twice, assert the second is a no-op) and a "non-`did:webvh` identity is a no-op" test are cheap, and the second one becomes the regression pin for the guard-broadening issue above once it's fixed.

### Landing order

0.12.44 assumes #824 (0.12.43) merges first, and both PRs insert at the same `CHANGELOG.md` anchor. But #824's resolver reseed is a no-op until *this* PR populates the `webvh` keyspace — so the useful order is 827 then 824, which means renumbering. Worth settling before either merges.

PR #831, opened 2026-07-27
docs(tee): document the serverless did:webvh identity lifecycle

marat-affinidi:
## Summary

Documents two load-bearing but previously-undocumented properties of a **serverless
TEE VTA** — one that sets `[tee.kms].vta_did_template` and mints its own `did:webvh`
from the enclave seed instead of publishing a `did.jsonl` to a hosting endpoint:

1. **The auto-generated DID record + log must be persisted to the `webvh`
   keyspace**, not only the key-material store, because `GET /services`, the self-DID
   resolver preload at startup, and `GET /.well-known/did.jsonl` all resolve the
   VTA's own DID from that keyspace.
2. **A serverless `did:webvh` is resolvable only from the local resolver cache** —
   there is no hosting endpoint and the enclave has no direct network egress, so all
   resolution routes through the parent-side network-mode resolver sidecar, and the
   self-DID is re-seeded into cache before the first-enable mediator handshake.

## Changes

- Add the `maybe_generate_vta_did()` → `webvh`-keyspace backfill step to the TEE
  startup sequence in `docs/02-vta/tee-architecture.md`.
- Add a new **"Self-issued did:webvh identity and resolution"** subsection.

Docs-only; no crate sources or versions touched.

### DIFF
--- a/docs/02-vta/tee-architecture.md
+++ b/docs/02-vta/tee-architecture.md
@@ -407,11 +407,45 @@ init_auth()
   ├─ Use bootstrapped JWT key (not config file)
   └─ Derive VTA identity keys → secrets_resolver
   ↓
+maybe_generate_vta_did()        [only if vta_did_template is set]
+  ├─ First boot: derive a did:webvh from the enclave seed and build did.jsonl
+  ├─ Persist the DID + log under the KEYS keyspace (key-material side)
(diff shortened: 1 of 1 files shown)


marat-affinidi:
Folded into #827 — the serverless did:webvh identity lifecycle doc now ships alongside the code that implements the persistence, in `docs/02-vta/tee-architecture.md`.

PR #832, opened 2026-07-27
docs(security-model): re-ground the TEE security model against current code

marat-affinidi:
## Summary

Re-grounds `docs/01-concepts/security-model.md` against the current TEE
implementation. The document had drifted from the code in several places; this
corrects the inaccurate/stale claims and fills gaps, verified against the
implementation.

## Corrections

**Pass 1 — stale/incomplete claims:**
- **KMS attestation (Recipient):** `GenerateDataKey`/`Decrypt` now pass an
  NSM-attested `Recipient`; KMS returns `CiphertextForRecipient` wrapped
  (RSAES-OAEP-SHA256) to an ephemeral in-enclave RSA-2048 key. Updated the
  residual-risk row (was "TODO/not implemented"), the AT1 tree, added a Layer 2
  bullet + A2 mitigation and a crypto-inventory row.
- **vsock channels:** Layer 7 said "three"; the proxy actually brokers seven
  (REST, mediator, HTTPS, IMDS, storage, resolver, logs). Enumerated all and added
  the log channel to the enclave↔proxy diagram.
- **Fail-closed boot:** documented that a KMS decrypt failure only re-mints
  identity on `AccessDenied` (post-rebuild PCR0 change) or explicit
  `allow_kms_reinit`; all other error classes refuse to start.
- **Anti-rollback anchor:** added the EBS rollback / state-replay threat and its
  mitigation (external monotonic counter, config-gated) to the A2 adversary, AT3
  tree, and residual-risk table.
- **Deployment checklist:** enclave-eligible instance floor + optional
  anti-rollback anchor configuration.

**Pass 2 — six code-grounding fixes:**
- Overview: non-TEE deployments use layers 5, 6, 8 (not "4–8") — Layer 4 (config
  locking) is KMS-gated and Layer 7 (vsock) is TEE-only.
- Layer 3: storage value format is `[4-byte magic "VAE1"][12-byte nonce][ciphertext
  + 16-byte GCM tag]`, AAD-bound to its `(keyspace, key)` location.
- Layer 5 / A2: the DID-method restriction is an optional allowlist (TEE-only, off
  by default) enforced at auth time — it does not unconditionally block `did:web`.
- Trust Boundaries: fjall storage physically lives on the parent EBS (semi-trusted);
  the enclave holds only the storage-encryption key in memory.
- Attack tree: KMS `Decrypt` uses end-to-end TLS enclave↔KMS over a CONNECT tunnel
  not terminated by the parent; `webpki-roots` is the proxy's mediator/resolver
  trust store, not the KMS path.
- Crypto inventory: add AES-256-CBC (KMS `CiphertextForRecipient` CMS content
  encryption).

Docs-only; no crate sources or versions touched.

### DIFF
--- a/docs/01-concepts/security-model.md
+++ b/docs/01-concepts/security-model.md
@@ -3,9 +3,10 @@
 ## Overview
 
 The VTA implements a defense-in-depth security model with eight layers of
-protection when deployed in TEE mode. Non-TEE deployments use layers 4-8;
-TEE deployments add hardware isolation, KMS-backed secrets, and encrypted
-storage.
+protection when deployed in TEE mode. Non-TEE deployments use the
+identity/access, seal, and audit layers (5, 6, 8); TEE deployments add
+hardware isolation, KMS-backed secrets, encrypted storage, config locking,
+and the vsock network controls (layers 1-4, 7).
 
 For TEE implementation details (KMS bootstrap code, encrypted store layer,
 config changes), see [TEE Enclave Security Design](tee-architecture.md).
@@ -35,7 +36,7 @@ config changes), see [TEE Enclave Security Design](tee-architecture.md).
 - Storage key derived from master seed via HKDF-SHA256
 - Deterministic derivation -- same seed produces same key across restarts
 - Keys stored in plaintext for prefix scans; values always encrypted
-- Each value: `[12-byte random nonce][ciphertext][16-byte auth tag]`
+- Each value: `[4-byte magic "VAE1"][12-byte random nonce][ciphertext + 16-byte GCM tag]`, AEAD-bound (AAD) to its `(keyspace, key)` location
 
 ### Layer 4: Configuration Locking
 - When KMS bootstrap is active, environment variable overrides are blocked
@@ -53,7 +54,7 @@ config changes), see [TEE Enclave Security Design](tee-architecture.md).
   - **Admin**: full key/DID/audit management within contexts
   - **Monitor**: infrastructure-only (metrics, health)
 - Context scoping restricts access to assigned application contexts
-- DID method whitelist blocks unsafe `did:web` in TEE mode
+- Optional DID-method allowlist (`tee.allowed_did_methods`, TEE-only, off by default) can restrict which DID methods may authenticate
 - Session state machine prevents challenge replay
 
 ### Layer 6: VTA Seal
@@ -240,7 +241,7 @@ Client                          VTA (in enclave)
 │     Semi-Trusted Zone            │
 │  (Parent EC2 instance)           │
 │  - enclave-proxy                 │
-│  - EBS volumes (ciphertext only) │
+│  - EBS (fjall, ciphertext only)  │
 │  - IAM role (limited)            │
 └──────────────┬───────────────────┘
                │ vsock (no network)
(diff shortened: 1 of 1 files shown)


stormer78 (commented):
## Review — every claim verified against the tree except one

Reviewed in a dedicated worktree against `origin/main` (6bb8b34d). **Synced:** merged `main` into this branch and pushed (`10d065d5` → `5765a2b6`).

This is the right kind of docs PR — it deletes a stale TODO that was understating what the code actually does, rather than adding aspiration. I checked each substantive claim:

| Claim | Verified against |
|---|---|
| `[4-byte magic "VAE1"][12-byte nonce][ct + 16-byte tag]`, AAD-bound to `(keyspace, key)` | `vti-common/src/store/encryption.rs:17`, `:35`, `build_aad` at `:44-50` |
| KMS `Recipient` → `CiphertextForRecipient`, RSAES-OAEP-SHA256 to an ephemeral in-enclave RSA-2048 key | `vta-tee/src/kms_bootstrap.rs:398-425` — `PrivateDecryptingKey::generate(KeySize::Rsa2048)`, `KeyEncryptionMechanism::RsaesOaepSha256` |
| CMS content encryption is AES-256-CBC | `kms_bootstrap.rs:774-796`, OID `2.16.840.1.101.3.4.1.42` |
| Fail-closed on KMS decrypt: only `AccessDenied` or `allow_kms_reinit` resets identity | `kms_bootstrap.rs:127-149` |
| `tee.allowed_did_methods` optional, TEE-only, off by default | `vta-config/src/lib.rs:287`, default `None` at `:451`; enforced at `vta-service/src/auth/backend.rs:140` |
| `tee.kms.anchor` anti-rollback | `vta-config/src/lib.rs:389` |
| Seven vsock channels: 5100/5200/5300/5400/5500/5600/5700 | `deploy/nitro/enclave-proxy/src/main.rs:42-66` |
| Non-TEE deployments get layers 5, 6, 8 only | no `storage_encryption_key` path in `vta-service/src/main.rs` — storage encryption really is TEE-only |

Replacing "DID method whitelist blocks unsafe `did:web` in TEE mode" with "optional allowlist, off by default" is a meaningful correction, not a wording change: the old text described a mitigation that a default deployment does not have. Same for the risk-table row — leaving "KMS Recipient parameter not implemented / TODO" in place while `nsm_attested_recipient()` exists is the kind of drift that makes a threat model actively misleading.

### Two follow-ups this surfaces (not blockers here)

**1. `deploy/nitro/parent-proxy.sh` has drifted and now collides.** It knows only four channels and assigns `VSOCK_RESOLVER_PORT=5400` (line 88) — but 5400 is the IMDS credential proxy on the enclave side (`enclave-entrypoint.sh:39`, `enclave-proxy/src/main.rs:54`), and the resolver is 5600. It has no concept of the storage (5500) or log (5700) channels at all.

The Rust `enclave-proxy` is clearly the canonical parent implementation now, and your new table matches it. So the doc is right and the script is wrong — but anyone deploying the socat path would wire the resolver into the IMDS port. Worth either fixing `parent-proxy.sh` or marking it deprecated in a follow-up.

**2. One checklist line I could not ground, and I think it's wrong.**

> - [ ] Parent instance is enclave-eligible (Graviton `.large`+ / x86 `.xlarge`+; T-family is not enclave-capable)

Nitro Enclaves requires at least 4 vCPUs on the parent instance. Graviton `.large` is 2 vCPU, so the floor should be `.xlarge` on both architectures, not `.large` for Graviton. Please double-check against the AWS supported-instance-types list before merging — it's the one line in this diff I couldn't verify against the tree, and a wrong minimum in a deployment checklist costs somebody an afternoon.

### Nit

Layer 7's heading is still "Network Controls" while the body now also covers storage I/O and log forwarding. "vsock channels" would describe the seven entries better. No strong feeling.

PR #835, opened 2026-07-27
feat(vta-service): hardened configuration — non-TEE at-rest encryption

yaroslava-kurash:
## Hardened configuration — non-TEE at-rest encryption

Adds `[hardened] enabled = true` in `config.toml` to bring storage encryption and sealed JWT key management to self-hosted (non-TEE) VTAs — the same protection as TEE layer 3, with the `[secrets]` backend as the trust anchor instead of the Nitro KMS.

### What changes

**Storage encryption** — all fjall keyspaces encrypted with AES-256-GCM (VAE1, same format as TEE). Key is HKDF-SHA256-derived from the master seed on every boot; never written to disk.

**Sealed JWT signing key** — random 32-byte key generated at first boot, AES-GCM sealed under the storage key in the `bootstrap` keyspace, SHA-256 fingerprint-verified on subsequent boots. Boot aborts on mismatch (tamper detection). `vta hardened rotate-jwt` clears it for rotation.

**Offline CLI** — all `vta` offline subcommands (`acl`, `keys`, `vault`, `webvh`, `bootstrap`, `unseal`, …) transparently derive and apply the storage key via the new `CliStore` wrapper.

**Memory safety** — `HardenedBootSecrets` holds `storage_key` and `jwt_key` and zeroizes both on drop, mirroring `tee::kms_bootstrap::BootstrappedSecrets`. `CliStore.enc_key` is also zeroized on drop.

### Crate versions

| Crate | Version | Change |
|---|---|---|
| vta-sdk | 0.20.7 → **0.20.8** | `BackupPayload` / `ImportedSecretBackup` `Drop` impls |
| vta-support | 0.1.x → **0.2.0** | `require_unsealed` signature change; `run_unseal_challenge` + friends gain `enc_key` param |
| vta-service | 0.12.x → **0.13.0** | Hardened configuration feature |
| vta-enclave | → **0.7.7** | Warn + ignore `hardened.enabled` (TEE KMS takes precedence); dep pin update |
| vta-backup, vta-tee | patch bump | vta-support dep pin update |

### Docs updated

- non-interactive-setup.md — canonical operator reference (`#hardened-configuration`)
- tee-architecture.md — mutual exclusivity with TEE
- security-model.md — layer notes
- secret-backends.md — backend selection callout


stormer78 (commented):
## Review — the crypto is sound; the migration path is the blocker

Reviewed in a dedicated worktree against `origin/main`. The core is genuinely good: HKDF derivation is domain-separated (`vta-storage-key/v1` vs the VTC's), pinned by a test vector so the algorithm can't drift silently, the JWT key is random-and-sealed rather than derived (so it rotates without rotating the seed), and `load_or_generate_jwt_key` has real integration tests against a live fjall store covering first boot, restart, wrong key, missing fingerprint, tampered fingerprint and rotate. `HardenedBootSecrets` mirrors `BootstrappedSecrets`'s zeroize-on-drop. The fresh-install path encrypts from the first byte, which is exactly right.

The problems are around the edges of that core, and one of them will brick a VTA.

### 1. Blocking — enabling this on an existing VTA destroys access to it, with nothing to stop you

`HardenedConfig`'s doc says it plainly (`vta-config/src/lib.rs:466`):

> **Migration note**: enabling on an existing plaintext-fjall VTA requires a one-time `migrate_to_encrypted` pass before starting with this flag.

That pass is not shipped, and nothing detects that it was skipped. `main.rs` sets `storage_encryption_key = Some(key)` and boots. There is no `vta hardened migrate` next to `vta hardened rotate-jwt`, and no `is_v1_encrypted` probe anywhere in `vta-service` or `vta-config` — I grepped.

The store is deliberately **strictly fail-closed** — `vti-common/src/store/mod.rs:156-158` says so in as many words: *"This deliberately does not add a lenient read-fallback to the decrypt path — that would reintroduce the cut-and-paste downgrade hole."* That's the correct design, and it means every pre-existing row now errors on read. Including the ACL keyspace. The operator flips one boolean in `config.toml`, restarts, and is locked out of their own VTA with no obvious way back — the data is intact but unreadable through the only handle the daemon will now open.

The pattern already exists in this workspace: `vtc-service/src/server.rs:511-512` and `:1523` run `migrate_to_encrypted` at boot for exactly this transition. Two options, either is fine:

- **Detect and refuse.** At boot, if `hardened.enabled` and any row in a representative keyspace fails `is_v1_encrypted`, exit with "run `vta hardened migrate` first". Cheap, and it converts a bricking into an error message.
- **Detect and migrate**, like the VTC does. `migrate_to_encrypted` is already idempotent and crash-safe, so an interrupted run completes on the next boot.

I'd want at least the first before this merges. The doc note is not a substitute — it describes a footgun rather than removing it.

### 2. The plaintext-backend warning is inverted for the worst case

`main.rs:2118-2127` fires the "you're on the plaintext fallback" warning only when **every** backend field is unset:

```rust
let looks_like_plaintext_backend = config.secrets.seed.is_none()
    && config.secrets.aws_secret_name.is_none()
    && ... 
```

But `secrets.seed` is *"Hex-encoded BIP-32 seed (config-seed feature)"* (`vti-secrets/src/config.rs:51-52`) — the seed inlined **in `config.toml` itself**. Setting it makes the condition false, so a config-seed deployment gets **no warning at all**.

That is the case that most needs one. This feature's stated purpose is keeping secrets out of `config.toml`; with `[secrets] seed = "…"` the storage key and the sealed JWT key are both re-derivable by anyone who can read that file, which is the same adversary the module header names ("a root user can forge any token", "readable by anyone with disk access"). The feature reduces to obfuscation, silently.

`config.secrets.seed.is_some()` belongs on the *warn* side, not the *don't-warn* side. I'd argue it should refuse rather than warn — `hardened.enabled = true` plus `[secrets] seed` is a contradiction, not a tradeoff — but at minimum invert it.

### 3. `unwrap_or(None)` conflates "hardened is off" with "couldn't get the key"

- `main.rs:1487-1489` — `load_storage_key_for_cli(&config).await.unwrap_or(None)` before `run_unseal_challenge`
- `status.rs:175` — `.ok().flatten()`

`load_storage_key_for_cli` returns `Ok(None)` for "hardened disabled" and `Err` for "seed store unreachable / no seed / backend misconfigured". Both collapse to `None`, and `None` means *open bare handles*. So a transient keyring or Vault failure during `vta unseal` doesn't say "couldn't reach the secret store" — it silently behaves as though encryption were off and fails somewhere further down with a deserialization error against `VAE1` bytes.

`cli_store.rs:24`'s own doc comment already says the right thing: *"If the seed load fails the error is returned and the caller should surface it to the operator."* Two callers don't. Propagate at both.

### 4. Version collisions with `main`

Two of the claimed versions have been taken since this branch was cut:

| Crate | Claims | On `main` | Needs |
|---|---|---|---|
| `vta-sdk` | 0.20.8 | **0.20.8** (#837) | 0.20.9 |
| `vta-tee` | 0.1.2 | **0.1.2** (#836) | 0.1.3 |
| `vta-service` | 0.13.0 | 0.12.47 | ok |
| `vta-support` | 0.2.0 | 0.1.2 | ok |

These are downgrades-in-waiting — `check-version-bumps.sh` will fail the PR. The branch is also `CONFLICTING` against `main` right now.

### Smaller notes

- **`main.rs:2107` still says "Hardened configuration (PoC)"** in the daemon boot path. For a shipped security feature that word should go.
- **The fingerprint is close to redundant here.** In TEE it detects KMS ciphertext tampering, where the KEK lives outside the store. Here the AES-GCM tag already authenticates the ciphertext under the storage key, so an attacker who can't derive that key can't produce a ciphertext that decrypts at all. What the fingerprint *does* add is a failure mode: deleting the single fingerprint row is fatal (`FingerprintMissing`), and the remedy the error prints is "clear both entries to regenerate" — i.e. one deleted row escalates to a forced JWT rotation that invalidates every live session. Worth either dropping it or making a missing fingerprint self-heal from the successfully-decrypted key.
- **The salt is documented as a "permanent per-VTA constant" but is never per-VTA.** Interactive setup always takes `HardenedConfig::default()` → `"vta-storage-v1"`, with a custom value only reachable via `--from <toml>` (`interactive.rs:189-206`). To be clear, this is **not** a cryptographic weakness — HKDF salts needn't be secret and the IKM is a per-VTA high-entropy seed, so there's no cross-VTA collision. It's a docs/UX mismatch: either generate a random salt at setup, or stop describing it as per-VTA.
- **Zeroize completeness.** `let storage_key = *derive_storage_key(...)` (`hardened_bootstrap.rs:292`) copies out of the `Zeroizing` wrapper into a plain `[u8; 32]`, which is then moved into `HardenedBootSecrets`; the moved-from stack slot isn't zeroized. Same shape at `cli_store.rs:39` and at `main.rs:2152` (`storage_key_copy`). This is inherent to `zeroize` + moves rather than a bug in this PR, and the exposure is small — worth knowing rather than fixing.

### Summary

The cryptographic design is right and the test coverage on the core is better than most of what lands here. (1) is a blocker — a config flag that silently bricks an existing deployment needs a guard, and the workspace already has the pattern. (2) is a small change with a large effect on whether the feature delivers what it claims. (3) and (4) are mechanical.

Happy to do the `main` sync and version renumber whenever you want it — I've held off since a rework for (1) may change the shape of the diff.

stormer78:
Synced with `main` and implemented the review items. Pushed as `a0bdf8ed` (sync) + `9b49f5b7` (fixes).

## Auto-migration on first hardened boot — the blocker

`hardened_bootstrap::migrate_store_to_encrypted` now runs at boot, before `server::run` opens a single encrypted handle. Idempotent (skips rows already in `VAE1`, so later boots are one prefix scan per keyspace and zero writes) and crash-safe (an interrupted run is completed by the next boot), mirroring what `vtc-service` already does.

**The `bootstrap` keyspace is deliberately excluded, and that exclusion is load-bearing.** Its hardened JWT ciphertext is written *and read back* through a bare handle — `server.rs` opens it "unencrypted, KMS-protected". Migrating it would wrap that ciphertext in a `VAE1` envelope, so the next boot's `aes_gcm_open` would fail as though the ciphertext had been tampered with: an unbootable VTA whose error message points at the wrong cause. `migration_leaves_bootstrap_keyspace_bare` pins it by running the exact call the next boot makes.

Three tests, all against a real fjall store:

- `migration_makes_preexisting_plaintext_rows_readable` — writes a legacy plaintext row, asserts the encrypted handle **errors** on it (confirming the fail-closed behaviour is real and not theoretical), migrates, then reads it back.
- `migration_is_idempotent` — second pass converts 0 rows and the value survives.
- `migration_leaves_bootstrap_keyspace_bare` — as above.

## The rest

**`[secrets] seed` now warns.** It was on the *don't-warn* side of the plaintext-backend check, so the one deployment that most needed the warning — master seed inlined in `config.toml`, from which both the storage key and the sealed JWT key are re-derivable — was the only one that never got it. I warn rather than refuse: mounting `config.toml` from a Kubernetes Secret is a legitimate pattern, and refusing would break it. Worth a maintainer call if you'd rather it be fatal.

**A missing JWT fingerprint row self-heals.** By the time it is checked, the AES-GCM tag has already authenticated the ciphertext under the storage key, so the fingerprint adds no assurance — but treating its absence as fatal handed a denial of service to anyone who could delete a row without being able to forge a ciphertext, because the documented remedy ("clear both entries and regenerate") invalidates every live session. The row is now rewritten from the authenticated key, with a warning. **Mismatch stays fatal** — that is two sources disagreeing, and a human should decide. `JwtKeyError::FingerprintMissing` is gone; its test now asserts the key survives.

**`load_storage_key_for_cli` errors propagate.** `Ok(None)` means "hardened is off"; `Err` means "could not reach the secret store". Collapsing both to `None` opened bare handles against an encrypted store. The `unseal` path now exits with the cause. `status` still degrades — it is a diagnostic and already tolerates a locked store — but prints a note instead of silently reporting confident nonsense about encrypted keyspaces.

**Per-VTA `storage_key_salt`.** `vta setup` mints a random 128-bit salt. To be straight about this one: it is **not** a fix for a break. An HKDF salt need not be secret and the IKM is already a per-VTA high-entropy seed, so a shared salt never let one VTA derive another's key. But the field is documented as a permanent per-VTA constant, so it should actually be one. Configs that omit the field keep the legacy `"vta-storage-v1"`, so **existing stores are unaffected** — that constant is now commented as a compatibility value that must never change, not a recommended default. The generator lives in `vta-service`, not `vta-config`, to avoid putting a `rand` dependency on a leaf config crate.

**Dropped the `(PoC)` comment** from the daemon boot path.

## Sync and versions

`vta-tee` → **0.1.3** and `vta-sdk` → **0.20.9**; #836 and #837 had taken 0.1.2 and 0.20.8 while this branch sat, so both were downgrades-in-waiting.

The tightened changelog guard (#836) then caught four bumps in this PR with no entry at all — `vta-backup`, `vta-keys`, `vta-policy`, `vta-tee`, all re-pinning to the breaking `vta-support` 0.2. Added a combined entry. That is the guard doing exactly the job it was tightened for, on the first PR after it landed.

## Verification

- `cargo test -p vta-service --lib` — **713 passed, 0 failed** (18 in `hardened_bootstrap`)
- `cargo clippy -p vta-service -p vta-config --all-targets` — clean
- `cargo fmt --check` — clean
- both release guards pass

One thing I did **not** touch: `cargo check -p vta-service --no-default-features --features rest,didcomm` fails on `mediator_registry` in `trust_tasks/consent.rs` and `consent_request.rs`. That field is `#[cfg(feature = "webvh")]` but is referenced unconditionally, so it is a pre-existing broken feature combination on `main`, unrelated to this PR. Flagging it rather than folding an unrelated fix in here.

stormer78:
Both done — pushed as `c8c00209`.

## The JWT key is now an ordinary encrypted row

You were right that this simplifies things, and it turns out to strengthen them slightly too.

The key was sealed with a bespoke AES-GCM blob and parked in the **unencrypted** `bootstrap` keyspace, with a SHA-256 fingerprint row beside it to detect tampering. Two encryption schemes where one would do — and the bespoke one was the weaker of the two: **its ciphertext carried no associated data**, so an envelope copied to another row would still have opened. `VAE1` binds every value to its `(keyspace, key)` location.

It now lives in the encrypted `KEYS` keyspace at `hardened:jwt_key` as a plain `VAE1` row. What that removes:

- the hand-rolled `aes_gcm_seal` path (the opener survives, marked legacy-only, for the one-time import)
- the fingerprint row, `jwt_key_fingerprint`, and the self-heal logic I added last round — all of it existed to partly cover what the AEAD tag now covers properly
- two of the four `JwtKeyError` variants; a wrong salt or seed is simply a decrypt failure on read

`jwt_key_row_is_bound_to_its_location` pins the gain: it lifts the envelope, writes it to a second row, and asserts it no longer decrypts.

**Existing deployments migrate rather than rotate.** The legacy sealed row is opened, rewritten into `KEYS`, and both legacy rows deleted — carrying the key across unchanged, so live sessions survive. A legacy row that will not open is an error, not a silent regeneration; regenerating would invalidate every session without saying so (`undecryptable_legacy_row_errors`).

One ordering constraint this created, now owned by `load_boot_secrets` rather than left to the caller: the key sits behind an encrypted handle, so `migrate_store_to_encrypted` **must** run before it is read, or the read fails on the first unconverted row. The function does seed → migrate → key, in that order, and says why.

`rotate-jwt` now needs the storage key (the row is encrypted), so it derives it up front — which means a wrong `storage_key_salt` or an unreachable secret store fails there with a clear cause instead of at the next daemon start. It also clears any stale legacy rows, so rotating can never leave one behind for a later boot to "carry over" as though it were current.

`bootstrap` stays excluded from migration, and the test that pins it now covers the two real reasons: TEE's `tee:did_log`, which the parent-side proxy reads with no storage key, and the legacy sealed row the import still has to open.

## The `didcomm`-without-`webvh` break

Root cause: `AppState::mediator_registry` is `webvh`-gated, but `PendingResponse`'s module needs only `didcomm`. Three trust-task call sites buffered into the field from inside `#[cfg(feature = "didcomm")]` blocks — `consent.rs`, `consent_request.rs` (×2), `step_up.rs`. Naming the type compiled fine; the field simply was not there. Every other reference in the crate is `webvh`-gated, so these were the outliers, not the gate.

Gated them on `webvh`. The `send_guaranteed` calls sitting next to them stay on `didcomm` — those are the delivery-critical path, while the registry buffer is a fast-path optimisation for a listener that is already attached, so dropping it without `webvh` costs latency, not delivery. Two follow-on warnings fixed honestly rather than blanket-allowed: `CONSENT_APPROVE_REQUEST_TYPE` is now gated `all(didcomm, webvh)` to match its only consumer, and the `WebvhPathCounter` re-export keeps its unconditional `pub use` (it is the type of a non-gated `Guards` field, so it belongs in the public surface either way) with a scoped `allow` and a note.

**CI never built this combination** — the `tee + webvh` step pulls `webvh` in via `tee`, which is exactly why it rotted unnoticed. Added a `vta-service didcomm without webvh` step to the combos job so it cannot regress.

## Verification

- `cargo test -p vta-service --lib` — **709 passed, 0 failed** (14 in `hardened_bootstrap`)
- `cargo clippy -p vta-service --all-targets -- -D warnings` — clean
- feature combos `rest,didcomm` / `rest,didcomm,webvh` / `tee,rest,didcomm` — all build
- `cargo check --workspace --all-targets`, `cargo fmt --check`, both release guards — clean

Note the commit is marked `!`: `hardened_bootstrap`'s public surface changed (`HARDENED_JWT_CT_KEY` → `LEGACY_JWT_CT_KEY`, `aes_gcm_seal` gone, `load_or_generate_jwt_key` and `load_boot_secrets` re-signed). `vta-service` was already going to 0.13.0, so the version covers it.

One thing I left alone: `--features rest` on its own still emits unused-import warnings in files unrelated to this PR. Pre-existing, and I did not want to fold an unrelated cleanup into a security review.

stormer78:
Fixed the `rest`-only warnings too — pushed as `a6d60c3b`.

All six traced to consumers behind `#[cfg(feature = "didcomm")]`, so each is gated to match rather than blanket-allowed:

| Item | File |
|---|---|
| `vta_sdk::acl_setup` | `server.rs` — only the DIDComm listener wires the client ACL on connect |
| `std::time::Duration` | `consent_request.rs`, `step_up.rs` — only the DIDComm sends bound their delivery window |
| `CONSENT_PUSH_DELIVER_BY_SECS`, `TASK_CONSENT_GRANTED_0_1` | `consent_request.rs` |

`flush_queues` is the one exception: it's a parameter of `server::run`, and callers pass it unconditionally, so removing it would churn the signature for no gain. It keeps a scoped `allow(unused_variables)` and a note that a REST-only build has no queues to flush.

Added `vta-service rest only` to the CI combos job as a `clippy -- -D warnings` step, next to the `didcomm without webvh` one from the previous commit. Neither combination was built by anything, which is why both rotted — the existing `tee` step pulls `webvh` in.

Verified across the matrix, all clean with zero warnings:

```
rest                     clean
rest,didcomm             clean
rest,didcomm,webvh       clean
tee,rest,didcomm         clean
rest,didcomm,tsp         clean
```

Plus `cargo clippy -p vta-service --all-targets -- -D warnings`, `cargo test -p vta-service --lib` (**709 passed**), `cargo check --workspace --all-targets`, and `cargo fmt --check` — all clean. Changelog notes both build fixes as separate from the hardened-configuration work, since they're independent of it.

stormer78:
Kept them as warnings then, and sharpened the wording — pushed as `7d5cb672`.

Both now carry the `SECURITY:` prefix the neighbouring `VTA_AUTH_JWT_SIGNING_KEY` warning in the same block already uses, and state the capability directly rather than characterising it:

```
SECURITY: hardened.enabled = true but the master seed is inlined in config.toml
via [secrets] seed. The storage-encryption key and the JWT signing key are both
derived from that seed, so anyone who can read config.toml can re-derive both,
decrypt the whole store and forge any token — the exact capability hardened
configuration exists to remove. At-rest encryption is providing no protection in
this configuration. Move the seed to a real secret-store backend (OS keyring,
aws-secrets, gcp-secrets, azure-secrets, vault, k8s) before relying on it.
```

The plaintext-file fallback gets the matching treatment, and additionally notes that the seed file sits in the same data directory as the store it is meant to protect.

My earlier phrasing — "reduces hardened configuration to obfuscation" — was accurate but read as a downgrade rather than as *this provides no protection*, which is what it actually is.

The docs now also record **why** these are warnings and not a hard refusal: `config.toml` can legitimately be mounted from a Kubernetes Secret, where the seed is not on a disk an attacker reaches. Stating the reason lets an operator decide whether the warning applies to them, instead of learning to ignore it.

`cargo test -p vta-service --lib` 709 passed, `clippy --all-targets -D warnings` clean, fmt and changelog guard clean.

PR #836, opened 2026-07-27
chore(tee): one home for the TEE store keys, and a changelog guard that can fail

stormer78:
Three small cross-cutting fixes that surfaced while reviewing the TEE / Nitro batch (#819, #824, #826, #827, #832). None of them belongs to any one of those PRs, so they are collected here rather than bolted onto whichever happened to touch the file.

Independent of each other and of the PRs that surfaced them — mergeable in any order.

## 1. `tee:did_log` / `tee:vta_did` get one home

The DID-log store key was a bare string literal in **four** places across three crates:

| Site | Role |
|---|---|
| `vta-tee/src/did_autogen.rs` (×2) | writes it to the `KEYS` (encrypted) and `BOOTSTRAP` (plain) keyspaces |
| `vta-service/src/routes/attestation.rs` | reads it to serve `GET /attestation/did-log` |
| `deploy/nitro/enclave-proxy/src/storage.rs` (×2) | parent-side reader that writes `did.jsonl` to disk |

Renaming the key would have broken the attestation route silently — nothing links the writer to the reader.

`vta_tee::did_autogen` now exports `DID_LOG_STORE_KEY` and `VTA_DID_STORE_KEY`, and every in-workspace reader takes the constant. `enclave-proxy` deliberately **cannot**: it is a separate cargo project running on the *parent* and does not link enclave code. It keeps the literal, now with a comment naming the canonical definition — an honest cross-reference rather than a fake abstraction.

Note for #827: `vta-service/src/tee_webvh.rs` introduces a fifth copy, with a doc comment justifying it as "so this bridge doesn't pull `vta-tee` into `vta-service`'s dependency graph". That rationale doesn't hold — the module is `#[cfg(feature = "tee")]` and `lib.rs` already does `pub use vta_tee as tee;` under the same feature. Whichever of these lands second should use the constant.

## 2. `parent-proxy.sh` bound the DID resolver to the IMDS port

It defaulted `VSOCK_RESOLVER_PORT` to **5400**. On the enclave side 5400 is the IMDS credential proxy and the resolver is **5600** — see `enclave-entrypoint.sh:39-40` and `enclave-proxy/src/main.rs:54,62`. A VTA brought up with this script would have had its resolver leg wired into the credential channel.

The script is also socat-only and covers four of the seven vsock channels — no storage proxy (5500), no log forwarding (5700) — so a VTA deployed with it has no persistent store and no enclave logs on the parent. The header now says that plainly and points at the Rust `enclave-proxy` as the canonical parent implementation, so nobody reaches for the socat path for a real deployment.

Surfaced by #832, which documents all seven channels correctly.

## 3. `check-changelogs.sh` could not fail the case it was written for

The guard matched the version **string anywhere in the file**, not a `<crate> <version>` entry. With a dozen subsystem crates all sitting in the 0.1.x range, patch numbers collide constantly — so `vta-tee` 0.1.0 → 0.1.1 (#819) **passed with no `vta-tee` entry at all**, because `0.1.1` already appeared for `vta-audit`, `vta-backup`, `vta-vault` and `vta-webvh`. It is now on `main` as an undocumented published version; this PR backfills the entry.

A release guard that green-lights an undocumented release is worse than no guard, because it is trusted.

The match now requires the crate name immediately before the version, accepting both heading shapes actually in use:

```
### vta-service 0.12.43 — summary
### vti-secrets 0.1.7 / vta-config 0.1.1 — summary
- `vta-audit` 0.1.1
```

The right-hand token boundary is unchanged, so a `0.1.30` entry still cannot satisfy a bump to `0.1.3`.

**Validation** — replayed the tightened guard over the last 25 commits of `main`, running it at each commit with `BASE=<commit>^`:

- **16** commits carrying version bumps still pass — no false positives.
- **1** new catch: `63539989` (#819), `MISSING vta-tee: 0.1.0 -> 0.1.1`. That is the bug, not a regression.

## Verification

- `cargo check -p vta-service --no-default-features --features tee,rest,didcomm` — clean
- `cargo check` in `deploy/nitro/enclave-proxy` — clean
- both release guards pass against `origin/main`

## Versions

`vta-tee` 0.1.1 → **0.1.2**, `vta-service` 0.12.42 → **0.12.45**. The gap is deliberate: #824 claims 0.12.43 and #827 claims 0.12.44, so this stays out of their way regardless of merge order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,57 @@ spec change.
 
 Unblocks the member-side half of OpenVTC/openvtc#185.
 
+### vta-tee 0.1.2 / vta-service 0.12.46 — the TEE store keys get one home, and the changelog guard stops passing vacuously
+
+Three small cross-cutting fixes that kept surfacing while reviewing the TEE and
+Nitro PRs (#819, #824, #826, #827, #832). None of them belongs to any one of
+those changes.
+
+**`tee:did_log` and `tee:vta_did` are now exported constants.** The DID log key
+was spelled as a bare string literal in four places: written twice in
+`vta_tee::did_autogen` (to the `KEYS` and `BOOTSTRAP` keyspaces), read by
+`vta_service::routes::attestation::did_log` to serve `GET
+/attestation/did-log`, re-declared a fifth time by `vta_service::tee_webvh`
+(#827), and read twice more by the parent-side `deploy/nitro/enclave-proxy`.
+Renaming it would have broken the readers silently. `did_autogen` now exports
+`DID_LOG_STORE_KEY` and `VTA_DID_STORE_KEY` and every in-workspace reader takes
+the constant. (`tee_webvh`'s copy was justified as avoiding a `vta-tee`
+dependency, but that module is `cfg(tee)` and `crate::tee` is already
+`pub use vta_tee` under the same feature — there was no dependency to avoid.) `enclave-proxy` is a separate
+cargo project on the parent that deliberately does not link enclave code, so it
+keeps the literal — now with a comment naming the canonical definition.
+
+**`deploy/nitro/parent-proxy.sh` bound the DID resolver to the IMDS port.** It
+defaulted `VSOCK_RESOLVER_PORT` to 5400, but 5400 is the IMDS credential proxy
+on the enclave side and the resolver is 5600 (`enclave-entrypoint.sh`,
+`enclave-proxy/src/main.rs`). The script is also socat-only and serves four of
+the seven vsock channels — no storage (5500) and no log forwarding (5700) — so
+its header now says that outright and points at `enclave-proxy` as the
+canonical parent implementation.
+
+**`scripts/check-changelogs.sh` matched the version string anywhere in the
+file, not a `<crate> <version>` entry.** With a dozen subsystem crates sharing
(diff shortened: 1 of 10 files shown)


stormer78:
Rebased onto `main` after #827 merged, and extended to cover it.

#827 landed the fifth copy of the `tee:did_log` literal in `vta-service/src/tee_webvh.rs`, so this PR now replaces that one too. Its stated rationale — *"Mirrored here (rather than imported) so this bridge doesn't pull `vta-tee` into `vta-service`'s dependency graph"* — doesn't hold: the module is `#[cfg(feature = "tee")]` and `lib.rs` already does `pub use vta_tee as tee;` under that same feature. There was no dependency to avoid, only a literal to keep in sync.

After this change, `grep -rn '"tee:did_log"|"tee:vta_did"'` across the repo returns exactly four hits: the two canonical `pub const` definitions in `vta_tee::did_autogen`, and the two in `deploy/nitro/enclave-proxy` — the parent-side proxy that genuinely cannot link enclave code, now carrying comments naming the canonical definition.

Version renumbered `vta-service` 0.12.45 → **0.12.46**, since #827 took 0.12.44 and #824 has been renumbered to 0.12.45. `vta-tee` stays at 0.1.2.

PR #837, opened 2026-07-27
feat(credential-exchange)!: bind published URIs, and guard the whole spec authority (#821)

### LINKED ISSUE #821: credential-exchange Trust Tasks: specs live only in this repo, and three bound URIs have no spec at all
Split out of #709, which listed `trust-tasks/credential-exchange/{issue,offer,present,query,request}/1.0` as directories "in neither the manifest nor any binding — decide whether those are published or deleted".

Half of that premise no longer holds, and what remains is a different problem.

## Current state

The five directories carry **conformant public-registry IDs** and are **bound live**:

```
https://trusttasks.org/spec/credential-exchange/{issue,offer,present,query,request}/1.0
```

bound by `vta_sdk::protocols::credential_exchange` and dispatched by `vtc-service`. They are absent from `trust-tasks/index.json` **correctly** — that manifest is scoped `org: openvtc, domain: vtc`, and these are cross-domain canonical slugs. Deleting them would delete the only copy of a live wire contract.

## The two real gaps

**1. They are not in the upstream registry.** `trust-tasks-rs` 0.2.39 publishes no `spec/credential-exchange/*` slug — checked against `schema_index.rs` and `src/specs/`. So the specs exist as files in this repo and nowhere a consumer can resolve them, despite claiming a `trusttasks.org` ID. That is the same authority-claim problem #710 fixed for `spec/vtc/*`, one family later.

**2. Three bound URIs have no spec on disk at all:**

```
https://trusttasks.org/spec/credential-exchange/pending-list/1.0
https://trusttasks.org/spec/credential-exchange/pending-approve/1.0
https://trusttasks.org/spec/credential-exchange/pending-deny/1.0
```

`vta-sdk/src/protocols/credential_exchange.rs` defines `PENDING_LIST` / `PENDING_APPROVE` / `PENDING_DENY`, but `trust-tasks/credential-exchange/` has only the five `{issue,offer,present,query,request}` directories. No spec, no schema, not in any manifest.

## Why nothing catches this

`every_bound_vtc_task_exists_in_the_registry` only scans the `spec/vtc/` prefix. `spec/credential-exchange/` bindings pass through every census test untouched — which is how three specless URIs went unnoticed.

## Work

- Author the five existing specs upstream in `dtgwg-trust-tasks-tf`, or move them to an authority OpenVTC controls (SPEC 6.5 — the `trusttasks.org` domain is reserved for registry-published specs, the exact rule #710 turned on).
- Author the three `pending-*` specs, or retire the constants if the surface is not meant to be a published Trust Task.
- Widen the census guard so `spec/credential-exchange/*` is checked the way `spec/vtc/*` is, rather than being invisible to it.

Not urgent — this is a published-authority-claim and coverage gap, not a runtime defect. The bindings work; they just reference specs no consumer can fetch.

stormer78:
Closes #821.

The eight `credential-exchange` Trust Tasks now bind URIs the registry actually publishes — `trust-tasks-rs` 0.2.41, authored in [trust-tasks-tf#148](https://github.com/trustoverip/dtgwg-trust-tasks-tf/pull/148).

Five of them previously existed **only** as files in this repo while claiming a `trusttasks.org` ID no consumer could resolve. The other three — `pending-list`, `pending-approve`, `pending-deny` — had no spec anywhere at all.

## Breaking wire changes on all eight

- Renumbered `1.0` → `0.1`, matching how the rest of the registry versions.
- The operator surface is nested as `pending/{list,approve,deny}` rather than flat siblings of the protocol tasks. The family now reads as what it is: a party-to-party wire protocol, plus an operator surface over the consent backlog it generates.
- **`PendingPresentationSummary` and `RequestedCredentialSummary` serialize camelCase** — `verifier_did` → `verifierDid`, `created_at` → `createdAt`, `credential_query_id` → `credentialQueryId`. Every member of those two is ours, so the registry's casing convention applies.

Two things deliberately *not* changed, because they look like the same class and aren't:

- The **OID4VCI / OID4VP bodies keep snake_case**. `credential_offer`, `dcql_query`, `vp_token` are those specifications' own field names. Renaming them would be wrong, not consistent.
- The **stored deferral record keeps snake_case**. It is fjall-persisted internal state, not a wire type; renaming it would need a data migration for no benefit. The test fixture that builds one stays as it was, which is why only the response assertions moved.

No external consumers of these constants exist, so the rename is free. `trust-tasks/credential-exchange/` is deleted — the specs live upstream now, and a second copy is how the two drift.

## The guard that should have caught this

`every_bound_vtc_task_exists_in_the_registry` only ever checked the `spec/vtc/` prefix. That is precisely why eight URIs on the same authority went unverified for as long as they did.

It is now `every_bound_canonical_task_exists_in_the_registry` and checks **every** `https://trusttasks.org/spec/` URI the workspace binds. A per-family prefix is the wrong shape for the assertion: it defends the family someone remembered to name and silently exempts the next one. The claim being tested is about the *authority* — binding a `trusttasks.org/spec/` URI asserts the registry serves it.

Widening it required separating task URIs from other strings sharing the prefix, done by shape: a Type URI ends in a `MAJOR.MINOR` segment (SPEC §6.1). One rule, two classes excluded — family prefixes used to build or assert URIs (`vta-sdk`'s `ALLOWED_PREFIXES`), and shared-schema `$id`s, which are components rather than tasks.

## It immediately found 67 more

All pre-existing. **None introduced by this PR**, and none a runtime defect — the bindings work, they just reference specs no consumer can fetch.

| Family | Unpublished | What |
|---|---:|---|
| `spec/vta/` | 55 | The bulk of the VTA's own Trust Task surface at `1.0` — keys, contexts, backup, seeds, acl, audit, attestation, config, discovery, management, provision-integration, webvh dids/servers/agent-name. The registry publishes 22 `vta/*` tasks; the workspace binds 77. |
| `spec/vault/` | 12 | The vault + credential-store archival lifecycle (#540), authored as local "openvtc 0.1 extensions" and never taken upstream. |

Plus `spec/trust-task-error/0.1`, which is a permanent and legitimate exception — the framework's error *envelope* is a response type, deliberately absent from the task index.

These are recorded in `UNPUBLISHED_CANONICAL_OK` as family-level exceptions **with asserted counts**. Listing 68 URIs individually would be unreadable; excluding a family without a count would let the next unpublished task in it pass unnoticed, which is the failure the assertion exists to prevent. Pinning the number means the debt can only shrink — publish some upstream and the count drops, add a new unpublished one and the test fails with a message saying which direction it moved and why raising the number is the wrong fix.

I'd suggest tracking the `vta/*` and `vault/*` rows as their own issue rather than folding them in here; they are a much larger piece of work than #821 was, and the same decision (author upstream vs. move to an authority we control) has to be made per family.

## Verification

- `cargo fmt --check` clean
- `cargo clippy --workspace` clean (as CI runs it)
- `cargo test -p vta-sdk -p vta-service -p vtc-service -p vti-common`: 88 test binaries, exit 0, zero failures (built with the admin UI)
- The widened census resolves all eight new URIs against the published `trust_tasks_rs` 0.2.41 rather than a hand-maintained list

`vta-sdk` 0.20.8, `vta-service` 0.12.43, `vtc-service` 0.11.37.


stormer78:
Resolved the conflicts and re-synced with `main` (now `1f3b6194`). Pushed as `05f91c33`.

Three merges landed underneath this branch while it sat (#824, #827, #838), and two of them collided on versions:

| Crate | Was | Now | Why |
|---|---|---|---|
| `vta-service` | 0.12.43 | **0.12.47** | main moved to 0.12.45 via #824/#827; 0.12.46 is claimed by #836, which is green and ahead in the queue |
| `vtc-service` | 0.11.37 | **0.11.38** | #838 took 0.11.37 |
| `vta-sdk` | 0.20.8 | 0.20.8 | unchanged — main is still on 0.20.7 |

Both were genuine downgrades-in-waiting, not just staleness: `check-version-bumps.sh` would have failed the PR once `main` overtook it. Changelog heading renumbered to match, and #838's entry kept above this one.

Verified after resolution:
- diff vs `main` is **20 files, +261/−434** — identical to the PR's original stats, so nothing was lost in the merges
- `cargo check --workspace --all-targets` clean
- `cargo fmt --check` clean
- both release guards pass

I did not touch the substance of the change — only the version numbers, the changelog heading, and the lockfile.

Heads-up on ordering: **#836 should merge before this one** (it holds `vta-service` 0.12.46). If this merges first instead, #836 needs one more renumber, which is mine to absorb — so no need to hold this on my account, just ping me.

PR #843, opened 2026-07-28
Vta mcp acl

raghavendrram:
Adds acl-setup to the vta-mcp cargo config to allow managing ACLs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ## Unreleased
 
+### vta-service 0.13.6 — `contexts/create` is an idempotent no-op when the context already exists
+
+`create_context` checked `require_super_admin()` for a top-level context
+*before* the existence check, so a context-admin caller re-running a
+provisioning flow against an already-existing top-level context was rejected
+with `super admin required`. This broke fresh-instance domain wake: the console
+pre-creates the placeholder context offline with a super-admin credential, then
+at wake the operator (context-admin only) re-issues `contexts/create`, which
+must fold onto the existing context rather than demand elevation.
+
+The existence short-circuit now runs *before* the authz gate and returns
+idempotent **success** (the existing context's result body) for a caller that
+administers it (`require_context`). Previously an existing context returned
+`Conflict`; but the trust-task framework has no standard `conflict` reject code,
+so over DIDComm that surfaced as an opaque `taskFailed` the SDK could not fold
+back into `VtaError::Conflict` — meaning the online provisioner's
+already-exists-is-success handling never fired. Returning success makes
+idempotent re-provisioning behave identically across REST and DIDComm. The
+super-admin gate is unchanged for actually creating a *new* top-level context;
+sub-context creation still requires admin-of-parent. Folding onto an existing
+context now needs only the right to use it, and no longer leaks a context's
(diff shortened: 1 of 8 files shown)


stormer78:
Superseded by three focused PRs, all rebased onto current `main` (this branch was 86 commits behind). @raghavendrram is credited as co-author on every commit — thanks for the original work and the diagnosis behind it, particularly the DID-routing mediator analysis.

| Replacement | Content |
|---|---|
| #951 | `vta-mcp` enables the SDK `acl-setup` feature |
| #952 | `create-did-peer --mediator-did` (+ the tests this lacked, and clap-enforced mutual exclusion) |
| #953 | `create-did-key` external key import, reworked to a **file** input with zeroization |

## What did not carry over, and why

**The `contexts/create` idempotency change.** The diagnosis in the changelog entry is correct and still true — `trust_tasks/helpers.rs:104` collapses `AppError::Conflict` into `taskFailed`, so the SDK cannot fold it back into `VtaError::Conflict` over the Trust-Task transport. But making the server return success for everyone treats the symptom:

- The **DIDComm protocol-message** path already handles this properly (`messaging/handlers.rs:55` → `ProblemReport::conflict`, `vta-sdk/src/error.rs:241` → `VtaError::Conflict`). Only the Trust-Task reject mapping is missing it. That is the actual fix, and it is one match arm.
- It silently breaks a documented operator path: `vta-cli-common/src/commands/contexts.rs:270` matches `Err(VtaError::Conflict(_))` to print the "already exists — use `acl create` instead" guidance. With this change, `pnm contexts create --id X --admin-did D` against an existing `X` would print **"Context created:"** (untrue) and then grant the ACL anyway. No test asserts the conflict, so CI would not have caught it.
- The capability already exists: `operations/provision_integration/preconditions.rs:136` `ensure_target_context_or_create` is the sanctioned create-if-missing path (exists → `Ok(false)`, no authz check on that branch), already wired through REST, DIDComm and the offline CLI and surfaced as `context_created: bool`.
- Semantics would become timing-dependent: the `store_new_context` race-loser at `operations/contexts.rs:126` still returns `Conflict`, so sequential re-create → success, concurrent re-create → conflict.

If the provisioner genuinely needs create-or-fold, the two clean options are the `conflict` reject code, or an explicit `if_exists: fold` flag on `contexts/create` — an opt-in the caller asks for, rather than a semantics change for every caller. Happy to open either.

**The `Dockerfile`.** Currently correct — I diffed all 26 workspace members against it, zero drift. Held back because nothing in `.github/workflows/` builds it, so it rots the day someone adds a crate, and it hand-maintains the member list four times in one file (`Dockerfile.nitro` does the same, so this would make eight copies). It also builds without `--locked`. Worth landing bundled with a CI build job and `--locked`; happy to do that as a fourth PR if wanted.

**The `CHANGELOG.md` edit and the `vta-service` version bump.** Both removed as a matter of process — #938 moved changelog generation to git-cliff from conventional commits, root `CHANGELOG.md` is frozen, and versions are assigned by the release-plz Release PR. The changelog prose was good; it now lives in the commit bodies, which is where git-cliff reads from. (The bump was also stale: it claimed `0.13.6` against a `main` now at `0.14.37`.)

Issue #517, opened 2026-06-17
vault/upsert/0.2 rejects camelCase VaultSecret.kind — inner sealed-secret discriminator still kebab-case (Framework 0.2 migration gap)


stormer78:
## Summary

`vault/upsert/0.2` rejects any `VaultSecret` whose `kind` discriminator is a **multi-word lowerCamelCase** value (`didSelfIssued`, `oauthTokens`, `didcommPeer`, `bearerToken`, `sshKey`). The deployed VTA unseals the inner secret against a **kebab-case**-tagged Rust enum, but the canonical **0.2** schema (and `SPEC.md §4.10`) require the `kind` discriminator to be **lowerCamelCase**. Single-word kinds (`password`, `passkey`, `custom`) collide harmlessly, which is why this stayed hidden until the first `did-self-issued` entry.

Verified against `origin/main` @ `66a9e6b` (fetched 2026-06-17).

## Reproduction

Creating a `did-self-issued` (VTA SIOP) vault entry from the browser plugin, which seals a spec-conformant `{"kind":"didSelfIssued", ...}` cleartext into the authcrypt JWE:

```
vault/upsert/0.2: /api/trust-tasks failed (422): task failed:
vault/upsert:sealed_secret_invalid — cleartext not a VaultSecret:
unknown variant `didSelfIssued`, expected one of `password`, `passkey`,
`oauth-tokens`, `did-self-issued`, `didcomm-peer`, `bearer-token`,
`ssh-key`, `custom`
```

## Root cause

The inner sealed secret rides inside the opaque authcrypt JWE and is deserialized directly against the kebab-tagged enum:

- `vta-service/src/operations/vault/upsert.rs:58` — `serde_json::from_value::<VaultSecret>(msg.body)`
- `vti-common/src/vault/mod.rs:220-221` — `#[serde(tag = "kind", rename_all = "kebab-case")] pub enum VaultSecret`
- `vti-common/src/vault/mod.rs:141` — `SecretKind` is likewise `#[serde(rename_all = "kebab-case")]`

The variant **fields** were already migrated to camelCase (commit `76287a5`, "camelCase serde rename on VaultSecret variants to match spec"), but the **discriminator** was left kebab-case — a half-completed migration. The enum doc-comment still pins the old `vault/_shared/0.1/vault-secret` schema.

[...]

Issue #829, opened 2026-07-27
VtaClient::sign() is the only signing entry point that cannot reach TSP


stormer78:
`keys.rs` has three signing entry points. Two of them route through the Trust-Task bridge and therefore work on every transport; the primary one does not.

| Method | Routes via | Works on TSP |
|---|---|---|
| `derive_and_sign` | `rpc_tt(TASK_KEYS_DERIVE_AND_SIGN_1_0, …)` (`vta-sdk/src/client/keys.rs:149`) | yes |
| `derive_and_sign_document` | `rpc_tt(TASK_KEYS_DERIVE_AND_SIGN_DOCUMENT_1_0, …)` (`keys.rs:176`) | yes |
| **`sign`** | `rpc(key_management::SIGN_REQUEST, …)` (`keys.rs:105`) | **no** |

`rpc` and `rpc_void` both end in `Transport::Tsp { .. } => Err(unsupported_over_tsp(msg_type))` (`client/mod.rs:1140`, `1172`), so a TSP client calling the raw-bytes signing oracle gets `UnsupportedTransport` — while the same client can derive-and-sign perfectly well.

## Everything needed for the fix already exists

- The Trust Task is defined and handled: `spec/vta/keys/sign/1.0` / `TASK_KEYS_SIGN_1_0` (`trust_tasks.rs:301`), handler `handle_sign` (`vta-service/src/trust_tasks/keys.rs:168`), reaching the identical `operations::keys::sign_payload` spine with the same `{ key_id, payload, algorithm }` body and the same base64url conventions.
- The bridge exists and is the documented pattern for exactly this: `rpc_tt` keeps the dedicated REST route while the DIDComm leg dispatches the Trust Task, and its TSP arm notes the surface "needed no per-operation work to reach TSP".

So the change looks like moving `sign()` from `rpc` to `rpc_tt` with `TASK_KEYS_SIGN_1_0`, keeping `POST /keys/{key_id}/sign` as the REST leg — the same shape `derive_and_sign` beside it already has. The three signing paths then behave identically across transports instead of one being the odd one out.

## Why it matters beyond tidiness

[...]

Issue #818, opened 2026-07-27
ACL grants cannot scope a caller to specific key ids (only to a whole context)


stormer78:
Split out of the design review in #817 (`docs/05-design-notes/multi-tenant-signing.md`, R2). Stands on its own merits — this is a least-privilege gap independent of the Cierge deployment that surfaced it.

## The gap

The **resource** dimension of an ACL grant is expressible only as a *context*.

`AclEntry.capabilities` is a flat `Vec<Capability>`, and `Capability::Sign` is unparameterized (`vti-common/src/acl/mod.rs:122-145`). So this cannot be said:

> this caller may invoke the signing oracle on **exactly key K**

The only way to scope a caller to one key today is to give that key **a context of its own**. That works — it is what the design note recommends for multi-tenant signers — but it forces context topology to encode per-key authorization, which is a different concern. Any consumer that legitimately needs one key out of a shared context must be handed the whole context.

## Why `signable_keys` is not the answer

`ContextPolicy.signable_keys` (`vta-sdk/src/context_policy.rs:48-50`) already restricts which key ids the signing oracle may be invoked on — but it is **resource-bound by design**: it constrains the key's context for *every* actor uniformly, including super-admin. That property is load-bearing (it is what lets a VTC- or fleet-pushed policy bind every signer, relaxable only through policy CRUD).

It therefore cannot express "*this caller*, this key". The two mechanisms are complementary, not substitutes — this issue asks for the actor-scoped half.

## Proposal

An optional actor-scoped key filter on the ACL entry, **intersected with** — never widening — the existing context scope:

```rust
/// Key ids this entry may invoke the signing oracle on. `None` = every key
/// in `allowed_contexts` (today's behaviour). Intersects with the context
/// scope; it can only narrow, never widen, so an entry naming a key outside
/// its contexts still cannot use it.
#[serde(default, skip_serializing_if = "Option::is_none")]
pub allowed_keys: Option<BTreeSet<String>>,
```

Enforced in `operations::keys::sign_payload` as a fourth gate, **after** the existing `auth.require_context(ctx)` (`vta-service/src/operations/keys.rs:840`) so a caller can never reach a key outside its contexts by naming it here.

Additive and fail-closed: `None` on every existing row preserves behaviour byte-identically, the same way `capabilities` was introduced.

## Why a separate field rather than parameterising `Capability::Sign`

[...]

PR #866, opened 2026-07-29
fix(conformance): every published task's wire shape conforms to its schema, and a sweep that keeps it that way

### LINKED ISSUE #856: acl/update is rejected at the dispatch spine: its request body is not canonical (subject vs did)
## Problem

`acl/update`'s **request** body is not canonical. It emits `did`/`allowedContexts` where `acl/update/0.1` requires `subject`/`scopes`, and it has no nested `stepUp`/`approve` objects:

```
unknown field `did`, expected one of `approve`, `expiresAt`, `ext`,
`label`, `reason`, `scopes`, `stepUp`, `subject`
```

Because `spec/acl/update/0.1` **is** published, `trust_tasks/mod.rs::validate_payload` finds a schema for it and validates against it. A conforming `acl/update` document is therefore rejected at the dispatch spine with `malformed_request`.

## Why nothing caught it

The other two transports carry their own bodies, so neither exercises the canonical shape:

- **REST** `PATCH /acl/{did}` binds its own `UpdateAclRequest` (`routes/acl.rs`).
- **DIDComm** uses the legacy `firstperson.network/protocols/acl-management/1.0/update-acl` message, which has no schema and so is never validated.

Only the **Trust Task** transport hits the canonical schema, and it had no test.

## Origin

#842 folded the ACL family onto canonical `acl/*`. It made the *response* canonical — via the shared `AclEntry` component plus a compatibility layer on the client types — and repointed the URI, but left the request shape as it was.

## Fix

Give `UpdateAclBody` the same compatibility-layer treatment #842 gave the response: keep the historical Rust field names, map them to the canonical wire with serde renames.

| Rust field | Canonical wire |
|---|---|
| `did` | `subject` |
| `allowed_contexts` | `scopes` |
| `step_up_approver`, `step_up_require` | nested `stepUp` object |
| `approve_scope` | nested `approve` object |
| — | `expiresAt`, `reason` (not currently carried) |

The nested `stepUp`/`approve` shapes already exist as canonical components in `vta_sdk::protocols::acl_management::entry`, added by #842 for the response — reuse them rather than defining second versions.

Verify by round-tripping the serialized body through `trust_tasks_rs::specs::acl::update::v0_1::Payload`, which carries `deny_unknown_fields` and the spec's required set.

## Notes

- Currently pinned by a comment in `vta-service/src/trust_tasks/acl.rs` tests (added in #855) so it cannot be mistaken for conformant, but **not asserted** — the test would fail for this unrelated reason.
- Found in #855 only because a conformance assertion was passing for the *wrong* reason; see #857 for the systemic version.
- Related but distinct from #854, which covers the opposite direction (implemented URIs with **no** registry spec). This one is a URI that *has* a spec and does not conform to it.

### LINKED ISSUE #857: No check that a canonically-bound task's payload conforms to its published schema
## Problem

Nothing checks that a task bound to a **published** canonical URI actually conforms to that URI's schema. We have two parity checks today and this is the missing third:

| Check | Exists? | Covers |
|---|---|---|
| Every SDK constant is dispatched | yes | routing |
| Every dispatched URI has a registry spec | proposed in #854 | coverage |
| **Every specced URI's payload conforms to its schema** | **no** | correctness |

The gap is not theoretical — it already produced a live defect. `acl/update` has been rejected at the dispatch spine since #842 because its request body emits `did` where the schema requires `subject`. See the companion issue.

## Why it hides

- **Requests** are validated at runtime by `validate_payload`, but only on the Trust Task transport. REST and the legacy DIDComm messages carry their own bodies, so a non-conformant canonical body is invisible unless someone exercises that one path.
- **Responses are never validated at runtime at all.** Only a test can catch a non-conformant response.
- Folding a task onto a *published* URI silently turns schema validation on. An unpublished `spec/vta/*` URI has no entry in `trust_tasks_rs::schema_index`, so it dispatches unvalidated; repointing the constant enables `additionalProperties: false` and required-field checks on live traffic, with no compile-time signal.

Only three tasks have ever had a conformance test: `config/*` (#841), `audit/list` (#848), `acl/change-role` (#855). Everything else folded in this programme is unverified.

## Proposal

One sweep test asserting, for every canonically-bound task, that a representative request body deserializes into its generated `trust_tasks_rs::specs::…::Payload`, and the response body into `…::Response`. The generated types carry `deny_unknown_fields` plus the spec's required set, so the round-trip *is* the conformance check:

```rust
let json = serde_json::to_value(&our_body)?;
serde_json::from_value::<canonical::Payload>(json.clone())
    .unwrap_or_else(|e| panic!("not canonical: {e}\n{json:#}"));
```

**Write it so it cannot pass vacuously.** The `acl/update` defect went unnoticed across three PRs because an assertion of the form "adding a forbidden member makes this fail" passed while the body was *already* failing for an unrelated reason. Assert the good case parses, then prove the check has teeth by drifting a member and asserting rejection.

## Scope

Candidates are every URI under `https://trusttasks.org/spec/` that `trust_tasks_rs::schema_index::schema_for` resolves — the same set `vtc-service/tests/trust_task_manifest.rs` already enumerates for the census, so the list can be derived rather than hand-maintained.

Worth running before further work on the fold programme (docs/05-design-notes/canonical-task-reduction.md, phases C–E), since it either clears what has already shipped or finds the rest of this class in one pass.

## Related

- #854 — registry↔implementation drift in the opposite direction (implemented URIs with no spec).
- #840 — the fold programme that produced the bindings this would check.

stormer78:
Closes https://github.com/OpenVTC/verifiable-trust-infrastructure/issues/856
Closes https://github.com/OpenVTC/verifiable-trust-infrastructure/issues/857

## What this does

**#857 — the schema-conformance sweep.** `vta-service/src/trust_tasks/conformance.rs` asserts, for **every** task the dispatcher binds to a published canonical URI, that a representative request body round-trips through the generated `trust_tasks_rs::specs::…::Payload` and the response through `…::Response`. The URI census is derived, never hand-listed: `dispatched_uris()` (macro-generated from the same declarations as the dispatch match arms — vta-service's equivalent of the `vtc-service/tests/trust_task_manifest.rs` source scan) filtered to what `trust_tasks_rs::schema_index::schema_for` resolves. That is **63 URIs** under the default feature set.

Two assertions:
- **Coverage** (`every_published_dispatched_uri_has_a_witness`): witness table ⇄ derived census must agree exactly, both directions.
- **Correctness** (`every_witnessed_task_round_trips_through_its_generated_types`): the good case parses, and — per the issue's non-vacuity requirement — every witness is then drifted (unknown member injected) and the drifted form must be **rejected**, so a witness cannot pass against a schema that accepts anything.

An annotated `KnownDrift` allowlist mechanism exists for structural non-conformance. **It ships empty** — every drift the sweep found is fixed in this PR (below), so there are no follow-up issues from this class today.

**#856 — `acl/update`'s request is canonical.** `UpdateAclBody` keeps its historical Rust field names and maps them to the canonical wire with serde renames (`did`→`subject`, `allowed_contexts`→`scopes`), nesting the step-up / approve-authority members under the **shared** `StepUp`/`Approve` components #842 already added in `vta_sdk::protocols::acl_management::entry` (no second definitions). The previously-uncarried `expiresAt` (RFC 3339 on the wire, epoch in the store) and `reason` (recorded via `audit::record_with_detail`) members are now carried. The comment-only pin in `trust_tasks/acl.rs` (#855) is a real assertion: full round-trip through `specs::acl::update::v0_1::Payload`, plus a teeth check that reintroduces the pre-fix `did` spelling and requires rejection.

## What the sweep caught (all fixed here, not allowlisted)

The first full run found eight more non-conformances beyond #856's known one. All were the trivial class (wire spelling / missing-or-extra members with the data already in hand), so per the stream instruction they are fixed in this PR:

| Task | Drift | Fix |
|---|---|---|
| `acl/list` response | legacy bare `Vec<CreateAclResultBody>` (snake_case rows) | handler now uses the existing canonical wrapper `operations::acl::list_entries` |
| `acl/show` response | legacy flat row | handler now uses `operations::acl::show_by_subject` |
| `acl/change-role` response | bare flat row | wrapped in canonical `{entry}` (`CreateAclResponseBody`) |
| `acl/swap-key` response | bare flat row; `previousSubject` missing | new `SwapKeyResultBody {entry, previousSubject}` |
| `auth/revoke-session` response | empty `{}`; schema requires `revokedCount` | field added (vta-mobile-core already parsed `revokedCount` — the SDK type was the outlier) |
| `device/wipe` response | `wipedAt`/`disabledAt`/`reason` not in the closed schema; `completedAt` missing | emits canonical `{deviceId, scope, completedAt}` |
| `vault/proxy-login` response | cleartext `sessionId`/`expiresAt` at the root of a closed object | moved under vendor-namespaced `ext` (`org.openvtc.vault-session`, following the step-up slice's `org.openvtc.*` convention) |
| `provision/integration` request+response | snake_case / kebab-case emission vs the canonical 0.2 camelCase (`ProvisionSummary`, `AssertionMode`, `vcValiditySeconds`, `createContext`) | emission flipped to camelCase; snake/kebab kept as intake aliases. This reverses #517's "emit conservative" direction deliberately: 0.2 is now the **only** URI the dispatcher binds, and its schema is `additionalProperties: false` |
| `vta/webvh/dids/update` response | snake_case emission (`new_version_id`, …) | `UpdateDidWebvhResultBody` emits camelCase per its published schema; snake aliases kept on intake |

Also fixed in passing: the SDK client's legacy-DIDComm `update_acl` body already sent `subject` (so it could never have matched the old `did` field — that path was already non-functional); its remaining `allowed_contexts` key moves to the canonical `scopes`.

## Scope notes

- The sweep lives in vta-service because that is where the dispatch table (and its macro-generated census) lives. Extending the same pattern to vtc-service's REST-bound tasks is possible but is a separate ~60-URI witness effort against axum route types; not attempted here.
- 0.1↔0.2 dual-accept URIs: `wire_v0_2` edge-transformed 0.2 URIs are not dispatch arms (they are down-converted before dispatch) and keep their own transform tests; the genuinely dual-dispatched `auth/step-up/approve-response/0.1|0.2` has one witness per version with version-distinct `evidence.kind` fixtures.
- Pre-production wire changes (device/wipe, proxy-login, provision, dids/update) are clean canonical fixes rather than compatibility shims, per the programme's direction; intake keeps accepting the legacy spellings via serde aliases wherever a legacy producer could exist.

## Rebase expectation

~8 sibling consolidation streams are currently rebinding URIs. As they merge, the sweep's derived census will change and its **coverage assertion will fail asking for new witnesses — that is the check doing its job**, not a defect of this PR. This already happened once during development: the branch is rebased over #858–#862, and #860's constant-rename pass surfaced immediately in the sweep table (absorbed here) — exactly the behaviour the check exists for. Version bumps / CHANGELOG placement may need the same trivial rebase again as more siblings merge (currently vta-sdk 0.20.19 → **0.20.20**, vta-service 0.13.11 → **0.13.12**).

## Validation

- `cargo test -p vta-service` — full suite green (includes the two sweep tests over 63 URIs and the un-pinned #856 assertion).
- `cargo test -p vta-sdk` — green.
- `cargo test -p vta-mobile-core -p pnm-cli -p cnm-cli -p vta-webvh -p vta-cli-common` — green; `cargo check --tests` green for every vta-sdk dependent crate.
- Local dry-runs of the CI release guards pass: version-bump guard, changelog guard, lockfile self-pins.
- `cargo fmt` clean on touched crates.

## Reviewer checklist

- [ ] `conformance.rs`: the witness table's fixtures faithfully mirror the handlers whose local types are module-private (consent, task-consent, vault) — each is annotated with the type it mirrors.
- [ ] The wire changes in the table above are acceptable as clean fixes (no compatibility shim) given pre-production status; intake aliases retained where a legacy producer could exist.
- [ ] `vault/proxy-login`: agreement that `ext["org.openvtc.vault-session"]` is the right home for the cleartext session hints.
- [ ] `UpdateAclParams` gained `expires_at`/`reason`; REST `PATCH /acl/{did}` deliberately does not carry them (its own body is unchanged).


stormer78:
Follow-up note on the `provision/integration` change, for the reviewer of that hunk specifically:

`request_body_for_version` / `response_body_for_version` are now **inverted** rather than removed. Previously the types serialised the 0.1 snake_case form and these helpers up-cased for a 0.2 destination; now the types serialise the canonical 0.2 form (which is what the published schema requires, and 0.2 is the only URI the dispatcher binds) and the helpers **down-cast for a 0.1 destination**. The signed VP subtree is still never touched in either direction — its `DataIntegrityProof` covers its exact bytes.

So a 0.1 peer still receives exactly what it received before, and a 0.1 producer is still accepted via the serde aliases. The only behavioural change is which form is the default.

stormer78:
Rebase note (the one the PR body predicted, arriving on schedule).

Merged `origin/main` in — picking up #863 and #864. **#864 is exactly the churn case this check exists for**: it folded the twelve `vta/{contexts/,}did-templates/*/1.0` URIs onto the merged six-task 2.0 family, the sweep's derived census dropped them, and the coverage assertion failed asking what happened to those twelve witnesses.

Resolution: the witnesses are **removed, not repointed**. The 2.0 specs are published upstream but not yet in the pinned `trust-tasks-rs` (the same publish-lag `vtc-service/tests/trust_task_manifest.rs` already records at +6), so `schema_for` resolves none of them and they are legitimately out of scope today. The next `trust-tasks-rs` bump puts all six back in the census and the coverage assertion will ask for their witnesses then — with a comment in `conformance.rs` saying so, so whoever does that bump isn't left guessing.

Sweep is now **51 witnessed URIs**, still an empty drift allowlist.

Merged rather than rebased so the push stays a fast-forward (no force). Versions land above main's: vta-sdk **0.20.23**, vta-service **0.13.14**.

stormer78:
## The `Test` job failure is pre-existing on `main`, not from this PR

Evidence, three independent ways:

**1. `main`'s own CI is red with it.** The CI run on `main` at #863 (`f6257a5f`, id 30419616926) and at #864 (`c314ae5a`, id 30421706004) both fail the `Test` job with `test result: FAILED. 21 passed; 21 failed` — the same count this PR shows.

**2. The failing test names are identical.** Diffing the failure list from `main`'s run against this PR's: **zero PR-only failures.** Same 21 tests, same file, same assertion.

**3. A clean worktree at unmodified `origin/main` reproduces it.** Checked out `704bdda6` (newer than this PR's merge base) with no changes at all: `cargo test -p vti-e2e-tests --test client_didcomm` → `21 passed; 21 failed`.

### What's actually broken

All 21 fail with `NotFound("no handler")` from the mock responder in `tests/e2e/tests/client_didcomm.rs`. #861 ("route every trust-task-twinned client method through `rpc_tt`", #829) moved the SDK client methods onto Trust Task envelopes, but the e2e DIDComm mocks still match on the **legacy** message types (`discovery::DISCOVER_CAPABILITIES` and friends), so the mock falls through to its `no handler` branch. The failures span `capabilities`, `keys`, `seeds`, `contexts`, `restart`, `sign` — surfaces this PR does not touch.

Fixing it means repointing those mocks at the Trust Task envelope type, which is #861/#829's follow-up and a different stream's scope. **Flagging rather than fixing here** — it isn't this PR's to carry, and folding it in would hide a main-is-red signal that someone should see.

All 10 other checks on this PR pass, and every crate this PR touches is green locally (vta-service 723 lib + 64 + 120 integration, vta-sdk, vtc-service, vta-mcp, vti-common, vtc-client, vta-mobile-core, pnm-cli, cnm-cli, vta-webvh, vta-cli-common).

stormer78:
## Merged `origin/main` (through #865) — conflict resolution

`main` and this branch both rewrote the same ACL structs for different reasons, so
every conflict was a *both-must-survive* case rather than a pick-a-side. Nothing was
dropped from either side.

**Conflicting files:** `Cargo.lock`, `vta-sdk/Cargo.toml`, `vta-service/Cargo.toml`,
`vta-sdk/src/protocols/acl_management/update.rs`,
`vta-service/src/{messaging/handlers.rs, operations/acl.rs, routes/acl.rs, trust_tasks/acl.rs}`.

### What was kept from each side

| Concern | Resolution |
| --- | --- |
| `UpdateAclBody` field shape | **This branch (#856)**: canonical wire — `did`→`subject`, `allowed_contexts`→`scopes`, nested `stepUp`/`approve`, plus `expiresAt`/`reason`, with the historical flat accessors (`step_up_approver()`, `step_up_require()`, `approve_scope()`) retained. |
| `allowedKeys` on the body | **main (#818/#865)**: the member, its doc comment, the explicit `rename = "allowedKeys"` and the `double_option` helper, layered on top of the canonical shape. Absent / explicit-`null` / array stay three distinct intentions. |
| `UpdateAclParams` (service) | Union — `expires_at` + `reason` (ours) **and** `allowed_keys` (main's). |
| Call sites (`routes/acl.rs`, `messaging/handlers.rs`, `trust_tasks/acl.rs`) | Union — ours pass `expires_at`/`reason` via the canonical accessors, main's `allowed_keys` conversion added alongside (`.clone()` where the body is borrowed). |
| `AclEntry.allowed_keys`, gate-4 enforcement in `operations::keys::sign_payload`, `routes/keys.rs`, `trust_tasks/keys.rs`, `vti-common/src/acl`, CLI plumbing (`vta-cli-common`, `pnm-cli`, `cnm-cli`) | **main, untouched** — these merged cleanly and are carried through as-is. |
| `update.rs` tests | main's three `allowedKeys` tests kept, retargeted at the canonical field names and the `subject` wire member. |
| Doc comment | main's note that the body is "snake_case pending the #856 rename" is now false — #856 is this PR — so it now explains that the rename is spelled out explicitly rather than left to the struct's `rename_all`. |

### Versions / CHANGELOG

`main` shipped vta-sdk 0.20.22 / vta-service 0.13.13 with #865, so this branch's entry
moves to the next free slots: **vta-sdk 0.20.24 / vta-service 0.13.15**. Both CHANGELOG
entries survive; #865's keeps its as-merged version numbers and was reordered above
#864's so the section list stays descending. `Cargo.lock` took `origin/main`'s and was
refreshed with `cargo update -p vta-sdk -p vta-service --workspace`.

### One thing a reviewer should look at

`allowedKeys` is **not in the published schemas yet.** `trust-tasks-rs` 0.2.43 —
what this workspace pins — has no `allowedKeys` in `acl/update/0.1` or in
`acl/_shared/0.1/acl-entry`, and both generated types are `deny_unknown_fields`.
The registry-side spec change for #818 is still in flight.

That was invisible on `main` (the body was non-canonical, so the dispatch spine
rejected every `acl/update` document anyway — the defect #856 fixes). It becomes
visible here, because after this PR the spine's schema validation actually works:
a trust-task `acl/update` document that *sets* `allowedKeys` is rejected as
`malformed_request` until the spec bump lands. REST and the DIDComm handler are
unaffected.

Resolution taken: the type keeps the member, but the **conformance witnesses leave it
absent** (`allowed_keys: None` in `conformance::acl_entry()` and the `acl/update`
sample, and in `trust_tasks::acl::tests::update_body()`), each with a comment saying
why. Populating them would assert a wire form the published spec rejects. Because the
sweep's census is derived, the member re-enters scope automatically at the
`trust-tasks-rs` bump that publishes it — same mechanism that handles the
`did-templates` 2.0 URIs. A paragraph in the CHANGELOG entry records this.

### Verification (all green, post-merge)

```
cargo check -p vta-sdk -p vta-service                       ok
cargo test  -p vta-service --lib          726 passed; 0 failed   (was 723 + 3 from #865)
cargo test  -p vta-sdk     --lib          224 passed; 0 failed
cargo test  -p vta-service --lib conformance
    every_published_dispatched_uri_has_a_witness ......... ok
    every_witnessed_task_round_trips_through_its_generated_types ... ok
cargo test  -p vtc-service --test trust_task_manifest       7 passed; 0 failed
cargo test  -p vta-cli-common --lib                        82 passed; 0 failed
cargo check -p pnm-cli -p cnm-cli                          ok
cargo clippy -p vta-sdk -p vta-service --all-targets -- -D warnings   clean
cargo fmt --all -- --check                                 clean
```

stormer78:
## Flagging a latent non-conformance that #865 introduced, with a tripwire for it

While verifying the `origin/main` merge that brought in #865 (allowedKeys), I found a gap the sweep could not see, so I probed it rather than assumed:

**`allowedKeys` is emitted by the ACL family but is not in the published `acl/_shared/0.1/acl-entry` component, which is closed (`additionalProperties: false`).**

Empirically, not inferred — an entry carrying the member is rejected by the generated type:

```
unknown field `allowedKeys`, expected one of `approve`, `createdAt`, `createdBy`,
`expiresAt`, `ext`, `label`, `role`, `scopes`, `stepUp`, `subject`, `updatedAt`, `updatedBy`
```

So **any `acl/*` response for an entry that actually uses the feature is non-conformant against the published spec today.** Nothing breaks in production, because responses are never validated at runtime — which is exactly why it is invisible, and exactly the class #857 exists to surface.

### Why the sweep did not catch it

The witnesses pin `allowed_keys: None`, and the field is `skip_serializing_if = "Option::is_none"`, so the member never reaches the wire in the sample. The witnesses pass honestly; they are simply blind to a drift that only exists when the member is populated. (The pinning was the right call — a populated witness would just fail.)

### What I added

A tripwire test, `allowed_keys_is_unpublished_and_the_witnesses_are_pinned_around_it`, that asserts the member is currently rejected. **It is written so that failing is the good outcome:** the day a `trust-tasks-rs` bump publishes `allowedKeys`, this test fails, and the fix is to delete it and fold the member into the witnesses properly. The doc comment says so at the failure site, so whoever hits it is not left diagnosing.

Same disposition as the `did-templates` case above: not fixed here, because the fix is upstream spec publication and belongs to #865/#818 — but no longer silent.

Issue #854, opened 2026-07-29
Reconcile registry↔implementation drift for VTA-owned task families (~45 unspecced URIs, version straddles)


stormer78:
## Problem

The registry (`trustoverip/dtgwg-trust-tasks-tf`) and this implementation have drifted in **both directions**. Since the consumer/server side owns its spec families, closing the gap is work that lives here.

### Implemented URIs with no spec folder in the registry (~45)

- The whole `vta/*` 1.0 surface: `vta/keys/*`, `vta/contexts/*` (beyond did-templates), `vta/seeds/*`, `vta/audit/*`, `vta/backup/*`, `vta/attestation/*`, `vta/discovery/capabilities`, `vta/management/reload-services`, `vta/webvh/**` beyond `dids/update`
- `vault/{archive,unarchive,restore,purge}/0.1` and all eight `vault/credentials/*/0.1`
- `task-consent/granted/0.1` (see the confirm-retirement issue), `consent/approve-request/0.1`
- `vtc/auth/login/0.1`; `vtc/install/claim/0.1|0.2` (registry has `claim/{start,finish}/0.1` instead — both shapes referenced somewhere)

Each of these is a live wire contract with no published schema, no registry page, no generated bindings, and no discovery entry.

### Registry versions this repo has skipped or straddled

- `auth/step-up/approve-request/0.2` unimplemented (0.1 in use) while `approve-response` dual-accepts 0.1|0.2 in a single match arm
- `device/wipe/0.2` is the only device 0.2 missing from the `wire_v0_2` table
- `policy/evaluate` (all three versions) and `policy/delete/0.1` unreferenced
- Constant-name drift acknowledged in `vta-sdk/src/trust_tasks.rs` ("`_1_0` suffix … rename pass" pending), with every const mirrored by hand in `docs/05-design-notes/trust-task-uri-registry.md`

### Root-cause item

`wire_v0_2.rs` exists solely because ten 0.1→0.2 registry bumps changed enum casing only. A registry-side versioning-policy fix is proposed as trustoverip/dtgwg-trust-tasks-tf#151 — once adopted, the adapter can stop growing.

## Suggested approach

1. Triage the ~45 unspecced URIs: spec the keepers (PR to the registry), delete/alias the accidental ones.
2. Decide `vtc/install/claim` shape (monolithic vs start/finish) and make code + registry agree.
[...]

Issue #851, opened 2026-07-29
Consolidate vta/did-templates/* and vta/contexts/did-templates/* into one six-task family


stormer78:
## Proposal

Collapse the twelve-task pair of spec families `vta/did-templates/{create,delete,get,list,render,update}/1.0` and `vta/contexts/did-templates/{same six}/1.0` into **one six-task family** with an optional `contextId` (absent = global scope, super-admin gated; present = that context, context-admin gated). VTA owns the `vta/*` spec namespace, so the spec change should be PR'd from here to `trustoverip/dtgwg-trust-tasks-tf`, with the implementation updated in `vta-service`.

## Evidence that these are one surface

- **Schemas:** each context-variant payload is byte-for-byte the global schema plus a required `contextId`, retitled descriptions, and a deeper `$ref` path. Only substantive deltas: context `render` injects ambient `CONTEXT_ID`/`CONTEXT_DID` vars, and the authorization gate differs.
- **Implementation:** both families dispatch into a single module (`vta-service/src/trust_tasks/did_templates.rs`) with paired handlers (`handle_list` / `handle_context_list`, …). Twelve URIs, one implementation.
- **History:** both families landed in the same trust-tasks PR (#89); neither supersedes the other — it's a deliberate scope split that a single optional field expresses just as well, in the style of the `provision/integration` `oneOf` precedent.
- **Smell:** `proofRequirement: REQUIRED` is blanket-stamped across the family, including pure reads (`get`, `list`, `render`), unlike the reasoned per-task levels elsewhere in the registry — consistent with mechanical duplication.

## Also in scope

- Move the ~100-line `DidTemplateRecord`/`DidTemplate` definition — currently duplicated inline in all twelve payload schemas — into `vta/_shared/` (registry-conventions issue: trustoverip/dtgwg-trust-tasks-tf#152).
- Re-derive per-task proof levels while touching the family (reads plausibly `RECOMMENDED`).

Net effect: −6 task specs, one dispatch table section, no capability loss (the per-scope ACL moves from slug-structure to the `contextId` field + party gating).

_From a cross-registry consolidation analysis, 2026-07-29._

PR #867, opened 2026-07-29
chore(deps): trust-tasks-rs 0.2.51, messaging SDK 0.18.65 + mediator 0.17.13, acl_set → account_update

stormer78:
Refs OpenVTC/verifiable-trust-infrastructure#854.

Bumps VTI's pinned `trust-tasks-rs` to the consolidation programme's published
output and sweeps every consumer break that follows. What the issue scoped as
one bump plus two list edits turned out to be a **four-link chain**, two links of
which only surfaced under CI.

## The chain

**1. `trust-tasks-rs` 0.2.43 → 0.2.51.** The issue targeted 0.2.50. That is
**not sufficient**, and the failure is silent at compile time — the whole
workspace builds clean at 0.2.50. It fails in #866's new conformance sweep:

```
https://trusttasks.org/spec/acl/list/0.1: response is not canonical:
  unknown field `allowedKeys`, expected one of `approve`, `createdAt`, ...
```

#865 added `allowedKeys` to VTI's ACL entry; the registry published it on
`acl/_shared/0.1/acl-entry` + `acl/update/0.1` one release later, at 0.2.51.
Both those PRs landed while this branch was in flight. 0.2.50 → 0.2.51 changes
nothing else in the workspace's view (identical URI index and module set).

**2. `affinidi-messaging-sdk` 0.18.62 → 0.18.65 — forced, not chosen.** 0.18.62
does not compile against the bumped registry: it constructs
`messaging::account::list::v0_1::Payload` and
`messaging::access_list::list::v0_1::Payload` with struct-literal syntax, and both
gained optional filter members.

> **Schema-additive is not source-additive.** A new *optional* field is additive to
> the JSON Schema and fully wire-compatible — old producers keep validating, old
> consumers keep parsing. But the generated Rust `struct` carries no
> `#[non_exhaustive]` and no builder, so every downstream crate that constructs the
> payload with a **struct literal** stops compiling: `missing field ...`. The break
> is invisible in the spec diff, invisible on the wire, and surfaces only as a
> downstream build failure one publish-cycle later, in a different repo.

The cheap immunisation is at the *construction* site: end every generated-payload
literal with `..Default::default()` (as `build_allow_all_acl` here does). Whether
the codegen should emit `#[non_exhaustive]` on payload structs — making the
property enforced rather than conventional, at the cost of banning literals — is
worth a decision by the codegen owners. Flagging, not proposing.

**3. `affinidi-messaging-mediator` 0.17.10 → 0.17.13 / `-test-mediator` 0.2.40 →
0.2.43 — the link that only CI could find.** It is a *dev*-dependency, so nothing
in a local `cargo check` touches it. The messaging cutover is clean on **both**
ends: 0.17.10 serves `messaging/acl/set` and answers `account/update/0.1` with

```
WARN  unsupported Trust Task type: https://trusttasks.org/spec/messaging/account/update/0.1
ERROR response failed classification=Status code: 501 Not Implemented
```

So the moment the SDK moves the client side, the embedded test mediator can no
longer provision an ACL at all — the client never gets `receive_forwarded`, and
every forwarded round-trip is refused. The server half ships in 0.17.13.

Stated plainly for deployments: **a mediator older than 0.17.13 cannot serve a VTA
built on messaging SDK 0.18.65.** There is no dual-accept window on either side.

**4. `vta-sdk/src/acl_setup.rs` migrated off the removed `acl_set`.**
affinidi/affinidi-tdk-rs#668 removed `TrustTasksOps::acl_set` outright. Replacement:

```rust
account_update(profile, did_hash, account_type, acl, queue_limits)
```

on `messaging/account/update/0.1`, which carries `acl` as one optional member
alongside `accountType` / `queueLimits`; `None` for the other two makes it the same
ACL-only partial update. `build_allow_all_acl` now returns
`account::update::v0_1::MediatorAcl` (identical field set). Fire-and-forget
semantics and the deliberate `debug!`-not-`warn!` on timeout are unchanged.

Swept the workspace for the other retired URIs (`account/{change-type,
change-queue-limits}`, `acl/set`, `admin/{add,strip,list,audit-log,config}`,
`access-list/{add,remove,clear,get}`): **no other call site**.

## Publish-lag debt flushed

The shrink-only staleness assertions turn a dependency bump into a forcing
function — these fail CI until removed, which is the harness working as designed:

- **`vta-service/src/trust_tasks/mod.rs`** — the six `vta/did-templates/*/2.0`
  entries leave `UNSPECCED_DISPATCHED_URIS` (#851/#864 authored them upstream;
  0.2.49 indexed them). List back to **56**.
- **`vtc-service/tests/trust_task_manifest.rs`** — `vtc/join-requests/decide/`
  leaves `UNPUBLISHED_CANONICAL_OK`, and `spec/vta/` goes **50 → 44** with the
  publish-lag sentence trimmed from its reason string.
- **`docs/05-design-notes/registry-drift-triage.md`** — records the rule this
  produced: *a publish-lag entry is debt with an expiry date, so every
  `trust-tasks-rs` bump is also an exception-list sweep.*

## Reconciliation the bump forces on #866

Indexing the six did-templates 2.0 URIs puts them back inside the conformance
sweep's **derived** census, whose coverage assertion then demands a witness for
each — exactly the hand-off #866 left in a comment. All six are now `checked!`
entries built from the slice's real wire types, and all six conform on the first
run (no `KnownDrift` needed). Each witness carries the **context** form; the
**global** form (`contextId` omitted) is asserted alongside, since the family's
whole point is that scope moved into one optional member — otherwise "optional"
would be the one thing about it the sweep never checks.

`allowedKeys` also enters the three witnesses whose comments promised it would "at
the bump that publishes the member". Worth separating from the exception lists:
**nothing asserted those comments.** A comment claiming a member is unpublished
cannot fail when it stops being true, and this one was stale by one patch release.
The design note records the corollary — a bump's sweep must also grep the pinned
crate for members that comments claim are missing.

## Test evidence

```
cargo test -p vtc-service --test trust_task_manifest   7 passed;   0 failed
cargo test -p vta-service --lib                      726 passed;   0 failed
cargo test -p vta-sdk --lib                          224 passed;   0 failed
cargo check --workspace --all-targets                Finished, no errors
cargo test -p vti-e2e-tests --test client_didcomm      42 passed;   0 failed
cargo fmt --all -- --check                           clean
cargo clippy -p vta-sdk -p vta-service -p vtc-service --all-targets   clean
scripts/check-changelogs.sh / check-version-bumps.sh  both pass
```

The e2e suite is included deliberately: it is the only one that exercises a real
mediator, and it is what proved link 3. It was **21 passed / 21 failed** before —
red on unmodified `main` since #861, independently of this branch, and fixed by
#868. This branch is rebased on #868 and the suite is fully green.

## Releases

`vta-sdk 0.20.25` / `vta-service 0.13.16` / `vtc-service 0.11.44`, root CHANGELOG
entry under `## Unreleased`. Version bumps may be rebased by the orchestrator as
sibling consolidation PRs merge — these slots were consumed out from under this
branch three times already (#864, #865, #866).

## Reviewer checklist

- [ ] `account_update(profile, Some(did_hash), None, Some(acl), None)` is the right
      translation of `acl_set(profile, did_hash, acl)` — confirm `None` for
      `account_type` / `queue_limits` means *unchanged* mediator-side. Note one real
      semantic difference found while reading 0.17.13: `account_update` requires the
      account to exist (`account_get` → 404 otherwise), where `acl_set` upserted via
      `get_did_acl(..).unwrap_or_default()`. The e2e suite covers the connect path
      and is green, but this is the sharpest edge in the migration.
- [ ] `account::update::v0_1::MediatorAcl` field set matches the old
      `acl::set::v0_1::MediatorAcl` — no capability silently dropped from the
      allow-all ACL.
- [ ] The mediator floor (≥ 0.17.13) is acceptable for every deployment target, and
      wants saying somewhere operator-facing beyond the CHANGELOG.
- [ ] The six new did-templates witnesses assert the response the handler *actually*
      emits, not the one the spec suggests.
- [ ] `spec/vta/` = 44 and `UNSPECCED_DISPATCHED_URIS` = 56 are the correct
      post-bump resting state (assertions pass, but the *numbers* are the artifact).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,106 @@
 
 ## Unreleased
 
+### vta-sdk 0.20.25 / vta-service 0.13.16 / vtc-service 0.11.44 — trust-tasks-rs 0.2.51, messaging SDK 0.18.65, `acl_set` → `account_update`
+
+Picks up the consolidation programme's published registry output and sweeps the
+consumer breaks it forces. The four bumps are one chain, not four choices
+(#854):
+
+- **`trust-tasks-rs` 0.2.43 → 0.2.51** (workspace `Cargo.toml`) — every spec
+  change from this programme, including `vtc/join-requests/decide/0.1`, the
+  merged `vta/did-templates/*/2.0` family, the rationalized `messaging/*`
+  family, and (at 0.2.51 specifically) `allowedKeys` on
+  `acl/_shared/0.1/acl-entry` + `acl/update/0.1`. **0.2.50 was the intended
+  target and is not sufficient** — registry PR #164 published `allowedKeys`
+  one release later, and 0.2.50 → 0.2.51 changes nothing else in the
+  workspace's view: identical URI index, identical module set, three
+  `allowed_keys: None` additions in the crate's own fixtures.
+- **`affinidi-messaging-sdk` 0.18.62 → 0.18.65** — forced, not optional.
+  0.18.62 does **not** compile against 0.2.50: it builds
+  `messaging::account::list::v0_1::Payload` and
+  `messaging::access_list::list::v0_1::Payload` with struct-literal syntax, and
+  both payloads gained optional filter members. Additive to the JSON Schema and
+  wire-compatible, but **source-breaking for struct literals** — a generated
+  `struct` with no `#[non_exhaustive]` and no builder makes every new optional
+  field a compile error in any crate that constructs it positionally. Worth
+  knowing when judging what a "patch" bump of the generated bindings can do:
+  schema-additive is not source-additive for downstream literal constructors
+  (`..Default::default()` at the construction site is the cheap immunisation).
+- **`vta-sdk/src/acl_setup.rs` migrated off `acl_set`** — the messaging family's
+  clean cutover (affinidi/affinidi-tdk-rs#668) removed
+  `TrustTasksOps::acl_set` outright. The replacement is
+  `account_update(profile, did_hash, account_type, acl, queue_limits)` on
+  `messaging/account/update/0.1`, which carries `acl` as one optional member
+  beside `accountType`/`queueLimits`; passing `None` for the other two makes it
+  the same ACL-only partial update. `build_allow_all_acl` now returns
+  `account::update::v0_1::MediatorAcl` (identical field set). Behaviour,
+  fire-and-forget semantics and the deliberate debug-not-warn on timeout are
+  unchanged. No other retired-URI call site exists in the workspace.
+- **`affinidi-messaging-mediator` 0.17.10 → 0.17.13 and
+  `affinidi-messaging-test-mediator` 0.2.40 → 0.2.43** — the fourth link, and
+  the one that is easy to miss because it is a *dev*-dependency. The messaging
+  cutover is clean on **both** ends: 0.17.10 serves `messaging/acl/set` and
+  answers `messaging/account/update/0.1` with `501 Not Implemented`
+  (`unsupported Trust Task type`), so the moment the SDK moves the client side,
+  the embedded test mediator can no longer provision an ACL at all — the client
+  never gets `receive_forwarded` and every forwarded round-trip is refused. The
+  server half of the cutover ships in mediator 0.17.13. Worth stating plainly
+  for deployments: **a mediator older than 0.17.13 cannot serve a VTA built on
+  messaging SDK 0.18.65.** There is no dual-accept window on either side.
+
+Publish-lag debt flushed by the bump — the shrink-only exception assertions
+fail until it is, which is the harness working as designed:
+
+- `vta-service/src/trust_tasks/mod.rs`: the six `vta/did-templates/*/2.0`
+  entries leave `UNSPECCED_DISPATCHED_URIS` (#851 authored them upstream as
+  trust-tasks-tf#162; 0.2.49 indexed them), putting the list back at 56.
+- `vtc-service/tests/trust_task_manifest.rs`: the `vtc/join-requests/decide/`
+  entry leaves `UNPUBLISHED_CANONICAL_OK` (the spec it was waiting on ships in
+  0.2.47). The `spec/vta/` count of 44 and the remaining 56 unspecced URIs are
+  unaffected — the bump publishes nothing else that would shrink them, so those
+  remain genuine unspecced surface rather than lag.
+- `docs/05-design-notes/registry-drift-triage.md` gains the rule that produced
+  this sweep: a publish-lag entry is debt with an expiry date, so every
+  `trust-tasks-rs` bump is also an exception-list sweep.
+
+**Reconciliation with the conformance sweep (#866).** Indexing the six
+`vta/did-templates/*/2.0` URIs puts them back inside the sweep's *derived*
+census, and its coverage assertion then demands a witness for each — exactly
+the hand-off #866 left in a comment where the twelve 1.0 witnesses used to be.
+All six are now `checked!` entries built from the slice's real wire types
+(`vta_sdk::protocols::did_template_management`), and all six conform on the
+first run — no `KnownDrift` was needed:
+
+- `list` / `create` / `get` / `update` / `delete` / `render`, each carrying the
+  **context** form of the request (the richer of the two scopes) and the
+  response the handler actually emits — `DidTemplateRecord` for `create`,
+  `get` and `update`, `{templates}` / `{name, deleted}` / `{document}` for the
+  rest.
+- The family's whole point is that scope moved out of the slug hierarchy into
+  one **optional** `contextId` (#851), and a witness can only carry one of the
+  two scopes. So the global form of all six requests is asserted alongside the
+  table: `contextId` omitted must still parse as the published `Payload`.
+  Otherwise "optional" would be the one thing about this family the sweep
+  never checks.
+
+**`allowedKeys` (#818) enters the witnesses.** #866 left it absent in three
+places — `conformance::acl_entry()`, the `acl/update` conformance sample, and
+`trust_tasks::acl::tests::update_body()` — each with a comment saying the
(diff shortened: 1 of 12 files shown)


stormer78:
## Reconciled with #866 (and rebased onto `main` @ `d38325bc`, which includes #868)

Rebased onto current `main` and finished the predicted reconciliation between this bump and the conformance sweep. Also **the pin moved 0.2.50 → 0.2.51** — see the correction below.

### Correction: 0.2.50 does not carry `allowedKeys`; 0.2.51 does

The plan for this PR assumed `trust-tasks-rs` 0.2.50 published `allowedKeys` (registry PR #164). It does not — I probed the pinned crate source before touching anything:

```
$ grep -rl allowedKeys trust-tasks-rs-0.2.50/src/specs/    # (no output)
$ grep -rl allowedKeys trust-tasks-rs-0.2.51/src/specs/
acl/{update,show,list,change_role,revoke,swap_key,grant}/v0_1.rs
```

`allowedKeys` landed one release later. So the pin is **0.2.51**, which is otherwise a no-op for this workspace — `diff -rq` between the two releases is: the seven `acl/*` spec modules, three `allowed_keys: None` additions in the crate's own fixtures (`consume.rs`, `dispatcher.rs`, `guards/acl.rs`), and packaging metadata. **Identical `schema_index.rs`, identical module set** — no new URIs enter the conformance census from 0.2.51.

### did-templates 2.0 — six witnesses, all conforming

Indexing the six `vta/did-templates/*/2.0` specs put them back in the sweep's derived census, and `every_published_dispatched_uri_has_a_witness` failed for exactly the six URIs #866 predicted. All six now have `checked!` entries built from `vta_sdk::protocols::did_template_management`'s real wire types:

| URI | request | response |
|---|---|---|
| `list/2.0` | `ListDidTemplatesBody` | `ListDidTemplatesResultBody` |
| `create/2.0` | `CreateDidTemplateBody` | `DidTemplateRecord` |
| `get/2.0` | `GetDidTemplateBody` | `DidTemplateRecord` |
| `update/2.0` | `UpdateDidTemplateBody` | `DidTemplateRecord` |
| `delete/2.0` | `DeleteDidTemplateBody` | `DeleteDidTemplateResultBody` |
| `render/2.0` | `RenderDidTemplateBody` | `RenderDidTemplateResultBody` |

**All six conformed on the first run — no `KnownDrift` entries were added, and the drift allowlist stays empty.** The teeth assertion (injected unknown member must be rejected) passes on all twelve request/response sides.

One addition beyond a plain witness. This family's whole point (#851) is that scope moved out of the slug hierarchy into a single **optional** `contextId`, and a witness can only carry one of the two scopes. The witnesses carry the *context* form (the richer); the **global** form of all six requests is asserted alongside the table — `contextId` omitted must still parse as the published `Payload`. Without that, "optional" would be the one thing about this family the sweep never checks.

### `allowedKeys` witnesses populated — and the class of debt they represent

#866 left `allowed_keys: None` in three places with comments saying the member was unpublished. All three now carry a value and the comments are gone:

- `conformance::acl_entry()` → `Some(["tenant-key-a"])` (reaches every ACL response that embeds the entry)
- the `acl/update` conformance sample → `Some(Some(["tenant-key-a"]))`
- `trust_tasks::acl::tests::update_body()` → `Some(Some(["tenant-key-a"]))`

Populated rather than left `None` on purpose: `None` and `Some(∅)` are *opposite grants* on this member, so only a present value proves it survives the wire under the canonical `allowedKeys` spelling.

Worth flagging for review: unlike the two exception lists, **nothing asserted this**. A comment claiming a member is unpublished cannot fail when it stops being true — and this one was stale by a patch release, which is how the 0.2.50-vs-0.2.51 discrepancy surfaced at all. `docs/05-design-notes/registry-drift-triage.md` now records the rule: a bump's exception-list sweep must also grep the pinned crate for members that comments claim are missing.

**I did not find a tripwire test** asserting `allowedKeys` is rejected — I searched every `allowedKeys` / `allowed_keys` occurrence in the workspace and re-read the #866 diff. The only assertions on that member are in `vta-sdk/src/protocols/acl_management/{entry,update}.rs` (absent/null/empty round-trips, camelCase pinning), none of which is publish-gated. Nothing needed deleting, and the full 726-test `vta-service` lib run is green, so no such test fired.

### Allowlist state flushed

- `UNSPECCED_DISPATCHED_URIS` — six `did-templates/*/2.0` entries gone, list back at 56. Staleness assertion (both directions) green; nothing else was stale.
- `UNPUBLISHED_CANONICAL_OK` — `vtc/join-requests/decide/` entry removed, `spec/vta/` count 50 → 44. Exact-count assertions green.

### Verification

```
cargo fmt --all -- --check                                     clean
cargo test -p vta-service --lib                                726 passed
cargo test -p vta-sdk --lib                                    224 passed
cargo test -p vtc-service --test trust_task_manifest             7 passed
cargo check --workspace --all-targets                          clean
cargo clippy -p vta-sdk -p vta-service --all-targets -D warnings clean
```

### Reviewer checklist

- [ ] 0.2.51 rather than 0.2.50 is the right call (it is the release that carries `allowedKeys`; 0.2.50 → 0.2.51 is otherwise inert here).
- [ ] The did-templates fixtures are representative — particularly `DidTemplateRecord`'s `scope` discriminator (`{"type":"context","contextId":…}`) matching the generated adjacently-tagged `Scope`.
- [ ] The global-scope side-assertion belongs in `table()` rather than as its own `#[test]`.
- [ ] `Some(["tenant-key-a"])` is the right `allowedKeys` witness value (vs. `Some([])`, the opposite-grant edge).

### Note on the push

The orchestrator's rebase onto `d38325bc` rewrote this branch, so the plain push was rejected as non-fast-forward. Pushed with `--force-with-lease=consolidate/s13-dep-bump:2ca7f1b2` — pinned to the exact expected remote SHA, and the only commit replaced is this stream's own superseded one.

PR #875, opened 2026-07-29
feat: raise backup password minimum length from 12 to 15 characters

yaroslava-kurash:
### What changed

The minimum password length enforced at **export time** is raised from 12 to 15 characters. Import/decrypt paths carry no length check — existing backups encrypted with shorter passwords remain fully decryptable (backward compatible).

| File | Change |
|---|---|
| `vta-backup/src/ops/mod.rs` | `< 12` → `< 15`; new unit test |
| `vtc-service/src/backup.rs` | `MIN_PASSWORD_LEN = 12` → `= 15` |
| `vtc-service/src/routes/backup.rs` | OpenAPI doc comment updated |
| `vtc-service/tests/backup.rs` | New integration test |
| `pnm-cli/src/commands/backup.rs` | Prompt text + guard ×2 |
| `cnm-cli/src/backup.rs` | Prompt text + guard |
| `vta-sdk/src/protocols/backup_management/descriptors.rs` | Doc comment |

### Why 15

A backup envelope carries the VTA's master seed, BIP-32 key hierarchy, and ACL — the highest-value secret in a deployment. The previous 12-character minimum falls below the 15-character threshold recommended by CIS Benchmarks for privileged-account credentials and widely adopted as industry best practice for secrets protecting cryptographic key material.

### Backward compatibility

Operators who created backups with a password between 12–14 characters can still import them — the length check only fires on new exports.

### Tests

- **`vta-backup::ops::tests::export_rejects_short_password`** (new) — unit test at the op layer: 14-char password rejected with error mentioning "15 characters"; 15-char boundary accepted.
- **`vtc-service::tests::backup::export_rejects_short_password`** (new) — integration test via `TestVtc`: same boundary check.
- **`vta-service::tests::api_integration::backup_export_rejects_short_password`** (pre-existing) — HTTP-layer test using `"short"` (5 chars); still passes unmodified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,26 @@
 
 ## Unreleased
 
+### vta-backup 0.1.3 / vtc-service 0.11.45 / pnm-cli 0.11.16 / cnm-cli 0.11.13 / vta-sdk 0.20.27 — raise backup password minimum to 15 characters
+
+The export-side minimum password length is raised from 12 to 15 characters
+across all enforcement points. Import/decrypt paths carry no length check —
+existing backups remain decryptable with shorter passwords (backward compatible).
+
+- **`vta-backup`**: `export_backup` rejects passwords shorter than 15 characters
+  with `AppError::Validation`; new `export_rejects_short_password` unit test
+  (14-char rejected, 15-char boundary accepted).
+- **`vtc-service`**: `MIN_PASSWORD_LEN` 12 → 15; new `export_rejects_short_password`
+  integration test in `tests/backup.rs`.
+- **`pnm-cli`**: prompt text and client-side guard updated on both export paths
+  (direct and trust-task descriptor).
+- **`cnm-cli`**: prompt text and client-side guard updated.
+- **`vta-sdk`**: doc comment on `InitiateExportBody.password` updated.
+
+The 15-character threshold aligns with CIS Benchmark recommendations for
+privileged-account credentials, reflecting the sensitivity of a backup envelope
+(master seed, BIP-32 key hierarchy, ACL).
+
 ### vta-service 0.13.18 — step-up approve-request minted as 0.2; inbound stays bilingual
 
 The deferred follow-up to 0.13.17 (#870): the minted step-up approve-request
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,19 +8,36 @@ The export-side minimum password length is raised from 12 to 15 characters
 across all enforcement points. Import/decrypt paths carry no length check —
 existing backups remain decryptable with shorter passwords (backward compatible).
 
-- **`vta-backup`**: `export_backup` rejects passwords shorter than 15 characters
-  with `AppError::Validation`; new `export_rejects_short_password` unit test
-  (14-char rejected, 15-char boundary accepted).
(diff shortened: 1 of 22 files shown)


stormer78:
Reviewed, and the change itself is correct. I verified the claims rather than taking them from the description:

- Only **five** `password.len()` sites exist in the workspace and all five are export-side, so the backward-compatibility claim holds — 12–14-char backups stay importable.
- The guard sits after `require_super_admin()` and before any keyspace I/O, so the new test's doc comment is accurate.
- The indirect paths are already covered without extra sites: the descriptor flow goes through `vta-backup/src/ops/descriptors.rs:111` → `super::export_backup`, and DIDComm through `messaging/handlers.rs:1170` → the same op.

I've pushed a follow-up commit (`3b76e1bf`) rather than leaving these as review comments, since a couple of them are mechanical. Happy to drop or amend any of it.

### Single source of truth

The 12 → 15 change needed five edits and still left seven docs saying 12 — which is the argument for the number not living in five places. Added to `vta-sdk`:

```rust
pub const MIN_BACKUP_PASSWORD_LEN: usize = 15;
pub fn validate_backup_password(password: &str) -> Result<(), String>;
```

Every export-side guard now calls it: the VTA op layer, the VTC op layer (its crate-local `MIN_PASSWORD_LEN` is gone), and the three CLI prompts, whose prompt text renders from the constant too. All five crates already depended on `vta-sdk`, so no new dependency edges. The `String` error was chosen so it converts into both `AppError::Validation` and `Box<dyn Error>` at the respective call sites without a wrapper.

### One judgement call worth your sign-off

The shared helper counts characters (`chars().count()`) rather than bytes. `str::len()` let a six-character non-ASCII password clear a check whose message read "at least 15 characters", since every non-ASCII scalar is 2–4 UTF-8 bytes. There's a unit test for it. This is a deliberate tightening slightly beyond the PR's original scope — it only rejects passwords that previously slipped through — so say the word if you'd rather it went in separately.

### Docs

Corrected the seven places still documenting a 12-character minimum: `CLAUDE.md`, `pnm-cli/README.md`, `docs/03-vtc/backup-restore.md` (×2), `docs/05-design-notes/vtc-backup-restore.md`, `docs/05-design-notes/backup-descriptor-pattern.md`, and the request example in `docs/02-vta/didcomm-protocol.md`.

There was also a design conflict worth naming: `docs/05-design-notes/vtc-mvp.md` §14.1 specifies `zxcvbn` (score ≥ 3) on the grounds that a bare length floor admits `Passw0rdPassw0rd` — so raising 12 → 15 is still the control that note rejects. Rather than quietly delete the position, I updated it to record that the entropy check never shipped, that a length minimum is what's actually enforced, and that 15 is an **interim** floor. Consolidating behind `validate_backup_password` is what makes that honest: swapping the length test for an entropy estimate is now a single-function change.

### Also

`vtc-service/tests/backup.rs`'s `PW` was `"twelve-char-pw!!"` — 16 chars, so it passed, but every export in that file now runs through the real guard, which makes its length load-bearing rather than incidental. Renamed with a comment saying why.

No version bumps needed; the branch already claims all five, still +1 from `main`.

Verified locally before pushing: `cargo clippy --workspace -- -D warnings` clean, `vta-backup --lib` 60 passed (including your `export_rejects_short_password` now running through the shared validator), `vtc-service --test backup` 7 passed, plus 2 new `vta-sdk` tests for the boundary and the character-vs-byte case.

PR #876, opened 2026-07-29
fix(vta-service): self-readiness gate + persistent mediator reconnect

marat-affinidi:
## Problem

On a cold start the VTA publishes its `did:webvh` and immediately opens the mediator DIDComm session. The mediator authenticates the VTA by resolving that DID itself; if it resolves before the VTA's DID document is externally reachable, the lookup fails and is negative-cached, producing a burst of authcrypt 403s. The previous one-shot connect then gave up permanently instead of recovering once the negative cache expired.

## Changes

- **Self-readiness gate** — waits until the VTA's own DID fully resolves over the network before initiating the mediator handshake, using a throwaway resolver so it exercises the same path the mediator takes. It honours the configured resolver (network vs local mode), so it also works where egress is restricted to a resolver sidecar. Endpoint-less methods (`did:key`) short-circuit.
- **Persistent reconnect supervisor** — replaces the one-shot connect: capped exponential backoff with full jitter, re-checking self-resolution before every attempt so a cold VTA never storms the mediator with unresolvable-sender auth. Configurable via `MediatorReadinessConfig` (`reconnect`, `reconnect_backoff_cap_secs`, `reconnect_max_elapsed_secs`; default retries indefinitely at a capped, jittered interval). The VTA recovers once the negative cache expires, without a restart.

## Testing

- `cargo test -p vta-service` (readiness unit tests) pass.
- Reproduced the cold-start race on a local DNS harness: without the fix the VTA 403-storms and never reconnects; with the fix it self-heals once the negative cache expires (no restart).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,26 +2,53 @@
 
 ## Unreleased
 
-### vta-service 0.13.19 / vta-config 0.2.1 / vta-backup 0.1.4 — mediator cold-start self-readiness gate + persistent reconnect (#876)
+### vta-service 0.13.19 / vta-config 0.3.0 / vta-backup 0.1.4 / vta-support 0.2.2 / vta-keys 0.1.4 / vta-policy 0.1.2 / vta-tee 0.1.4 — mediator readiness gate + connection supervisor (#876)
 
 Fixes a cold-start race where a freshly deployed VTA could fail to establish its
-mediator connection and stay disconnected. On startup, before the first connect
-attempt, the mediator DID and the VTA's own DID may not yet be resolvable
-(propagation / caching), so the initial DIDComm handshake is rejected and the
-service never retries.
+mediator connection and stay disconnected until an operator restarted it. The
+mediator authenticates a VTA by resolving that VTA's DID itself; on a cold start
+the DID document often isn't published yet, so the handshake is rejected — and
+the mediator's resolver negative-caches the failure, so the VTA's short retry
+burst fails too. It then gave up permanently.
 
-- **`vta-service`**: new `messaging::readiness` gate. Before connecting, the
-  service confirms its own DID resolves over the network; until it does, the
-  connect is deferred rather than attempted-and-failed. A `MessagingConnect`
-  supervisor then owns the mediator connection for the process lifetime,
-  reconnecting with capped exponential backoff and full jitter instead of
-  giving up after the first failure. Self-resolution honours the configured
-  remote resolver when one is set, matching how the rest of the service
-  resolves DIDs.
-- **`vta-config`**: `MediatorReadinessConfig` gains reconnect controls —
-  `reconnect` (enable/disable), `reconnect_backoff_cap_secs`, and
-  `reconnect_max_elapsed_secs`.
-- **`vta-backup`**: test-support constructor updated for the new config field.
+- **`vta-service`**: new `messaging::readiness` gate. Before connecting, the VTA
+  waits until its own DID **fully resolves over the network**, through the
+  configured resolver — the same operation the mediator performs, so a pass means
+  the mediator can authenticate us. Only `did:webvh` / `did:web` are gated;
+  `did:key` resolves without a network fetch and skips the wait.
+- **`vta-service`**: a `MessagingConnect` supervisor owns the mediator lifecycle
+  — gate, connect, and reconnect — with capped exponential backoff and full
+  jitter, re-confirming self-resolution before each attempt. It recovers both a
+  failed initial connect (once the mediator's negative cache expires) and an
+  established session whose inbound loop ended, where the VTA previously went
+  silently deaf for the rest of the process. Sessions shorter than 60s don't
+  reset the backoff, so a flapping mediator escalates rather than being retried
+  at full rate.
+- **`vta-service`**: the whole gate + reconnect runs in a background task.
+  Nothing blocks the startup path, so `/health` is live throughout and a
+  `SIGTERM` mid-gate is honoured immediately.
+- **`vta-service`**: `build_messaging` now tears the ATM down
+  (`graceful_shutdown`) on every error path after the profile is registered.
+  There is no `Drop` impl on `ATM`, so an abandoned socket keeps auto-reconnecting
+  while holding the mediator's one-socket-per-DID slot — without this a retrying
+  connect loop would leak one duelling socket per attempt. The 30s websocket
+  timeout is the sharp edge: it drops the future mid-flight, so the SDK's own
+  cleanup never runs.
+- **`vta-service`**: `DIDCommBridge` wiring is republishable rather than
+  set-once, so a reconnect actually re-points outbound sends at the new session
+  instead of silently leaving them on the dead one.
+- **`vta-config`** (breaking): `AppConfig` gains a `mediator_readiness` field,
+  and `MediatorReadinessConfig` / `ReadinessTimeoutPolicy` are new — `enabled`,
+  `retry_secs`, `backoff_cap_secs`, `max_wait_secs`, `on_timeout`, `reconnect`,
+  `reconnect_backoff_cap_secs`, `reconnect_max_elapsed_secs`. All default; the
+  defaults suit a normal LB/Kubernetes deployment. Minor bump because
+  `AppConfig` is not `#[non_exhaustive]`, so the new field breaks struct-literal
+  construction.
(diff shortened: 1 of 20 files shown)


stormer78:
## Review follow-ups pushed (93a639c)

Reviewed this and pushed fixes rather than leaving a list — the design is right, but the failure handling inside the new retry loop needed work, and a retry loop that leaks on failure is worse than the one-shot connect it replaces. Everything below is in one commit on this branch.

### The CI blocker, and why the HTTP probe is gone rather than feature-gated

`Feature combos` was failing with 8× `E0433 cannot find module or crate reqwest` — `reqwest` is behind `webvh` in `vta-service/Cargo.toml`, but `messaging::readiness` is behind `didcomm`, so `--features didcomm` didn't build. (The `didwebvh-rs` call was correctly gated; `reqwest` itself was the one missed.)

The gate's HTTP pre-probe of `did.jsonl` is **removed** instead, because it couldn't do its job:

- **It tests the wrong network path.** A direct fetch tests *the VTA's own egress to the DID host*. When egress is restricted to a resolver sidecar — the deployment the PR description explicitly calls out — the VTA cannot reach the DID host at all, so the probe would never return 200 and the gate would have timed out on **every** boot. Latent, and it would have looked like the gate "not working" rather than a bug.
- **A 200 never implied resolvability anyway**, which the code comment already acknowledged.

Resolving through the configured resolver is now the whole check, and it's what the mediator actually does. That deletes the `reqwest` dependency from the module, so the feature gate problem goes away by construction. The `did.jsonl` URL is still derived — for the operator-facing log and error message only, never fetched.

Method gating went from `did:webvh`-only to `did:webvh` + `did:web`; unrecognised methods stay ungated, since DIDComm shouldn't be withheld on the strength of a probe we can't reason about.

### Two unbounded resource leaks, both created by the retry

1. **Resolver network task, per probe.** `resolve_self_did` built a throwaway `DIDCacheClient` per call. In network mode `DIDCacheClient::new` spawns a *supervised, auto-reconnecting* websocket task to the resolver sidecar, and `affinidi-did-resolver-cache-sdk` has no `Drop` impl — teardown is the explicit `stop()`. Nothing called it. Probing before every attempt with `reconnect_max_elapsed_secs = 0` (the default: forever) meant one orphaned task + one reconnecting socket per attempt, for the life of the process. Now `stop()`ped on every path.

2. **Mediator websocket, per failed connect.** `build_messaging` had two error paths past `profile_add` that returned `Err` with a live socket and dropped the ATM. `ATM` has no `Drop` either. The SDK's own `graceful_shutdown` doc spells out the consequence: an abandoned socket *"auto-reconnects on its own timer and holds the mediator's one-socket-per-DID slot… that is how a service that opens a session per refresh cycle accumulates orphaned sockets which then duel over the same DID."* The sharp edge is the 30s `tokio::time::timeout` around `profile_enable_websocket` — it drops the future mid-flight, so the SDK's own `cleanup_failed_websocket()` never runs (its source comments that as a hard requirement).

   Pre-PR this could happen once per boot. The retry loop made it unbounded. Both paths now funnel through one `connect_transport` chokepoint that `graceful_shutdown`s the ATM before returning `Err`.

### The gate blocked shutdown for up to 5 minutes

`run_gate` was awaited inline in `server::run`, took no `CancellationToken`, and sat ~170 lines before `run` reaches its shutdown/restart `select!`. On `SIGTERM` mid-gate the axum thread exits but the process lingers for the rest of `max_wait_secs` (default 300s) with **no listener** — which reads as a hung pod.

The gate moved into the supervisor task and takes the cancellation token. The PR already stated the principle ("Backgrounding the retry keeps the reconnect horizon off the startup path so shutdown/restart stays responsive") — it just wasn't applied to the gate.

This also removed a duplicated probe: the supervisor re-checks self-resolution before every attempt, so the inline gate was doing the identical HTTP+resolve work twice on the happy path.

`on_timeout = "fail"` can no longer fail by returning `Err` from `run`, so it sets a flag + signals shutdown, and `run` converts that back into an `Err` after the threads join — the exit status still says "failed" for a systemd `Restart=on-failure`.

### It's now actually a connection supervisor

`MessagingConnect::run` returned on first success, so despite the name it only supervised the *initial* connect. If `run_inbound_loop` later exited (`"VTA inbound stream ended"`), nothing reconnected — the VTA went silently deaf while still looking healthy.

It now awaits the inbound loop instead of detaching it, and on return tears the session down and reconnects. Two things that fell out:

- **`DIDCommBridge` had to stop being set-once.** `inner` was a `OnceCell` and `set_messaging` did `let _ = self.inner.set(...)`, so a republish was **silently discarded** — a reconnect would have left every outbound send pinned to the previous session's dead socket. It's now a `RwLock<Option<Arc<BridgeInner>>>` with a `clear_messaging()`; accessors clone the `Arc` out and drop the guard before any `.await` (R1.3), which is why it's a `std` lock.
- **Sessions under 60s don't reset the backoff.** A connect that succeeds and instantly drops is treated as a failed attempt, so a flapping mediator escalates instead of being retried at base rate — reset-backoff-on-connect is exactly how this workspace got a churn storm before.

The `reconnect_max_elapsed_secs` horizon is also now measured from the start of the current failure run and reset by any healthy session, so a VTA that ran for a week and then dropped gets the full budget rather than one it exhausted six days ago.

### Destructive recovery was re-running per attempt

`drain_inbox_on_start` / `--flush-queues` sat inside the retry body. Both *delete* queued messages at the mediator, so on an unbounded retry they'd keep destroying traffic that arrived between attempts. Hoisted to one-shot, before the first connect.

### Testability

The give-up / how-long-to-wait arithmetic is extracted into `ReconnectPolicy` (`from_config`, `next_backoff`, `ceiling_for`, `session_was_healthy`). The branches that carried the risk — horizon expiry, `reconnect = false`, cap-below-base, the healthy-session floor — are now unit-tested without needing a mediator or an `AppState`; previously `MessagingConnect` had zero coverage. Also added a test that a cancelled token cuts the gate's wait short instead of burning the horizon. 20 tests in the module, up from 9.

As a side effect the supervisor asks a policy object for its backoff rather than reaching into `messaging::readiness` for loose arithmetic helpers, so those are private again.

### Versioning and docs

- **`vta-config` 0.2.1 → 0.3.0.** `AppConfig` isn't `#[non_exhaustive]`, so the new `mediator_readiness` field breaks struct-literal construction — the three `test_support` edits in this PR are the proof. A patch bump would ship that break to anyone on `vta-config = "0.2"`. #835 set the precedent (added an `AppConfig` field, bumped 0.1.1 → 0.2.0). Dependent pins moved to `"0.3"`, and the four crates whose manifests changed but weren't already bumped got patch bumps so `cargo publish` doesn't break: `vta-support` 0.2.2, `vta-keys` 0.1.4, `vta-policy` 0.1.2, `vta-tee` 0.1.4.
- **New `docs/02-vta/mediator-connection.md`** — operator guide: what the gate is for, what the supervisor does, every `[mediator_readiness]` setting with defaults, the three `on_timeout` policies, and what to check when the gate times out. There was no runtime-config reference for this at all, and `setup/from_toml.rs` writes `Default::default()`, so the section never appears in a generated `config.toml` — it was undiscoverable.
- **CLAUDE.md** gets an integration-flow entry, per its own "when adding a new flow, update both this section and the relevant docs" rule, recording the three invariants that are easy to undo: nothing on the startup path, teardown on every post-`profile_add` error, and the healthy-session floor.
- Stale doc comments corrected: `MediatorReadinessConfig`'s docs described polling an endpoint for a 200 and "hammering its own LB target", the CHANGELOG said the struct "gains" fields it introduces, and the module carried a stray `(AC #2)`.

### Verification

`cargo fmt --check`, `cargo clippy -p vta-service --all-targets`, `cargo clippy -p vta-service --no-default-features --features rest -- -D warnings`, all three CI feature combos (`rest,didcomm` / `rest` / `tee,rest,didcomm`), `cargo check --workspace --all-targets`, and `cargo test -p vta-service -p vta-config -p vta-backup` — all clean locally. All three release guards (version-bump, changelog, lockfile self-pins) pass.

### Not changed

The live-mediator handshake path (`live_prover`, `transient_handshake`) reads the bridge through the same accessors, so it picks up a reconnect for free — but nothing there was re-tested against a *mid-handshake* reconnect. Worth a look if you exercise `services didcomm update` on a flapping mediator; it's pre-existing behaviour either way, not something this PR introduces.

---

CI is green on all 12 checks, including `Feature combos` (13m15s), and the PR reports `MERGEABLE` / `CLEAN` against `main`.

PR #900, opened 2026-08-07
chore(deps): move DIDComm messaging onto the enforced-authcrypt SDK (…

oleksiipiliugin:
…0.19)

Bump affinidi-tdk 0.8->0.8.5 (pulls affinidi-messaging-sdk 0.19.0 + affinidi-messaging-didcomm 0.15.8) and the direct affinidi-messaging-sdk pins in vta-sdk/vta-service/e2e from 0.18 to 0.19, so the tsp feature and the e2e mediator fixture resolve to the same 0.19 line the default build gets transitively. This is upstream affinidi-tdk-rs #671: unpack now rejects plaintext/anoncrypt/forged-from envelopes by default. sdk 0.19 makes UnpackMetadata #[non_exhaustive]; the meta() test helper in vti-common/src/auth/didcomm.rs switches from a struct literal to Default + field assignment. No production behaviour change; the existing bind_authcrypt_sender guard and DIDComm inbound_gate remain as defence in depth.
Bumps vti-common 0.11.35 / vta-sdk 0.21.7 / vta-service 0.14.11.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,9 +90,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b7b538d0d00a984f3eff5c994553df5a0c8bb93b30305b964fd380d6d1e64554"
 dependencies = [
  "bls12_381_plus",
- "elliptic-curve",
- "ff",
- "group",
+ "elliptic-curve 0.13.8",
+ "ff 0.13.1",
+ "group 0.13.0",
  "hex",
  "rand 0.9.5",
  "sha2 0.10.9",
@@ -124,9 +124,9 @@ dependencies = [
  "ed25519-dalek",
  "getrandom 0.2.17",
  "hmac 0.12.1",
- "k256",
+ "k256 0.13.4",
  "multibase",
- "p256",
+ "p256 0.13.2",
  "p384",
  "p521",
  "rand 0.10.2",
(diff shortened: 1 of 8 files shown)


stormer78:
Closing in favour of #908, which carries this forward.

Thanks for spotting the enforced-authcrypt pin — the security reasoning here was
right, and #908 keeps it verbatim. What changed is that `main` moved underneath
this branch, so most of the diff no longer has anything to apply to:

- **#902 already landed the `affinidi-messaging-sdk` `0.18` → `0.19` pins** in
  `vta-sdk`, `vta-service` and `tests/e2e`, and the `UnpackMetadata`
  `#[non_exhaustive]` test-helper fix in `vti-common`.
- The three version bumps here (`vti-common` 0.11.35, `vta-sdk` 0.21.7,
  `vta-service` 0.14.11) are all now claimed on `main` by other PRs —
  `vta-service` is at 0.14.16.
- `changelog.d/900-enforced-authcrypt-sdk-0-19.md` collides with the existing
  `changelog.d/900-envelope-type-from-the-binding-crate.md`.

Rebased, this reduced to a single meaningful manifest line plus a lockfile that
had gone stale (it pinned sdk 0.19.0; `main` is past that).

Two notes carried into #908:

- **The `affinidi-tdk = "0.8.5"` pin is kept**, with the rationale recorded at
  the pin. The lockfile happened to hold 0.8.5, so the property held by
  accident — a `cargo update` or a clean re-resolve could have walked sender
  authentication back out of the library unnoticed. Good catch.
- **`affinidi-messaging-didcomm-service` → `0.3.22` would have been inert.** That
  dependency isn't in `Cargo.lock` at all and no workspace member consumes it;
  `vta-service` and `vtc-service` both carry comments recording that they
  replaced it with local `shim.rs` / `router.rs` equivalents. #908 removes the
  dead `[workspace.dependencies]` entry instead of bumping it.

#908 also finishes the deduplication this started: `affinidi-messaging-sdk`
0.18.65 was still alive behind the `affinidi-messaging-test-mediator` →
`affinidi-messaging-mediator` dev chain (the copy #902 called out as remaining).
`test-mediator` 0.2.47 moved to sdk 0.19, so the graph now holds one messaging
SDK.

PR #911, opened 2026-08-09
chore(deps): trust-tasks 0.4, and the consent digest becomes digestMultibase

stormer78:
## Why now

`affinidi-messaging-sdk` **0.19.3** depends on `trust-tasks-rs` **0.4.0**, which is what unblocks this. Before that release, bumping our direct pin resolved to **two** copies of `trust-tasks-rs` in one graph — the TDK held 0.2.x while we moved — and broke `vta-sdk::acl_setup`:

```
expected `MediatorAcl`, found a different `MediatorAcl`
note: there are multiple different versions of crate `trust_tasks_rs` in the dependency graph
```

There is now exactly one copy, at 0.4.0.

## ⚠️ This is a wire change, not a version number

0.4 moves `payloadDigest` to the shared `DigestMultibase` type — a multibase-encoded multihash matching `^[zumbfF][A-Za-z0-9+/=_-]+$` — and the spec is explicit that "a bare hex string or a `sha-256:`-style prefix hard-codes one algorithm into the wire contract and is non-conforming."

We emitted bare hex. Because the dispatcher validates payloads against the published schema and `task-consent/decision/0.1` is published, on 0.4 the VTA would have started **rejecting its own approvers' decisions**. Migrating is not optional at this version.

- `vta_policy::consent::{payload_digest, wire_digest}` now emit base58btc over the sha2-256 multihash (`0x12 0x20` || digest), matching the `did:key` / `did:webvh` convention already used by `vta_sdk::did_key::ed25519_multibase_pubkey`.
- `vta-mobile-core` **parses** the draft digest into `DigestMultibase` rather than passing a string through — so a stale hex digest fails at the device that would otherwise sign a decision the executor could never match. That guard immediately caught two stale test fixtures and a placeholder conformance witness (`"b64u-digest"`, which satisfied the old schema by accident and tested nothing about the encoding).
- `TrustTask` gained `parentThreadId`; the two unrouted parse-failure errors pass `None` — no thread to name, for the same reason there is no issuer.

`vta-policy` takes a **minor** bump rather than a patch, because the digest is a wire contract rather than an implementation detail. The dependent pin in `vta-service` moves with it.

## What reviewers need to decide

**Every party that computes the digest independently has to move together.** The browser plugin is the one outside this repository. A mismatched pair produces an approval that is given, accepted by the human, and then silently never takes effect — the failure shape `is_ceremony_task`'s header describes. In-flight pendings and grants are invalidated by the encoding change, but they are TTL'd at 900s, so the window is short.

If the plugin can't move in lockstep, this should be held or landed as part of a coordinated release rather than merged on its own.

## Known follow-up, deliberately not fixed here

`vta-mobile-core` derives the operator's 6-character match code from the first characters of the digest. Under multibase+multihash **the first three are always `zQm`** — the sha2-256 signature, not entropy:

```
zQmcdLJ…   zQmRTnb…   zQmb7oR…   zQmbu6r…      ← four different payloads
```

So the code now carries ~17.6 bits where it carried ~35 — a ~300,000× smaller space for an offline search to find a payload that displays the same code to the operator. It still *looks* like six random characters, which is what makes it worth stating.

I've left it alone: the value has to agree with whatever the requesting screen renders, and that screen isn't in this repo, so changing it unilaterally risks breaking matching outright rather than merely weakening it. The principled fix — derive from the decoded digest bytes, skipping the 2-byte multihash prefix — should ride the same coordinated release as the digest change, since both touch the approver screen.

## Testing

Full workspace suite green; `cargo clippy --workspace -- -D warnings` clean. The registry census is unaffected — the 0.2.60 → 0.4.0 spec index is purely **additive** (`vta/credentials/issue/0.2`, `vtc/relationships/request/0.1` added; nothing removed), so no allowlist grows and no URI we send was retired.

The `vta-policy` digest test now asserts the *structure* (multibase base58btc, `0x12 0x20` multihash prefix, 34 bytes) rather than a character count — base58 is not byte-aligned, so a length assertion was the wrong shape for this encoding.


stormer78:
Added the principled match-code fix (`94cf062f`), so both changes that touch the approver screen ship together.

`match_code_from_digest` now decodes the multibase and strips the multihash prefix before taking the six characters, instead of slicing the encoded string. The reason it matters is that the encoded form's first three characters are always `zQm`:

```
zQmcdLJ…   zQmRTnb…   zQmb7oR…   zQmbu6r…    ← four different payloads
```

Slicing that would have left ~17.6 bits where the operator believes they are comparing ~35 — an offline search for a colliding code drops from ~60 billion candidates to ~195k — while still looking like six random characters. Nothing on screen would have said it had weakened.

**The migration is invisible on this screen.** Because the digest is still SHA-256, decoding first reproduces exactly what the code was under bare hex (`hex(digest)[..6]` either way). The pre-existing assertion `assert_eq!(r.match_code, "3b0c7f")` — written before any of this — passes verbatim against the multibase form. That is the property in evidence rather than in argument, and it means the requesting screen's coordination burden is limited to adopting the same derivation, not to agreeing a new code format.

A digest that is not a multibase sha2-256 multihash is now refused rather than silently yielding a code, which is how the last stale hex fixture in the module was found.

Full suite green locally; re-running CI on the new head. The digest wire change still needs the browser plugin to move in lockstep — that part is unchanged.

stormer78:
Heads-up that the **"Known follow-up, deliberately not fixed here"** section of this description is stale — it describes an approach that was superseded inside this same PR.

It says the match code now carries ~17.6 bits and that the fix was left alone because the requesting screen isn't in this repo. But the shipped `match_code_from_digest` already does the principled thing:

```rust
let (_base, bytes) = multibase::decode(payload_digest)?;
let digest = bytes.strip_prefix(&MULTIHASH_SHA2_256_32)?;
Ok(hex::encode(&digest[..need])[..MATCH_CODE_LEN].to_string())
```

So the code derives from the digest bytes, keeps its full entropy, and reproduces exactly what the screen showed when the wire carried hex. The function's own doc comment says so, and `changelog.d/911-trust-tasks-0-4.md` gets it right too (*"The operator's match code keeps its entropy"*) — it's only the PR description that still describes the naive slice as the shipped behaviour.

**Why it's worth correcting rather than leaving:** this is the artifact a downstream integrator gets linked to. We're the requesting screen the section refers to (`affinidi/affinidi-webvh-service`), and reading it as a live caveat, we initially matched what it described — leaving our side slicing the encoded string. That would have shown `zQmcdL` where your device shows `d449ac`: every code mismatching, under a hint that tells the operator "if the codes differ, deny it." Caught it before merging, but only by reading the shipped source rather than the description.

Our side is now aligned with `match_code_from_digest` — decode, strip `0x12 0x20`, hex the leading bytes — and pinned against your test vector (`zQmSK9pGKFnmc77pqyNAPJyPKt8rMqctngfg3vwuMArwGYZ` → `3b0c7f`) so a future divergence on either side fails a test instead of quietly reaching an operator: affinidi/affinidi-webvh-service#159.

One suggestion, take it or leave it: that vector is now a cross-repo contract in everything but name. If it ever moves, it'd be worth a note in the changelog fragment, since the failure mode is silent on both sides.

PR #939, opened 2026-08-12
feat(nitro): un-bake tenant config, deliver to the enclave over vsock

oleksiipiliugin:
### Summary

This PR adds an opt-in fleet mode for Nitro Enclaves: one tenant-independent
EIF and one **PCR0** can serve multiple tenants.

Instead of sending a complete `config.toml`, the parent delivers a small,
typed tenant overlay over **vsock port 5800**. The enclave validates and applies
it to a PCR0-committed fleet policy base before KMS bootstrap.

The existing fully baked mode remains the default.

### Build modes

- **`BAKE_CONFIG=true` (default)** — bakes the complete `config.toml` into the
  EIF. The config is committed to PCR0, and the binary contains no vsock config
  fetch code. Intended for self-hosted and single-tenant deployments.
- **`BAKE_CONFIG=false`** — bakes `config.fleet-base.toml` and compiles the
  `tenant-overlay` feature. Tenant-scoped values are delivered at boot, allowing
  one image and PCR0 to serve multiple tenants.

Select fleet mode with:

```bash
VTA_BAKE_CONFIG=false ./deploy/nitro/build-vta.sh
```

### Tenant overlay

The parent serves a versioned JSON envelope:

```json
{
  "version": 1,
  "overlay": {
    "vta_name": "acme",
    "public_url": "https://vta.acme.example.com",
    "tee_kms": {
      "key_arn": "arn:aws:kms:us-east-1:111122223333:key/abcd",
      "vta_did_template": "did:webvh:{SCID}:acme.example.com:vta",
      "anchor_table_name": "vta-rollback-anchor-acme"
    },
    "messaging": {
      "mediator_did": "did:webvh:...:mediator"
    }
  },
  "integrity": null
}
```

The overlay uses `#[serde(deny_unknown_fields)]` and explicit field-by-field
application. It cannot carry `admin_did`, `vta_did`, `tee.mode`, any `allow_*`
flag, or other security-sensitive configuration.

`tee.kms.key_arn` must be a valid KMS key ARN from an account listed in the
baked, PCR0-committed `tee.allowed_kms_accounts`. Its region is derived from
the ARN. An empty account allowlist denies every overlay.

### Fail-closed delivery

Overlay fetching and validation run inside `vta-tee`, not the shell entrypoint.
The implementation includes bounded retries, connect/read timeouts, a 1 MiB
size cap, strict version validation, unknown-field rejection, and fail-closed
application.

On real Nitro, any fetch, parse, validation, or application failure aborts the
boot. `deploy-enclave.sh` also refuses to start a fleet bundle without a
rendered tenant overlay.

### Config attestation

The enclave captures a canonical, secret-free view of the effective config
after overlay application and DID reconciliation. Its SHA-384 digest is placed
in Nitro attestation `user_data`.

`POST /attestation/config-report` returns a fresh, nonce-bound report containing
the signed evidence and the exact `configView` that was hashed.

The SDK verifier requires the caller's nonce, approved PCR0, optional PCR8, and
the tenant's expected KMS key ARN. A successful
`ConfigAttestationReport::verify()` proves the Nitro chain and signature, PCR
pins, nonce freshness, config-view digest, and mandatory tenant KMS-key match.
It returns the `VerifiedConfigAttestation` typestate.

### Security notes

- PCR0 authenticates the image and baked fleet policy, not tenant-specific
  runtime values.
- The typed overlay prevents the parent from injecting unapproved fields.
- The config report proves which effective configuration the enclave booted.
- Mandatory KMS-key pinning prevents onboarding with a parent-controlled key.
- Each tenant's KMS key policy must constrain the calling principal as well as
  the Nitro PCRs; a shared PCR0 alone does not isolate tenants.
- The parent can withhold the overlay or report, but cannot forge or replay a
  valid report for a fresh nonce.

### Other changes

- Added `config.fleet-base.toml` and `render-tenant-overlay.sh`.
- Added parent-proxy routing derived from the tenant overlay.
- Added regional KMS and DynamoDB endpoints to the proxy allowlist.
- Preserved the enclave configuration floor as defense in depth.
- Made stored VTA identity take precedence over later configuration.
- Added tests for overlay parsing, forbidden fields, KMS validation,
  fail-closed boot behavior, secret stripping, and attestation verification.


stormer78:
## Review

Well-motivated change — one image / one PCR0 is the right goal. But I don't think
it's mergeable as-is: there's a functional gap that would brick the documented
deploy path, and a trust-boundary change shipped with its mitigation explicitly
stubbed out.

### Blocking

**1. The production parent proxy doesn't serve the envelope.**

The vsock:5800 server was added only to `parent-proxy.sh`
(`deploy/nitro/parent-proxy.sh:177-190`). But the documented and automated
production path starts the Rust binary instead — `deploy/nitro/deploy-enclave.sh:239-257`
builds and runs `enclave-proxy`, and `README.md:795-850` presents it as *the*
parent proxy. `enclave-proxy` declares channels for 5100/5200/5300/5400/5600
(`enclave-proxy/src/main.rs:42-62`) and has nothing on 5800.

So on a standard deploy: the enclave retries for 60 s, falls through to the
env-var fallback, which emits no `[tee.kms]` section
(`enclave-entrypoint.sh:160-185`) — and TEE bootstrap fails. Un-baking the
config breaks the standard deployment until `enclave-proxy` gets the same
channel.

Related: nothing *produces* `config-envelope.json`. `build-vta.sh:268-273,338`
still generates and ships a plain `config.toml` as a build artifact, and
`deploy-enclave.sh:78,152` still requires one in the bundle. The envelope is
described as "rendered off-box" with no command given. At minimum the README
needs the one-liner:

```bash
jq -Rs '{version:1, config_toml:., integrity:null}' config.toml > config-envelope.json
```

and ideally `build-vta.sh` emits it alongside the EIF.

**2. `integrity: null` hands the untrusted parent control of the enclave's
security policy.**

The Nitro threat model treats the parent as untrusted — that's the entire point
of the enclave. Baking config into the EIF meant PCR0 committed to it. This PR
removes that commitment and replaces it with an unauthenticated read, with the
integrity field documented as "reserved for a future attestation anchor."
Concretely, the parent now chooses:

- **`admin_did`** (`config.toml:62-69`) — "grants unrestricted access to this VTA
  on first boot."
- **`tee.mode`** — and `TeeMode`'s `#[derive(Default)]` is `Optional` ("TEE used
  if available, continues without it", `vta-config/src/lib.rs:714-721`). A served
  config that merely *omits* `[tee] mode` silently drops the enclave out of
  `required`. `simulated` is also accepted.
- **`tee.kms.key_arn`**, `vta_did` / `vta_did_template`, `resolver_url`,
  `mediator_did`.

Shipping the channel now and the anchor later means the window is open in
whatever gets deployed in between. Two cheap fixes that preserve the PR's
benefit:

- Have the enclave **commit the config digest into the NSM attestation
  document's `user_data`**. PCR0 stays tenant-agnostic; external verifiers pin
  (PCR0, config-digest), so the config is still attested. (KMS key policies can
  only condition on PCRs, so this doesn't help the KMS path — but it makes the
  config externally verifiable, which is most of the value.)
- Enclave-side **floor checks** regardless of what's served: refuse a config that
  sets `tee.mode` below `required`, or that carries `admin_did`, when running
  with a real `/dev/nsm`.

**3. Multi-tenant isolation moves off PCR0 and onto the KMS key policy.**

Previously `key_arn` was baked, so PCR0=X implied "this image only ever talks to
tenant X's key." Now one PCR0 is shared across all tenants and the parent
supplies `key_arn`, so anything satisfying `kms:RecipientAttestation:PCR0` on
tenant B's key can reach it. Separation now rests entirely on the key policy's
principal/account conditions. That's a direct consequence of the headline benefit
and the README should state it as a **requirement** — each tenant's key policy
must constrain the calling role, not just PCR0 — rather than leaving it implicit.

### Should fix

**4. Fail-open on config-delivery failure.** `enclave-entrypoint.sh:146-175`:
fetch fails → silently generate a default config and boot. In production, "the
parent's config server didn't come up" should be a hard stop with a clear
message, not a 60 s pause followed by a confusing downstream KMS error. Gate the
fallback behind an explicit `VTA_ALLOW_DEFAULT_CONFIG=1`. The parent side
compounds this — a missing envelope only prints `SKIP config server`, so the
whole misconfiguration is two soft warnings deep.

**5. `version` is never checked.** The entrypoint validates only `.config_toml`
presence. A versioned envelope whose consumer ignores the version buys nothing —
add `jq -e '.version == 1'`.

**6. Unbounded read from an untrusted source into enclave RAM.**
`socat -u "VSOCK-CONNECT:…" - > "$envelope_tmp"` has no size cap, and the enclave
rootfs is memory out of its fixed allocation. Cap it (`head -c 1048576`). Same
reasoning as the existing global 1 MB request-body cap, which is justified
precisely by TEE memory pressure.

**7. No timeout on the socat read.** If the parent accepts and then hangs, the
read blocks forever and the retry loop never advances — the "give up after ~60s"
comment isn't actually guaranteed. Add `,connect-timeout=5` plus an outer
`timeout`.

### Minor

**8. `changelog.d/939-…md` should be deleted.** `changelog.d/` was removed in
\#938, which is now on `main`; this branch is based on 76ce7a1f (pre-#938).
Rebase and move that text into the PR body — it becomes the squash commit body
and thus the changelog. Per CLAUDE.md: "A PR adds no changelog file and edits no
version."

**9. `SYSTEM:"cat ${CONFIG_ENVELOPE}"`** (`parent-proxy.sh:184`) spawns
`/bin/sh -c` per connection with an env-var-supplied path interpolated into the
command string. `OPEN:"${CONFIG_ENVELOPE}",rdonly` with `-u` is simpler and has
no shell.

**10. "Backward compatible" is misleading.** An EIF has no volume mounts, so "a
mounted/baked config.toml still works" holds only for plain `docker run` local
dev or an *old* image. Once you rebuild with this Dockerfile there is no baked
config. Phrase it as "existing EIFs keep working; rebuilt EIFs require the
envelope."

**11. The README note asks the reader to reconcile the document.** "Passages
below that still say 'config is baked into the EIF' describe the legacy
behaviour" leaves ~6 stale references (`README.md:179, 671, 1039, 1228, 1261,
1513`) in a runbook where getting this wrong bricks a tenant. Fix them rather
than annotating them.

---

Two things done right worth noting: the entrypoint's
`mkdir -p "$(dirname "$CONFIG_PATH")"` correctly covers the `/etc/vta` directory
that the removed `COPY` used to create implicitly, and using the fetch function
in an `elif` condition keeps `set -e` from tripping on the retry loop.

stormer78:
## Fresh review — 5 commits, rebased onto `51094b15`

**Every point from the earlier review has been actioned**, and well: `enclave-proxy`
now serves vsock:5800 (`channels.rs:run_config_server` + `--config-envelope` +
`deploy-enclave.sh` wiring), `build-vta.sh` emits `config-envelope.json`, the
envelope `version` is checked, the read is size-capped and timeout-bounded,
`SYSTEM:"cat …"` became `OPEN:…,rdonly`, the changelog fragment is gone with no
version edits, and the README's stale "baked into the EIF" passages are fixed
rather than annotated. The security floor + attestation anchor are real work, not
a hand-wave.

Two things still block it — one a mechanical regression introduced by the fix for
#4, one a hole in the floor check itself.

### Blocking

**A. `set -e` kills the entire fail-closed branch — `enclave-entrypoint.sh:225-256`**

The earlier review specifically noted that calling the fetch from an `elif`
*condition* was what kept `set -eu` (line 29) from tripping. The refactor moved it
to a bare call:

```sh
else
    fetch_config_over_vsock      # returns 1 or 2
    fetch_rc=$?                  # never reached
fi
```

A bare function call is not a condition context, so a non-zero return exits the
shell immediately. Verified with `/bin/sh`: the statement after the call never
runs. Consequences:

- `fetch_rc=1` (unreachable parent): silent `exit 1`. None of the FATAL
  diagnostics at 243-245 print, and **`VTA_ALLOW_DEFAULT_CONFIG=true` is dead
  code** — the documented local/dev fallback cannot be reached at all.
- `fetch_rc=2`: the function's own FATAL does print, but the script exits 2 and
  line 236 never runs.

Fix: `if fetch_config_over_vsock; then fetch_rc=0; else fetch_rc=$?; fi`. There's
no shellcheck in CI (`.github/workflows/` has ci/publish/release-mobile-ios only),
so nothing catches this class.

**B. The floor check requires `mode = required` but not `[tee.kms]` — `vta-enclave/src/main.rs:161-172`**

`TeeConfig.kms` is `Option<TeeKmsConfig>` with `#[serde(default)]`
(`vta-config/src/lib.rs:456`), and `AppConfig::load` records unknown keys but
explicitly does *not* reject them ("we don't reject — `validate()` warns", line
741). So a parent-served config with `[tee] mode = "required"` and `[tee.kms]`
absent (or mis-sectioned/typo'd) passes `config_floor_violation` cleanly, and
then:

- `main.rs:216` — `tee_bootstrap = None`, no KMS bootstrap at all.
- `main.rs:250` — seed comes from `create_seed_store(&config)`, i.e. the
  **parent-supplied `[secrets]` block** (`plaintext`/`config` seed included).
- `main.rs:190` — `#[cfg(vsock-store)]` store selection is
  `if config.tee.kms.is_some()`, so it falls to a local in-enclave store.
- `vta-config/src/lib.rs:768` — `kms_locked = config.tee.kms.is_some()` ⇒ **the
  env-override lockdown switches off**.
- `init_tee` (`vta-tee/src/lib.rs:53-72`) succeeds on real Nitro regardless of
  `kms`, and `maybe_bootstrap_admin` returns `Ok(())` early at line 52 — leaving
  `BOOTSTRAP_CARVEOUT_CLOSED_KEY` unset, so `POST /bootstrap/request` is open to
  whoever calls it first.

That is a complete bypass of the property the floor is meant to protect.
`config_floor_violation` needs a third rule: on `nsm_present`, `[tee.kms]` must be
present (and it deserves a unit test alongside the four existing ones).

### Should fix

**C. First-boot `key_arn` is parent-chosen, and the anchor that would catch it
isn't obtainable.** The floor stops mode downgrade and `admin_did`, but the parent
still picks `tee.kms.key_arn`. On first boot the enclave generates the seed and
encrypts it under whatever key the parent named; the ciphertext lives in the
parent-side fjall store. If that key is one the parent controls, the seed is
recoverable off-box without attestation. The mitigation for this is precisely the
config-digest anchor — and the code comment at `main.rs:337-350` is honest that it
only *produces* evidence: log-only, delivered over the parent's own log channel,
empty nonce. So the model's security currently rests on a verification step no
verifier can perform. Either land the attested pull path in this PR, or state in
the README as an **operational requirement** that a tenant must obtain and check
`(PCR0, config-digest)` before onboarding — right now the README presents the
anchor as a property rather than as a step someone has to take.

**D. The floor keys on `/dev/nsm`, not on config provenance.** The entrypoint's
documented precedence is "a mounted/baked `config.toml` wins" (line 143-160), and
`config.toml:74-79` is explicit that `admin_did` is fine for baked/mounted
configs — but the Rust check can't tell the two apart, so a legitimately baked
(PCR0-committed) config with `admin_did` set is refused on real hardware. Narrow
in practice since the Dockerfile no longer copies one, but the two documents
contradict each other. Clean fix: have the entrypoint export
`VTA_CONFIG_SOURCE=vsock` on the fetch path and gate the floor on that —
trustworthy precisely because the entrypoint is in PCR0.

### Minor

5. **`main.rs:93-101` re-derives `config_path`** with the same three-step
   resolution `AppConfig::load` uses. `AppConfig.config_path` is a pub field set
   by `load` (`vta-config/src/lib.rs:168, 750`) — use it for the digest so there's
   no way to hash a different file than the one loaded.
6. **`enclave-entrypoint.sh:316`** has the same `set -e` shape (`vta-enclave …`
   then `VTA_EXIT=$?`). Pre-existing, but this PR adds two new hard-exit paths, so
   the "VTA exited with code N" line and the `sleep 10` console keep-alive are
   exactly what an operator loses when the new floor fires.
7. **Oversized envelope reads as "not ready".** `head -c` truncates, jq fails to
   parse, and the loop reports `config server not ready` 30 times over 60 s. A
   `[ "$(wc -c < "$envelope_tmp")" -ge "$MAX_CONFIG_ENVELOPE_BYTES" ]` check with a
   distinct message would save a confusing debug session.
8. **`config.toml:80-82`** still says the auto-generated credential is
   "retrievable via `GET /attestation/admin-credential`". That endpoint was retired
   in Phase 3 (`routes/mod.rs:476`). It sits directly under the new UN-BAKED
   WARNING that now *forces* operators onto that path — worth correcting to
   `pnm bootstrap connect` in the same edit. (I checked the forced path itself
   works: `admin_did` unset leaves the carve-out open for Mode B,
   `admin_bootstrap.rs:131-141`.)
9. `"${PROXY_ENVELOPE_ARGS[@]}"` on an empty array under `set -u` needs bash ≥ 4.4
   — fine on AL2023, noting it only because the script is `set -euo pipefail`.

Worth calling out as done right: the version-mismatch path is deliberately
non-recoverable rather than falling through to a default, the single-owner comment
on vsock:5800 pre-empts a bind conflict that would otherwise be diagnosed the hard
way, and `AsyncWriteExt::shutdown` is fully-qualified with a comment explaining the
inherent-method collision.

stormer78:
Reviewed this and pushed the fixes straight onto the branch (d1580f30) rather than sending a list back — the design is sound, and everything below is a guard that reads as mandatory but has an input for which it silently does nothing.

### The one that matters most

**An empty `expected_pcr0` disabled the "mandatory" PCR0 pin** (`vta-sdk/src/attestation/mod.rs`). The signature takes `&str` rather than `Option<&str>` precisely to make the image pin non-optional, but forwarded it to `check_pcrs` unchecked — and the new `pcr_hex` helper renders the all-zero PCRs of a Nitro **debug-mode** enclave as `""`. So a consumer whose pin came from an unset config/env value calls `report.verify(&nonce, "", None, key_arn)`, compares `"" == ""`, passes, and gets a `VerifiedConfigAttestation` for an enclave the operator has console access to — off a genuinely AWS-root-signed document. Both pins are now shape-checked (96 hex chars, SHA-384) before any comparison, as a typed `InvalidExpectedPcr`.

### The rest

- **`allow_anchor_init` was bypassable by omitting `tee_kms`** (`vta-config/src/tenant_overlay.rs`). The post-condition sat inside `if let Some(kms_overlay)`, so `{"version":1,"overlay":{}}` applied cleanly into exactly the degraded manifest-only anti-rollback state the check exists to prevent. Only the placeholder `key_arn` aborting KMS bootstrap saves this today; any base baking a real `key_arn` alongside `allow_anchor_init` boots degraded. Now guards the *resulting* config on every apply.
- **`READ_TIMEOUT` was per-`read()`, not overall** (`vta-tee/src/tenant_overlay.rs`). The doc says "overall", but a parent dribbling one byte just inside every 10s window was bounded only by the 1 MiB cap — ~120 days — and the enclave hung instead of raising the fail-closed `OverlayFetchError` its caller is written to act on. The deadline now spans the whole envelope.
- **Materializing an absent `[messaging]` failed open**: `m.mediator_did.unwrap_or_default()` wrote `mediator_did = ""` and returned `Ok`, the one such path in `apply_tenant_overlay`. Now `MessagingMediatorDidRequired`.
- **`POST /attestation/config-report` was a permanent 500 on non-enclave `tee` builds.** The route registers under `#[cfg(feature = "tee")]` but only `vta-enclave` captures the snapshot, so the local daemon in simulated mode answered `AppError::Internal` on every call, forever. Absence is a capability answer — now 503, which is what the OpenAPI doc already advertised.
- **`parse_envelope` checked `version` after the strict parse**, so a future v2 envelope surfaced as `Parse("unknown field ...")` instead of `UnsupportedVersion(2)` — losing the only diagnostic the version field exists to provide. A lenient probe reads it first.
- **Parent-side proxy restart** SIGTERM'd and slept a fixed 1s. If the old process still held `vsock:5800` the new proxy's config listener failed to bind, `run_config_server` logged and returned while the process stayed alive, `kill -0` reported success — and the only symptom was the enclave exhausting its 30 fetch attempts and aborting the boot, with nothing on the parent side to say why. Now polls for the old PID to exit, and a config-channel bind failure exits the proxy.
- `parent-proxy.sh`'s new comment documented the retired `{"version":1,"config_toml":"…"}` shape; since `ConfigEnvelope` is `deny_unknown_fields`, an operator hand-rolling from it would hard-fail the boot.

### A gap worth knowing about

`vta-tee/tenant-overlay` is not a default feature and nothing in the default workspace graph enables it, so **`cargo test --workspace` never compiled `vta_tee::tenant_overlay` at all** — every test here for envelope parsing, the version gate and unknown-field rejection was dead in CI. Added `cargo test -p vta-tee --features tenant-overlay` to the feature-matrix job, plus a `cargo check` of the fleet `vta-enclave` build alongside the existing baked one.

### One thing I deliberately did not change

The unauthenticated `configView` is the full effective config, so any anonymous caller learns `tee.kms.admin_did`, the `allow_*` break-glass posture, `anchor.table_name`, `storage_key_salt`, `resolver_url`, `store.data_dir` and `trusted_presentation_verifiers`. I considered narrowing it and decided against: those fields are what make the attestation worth anything — a verifier that can't see the `allow_*` flags can't distinguish an enclave with break-glass off from one with it on — and each is public by nature, useless without AWS credentials, or documented as not-secret (`storage_key_salt` is an HKDF salt over a per-VTA high-entropy seed). Narrowing weakens the gate rather than hardening it. I documented the exposure and the reasoning on `compute_config_attestation_view` so it reads as a reviewed decision, and pointed the "add new secrets here" rule at the existing strip-test. Happy to trim it if you disagree.

Regression tests added for the PCR pin, the anchor bypass, the messaging default and the version ordering; design note §3.2 updated to say why both post-conditions guard the result rather than the input. `cargo test --workspace` is green (4372 tests), as are `fmt`, `clippy`, and the `tee` / `tenant-overlay` / fleet-enclave permutations. No `version =` fields touched.

PR #945, opened 2026-08-12
fix(vta-service): stop create-did-webvh minting a DID whose keys the store doesn't hold

robert-affinidi:
The offline CLI derived entity keys twice: once for a preview document
nobody sees (non-interactive mode), and once in the operation. The path
counter incremented both times, so the DID document embedded keys from
indices 0,1 while the store held keys from indices 2,3 — a silent
mismatch that broke DID document verification.

Skip the preview derivation when --url is set (non-interactive). The
operation derives once and uses the same keys in both the document and
the store.


stormer78:
Pushed a follow-up commit to this branch (`a42fd7b6`) — the diagnosis and the fix were right, but review turned up a regression and a coverage gap. Nothing of yours was rewritten; it's additive.

**1. Non-interactive DIDs stopped advertising DIDComm.** The deleted block set the service unconditionally in non-interactive mode — `want_didcomm` was `if interactive { prompt } else { true }` — but `add_mediator_service: false` stayed behind, with a comment saying the document handled it. With no document, `build_did_document` only emits the service `if add_mediator_service && config.messaging.is_some()`, so `--url` was minting DIDs with no `service` entry at all. Since the DID document is authoritative for which protocols a party speaks, that's an empty protocol intersection with every peer — an unreachable DID, failing far from here. Now set from `[messaging]` in the non-interactive branch; interactive still builds its own service into the document it passes, so nothing is advertised twice.

**2. The tests weren't running in CI.** The module is gated `#[cfg(all(test, feature = "config-seed"))]`, and nothing in the workspace enables that feature — so plain `cargo test --workspace` compiled it out. Your regression test never ran, and neither did the two create-did-webvh tests that predate it. Enabled `config-seed` on the dev-dep self-dep; all five run now. Both `cfg(not(config-seed))` sites are production fallbacks with nothing asserting them, so it costs no coverage.

**3. Added the service-array coverage** in both directions — advertised when `[messaging]` is set, absent when it isn't. Confirmed the first fails without the fix in (1).

**4. Dropped the `changelog.d/` fragment.** #938 retired that directory; entries now come from the commit subject via git-cliff. That makes the PR title the changelog entry, so I expanded it from "prevent dbl key deriv".

**Still open, deliberately:** the interactive path has the same bug. `operations/did_webvh/mod.rs` derives entity keys whenever `signing_key_id` is `None` — unconditionally, regardless of a caller-supplied document — so `build_interactive_did_document`'s preview keys (indices *n, n+1*) still differ from the ones the operation stores (*n+2, n+3*). A DID minted without `--url` still has the mismatch. Fixing it means the two sides sharing one derivation, which is a bigger change to `CreateDidWebvhParams` than belongs in this PR; I've written the limitation down at the call site. Worth a follow-up issue.

PR #948, opened 2026-08-12
fix(vault): send entryId on vault release, from both the CLI and the MCP bridge

### LINKED ISSUE #947: pnm vault release failing with protocol error on schema validation
## Summary

`pnm vault release <entry-id>` (and the equivalent `cnm` command) fails with a schema validation error because the CLI constructs the Trust Task payload with the key `id` instead of the required `entryId`.

### Step 1: Create the entry file

```json
{
  "contextId": "my-app",
  "targets": [
    { "kind": "web-origin", "origin": "https://api.github.com" }
  ],
  "label": "GitHub PAT for my-app",
  "secretKind": "bearer-token"
}
```

### Step 2: Create the secret file

```json
{
  "kind": "bearer-token",
  "token": "ghp_xxxxxxxxxxxxxxxxxxxx"
}
```

### Step 3: Store the entry

```bash
pnm vault upsert --entry-file entry.json --secret-file secret.json
```

### Step 4: Release the secret

```bash
pnm vault release vault_f6dfd09ba9ea41e2bba445ab7367843d
```

## Observed Error

```
✗ Protocol error: refusing to send a payload that does not conform to https://trusttasks.org/spec/vault/release/0.1: payload failed schema validation: Additional properties are not allowed ('id' was unexpected); "entryId" is a required property. The recipient would reject this as `malformedRequest`. An unset optional member must be ABSENT from the payload, not `null`.
```

## Root Cause

In `vta-cli-common/src/commands/vault.rs`, `cmd_vault_release` builds the payload with the wrong key:

```rust
// wrong
let mut payload = json!({ "id": id });
```

The `vault/release/0.1` schema requires `entryId`, which matches the server-side `VaultReleaseBody` struct (`entry_id: String` with `#[serde(rename_all = "camelCase")]`). All other release-family handlers (`handle_proxy_login`, `handle_sign_trust_task`) correctly use `entry_id` / `entryId`.

## Fix

```rust
let mut payload = json!({ "entryId": id });
```

## Scope

- **Affected**: `pnm vault release` and `cnm vault release` — both share `cmd_vault_release` in `vta-cli-common`
- **Not affected**: VTA service — `VaultReleaseBody` correctly deserialises `entryId`; the schema validation was working as intended
- **Version surface**: `vault/release/0.1` only; `0.2` has no dispatch handler registered yet

affrncsp:
cmd_vault_release was constructing the vault/release/0.1 Trust Task payload with key `id`, which fails schema validation. The schema requires `entryId` (matching VaultReleaseBody's camelCase serialisation on the server side).

Fixes #947

### DIFF
--- a/vta-cli-common/src/commands/vault.rs
+++ b/vta-cli-common/src/commands/vault.rs
@@ -169,7 +169,7 @@ pub async fn cmd_vault_release(
     id: String,
     target: Option<Value>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let mut payload = json!({ "id": id });
+    let mut payload = json!({ "entryId": id });
     if let Some(t) = target {
         payload["target"] = t;
     }
--- a/vta-cli-common/src/commands/vault.rs
+++ b/vta-cli-common/src/commands/vault.rs
@@ -169,11 +169,7 @@ pub async fn cmd_vault_release(
     id: String,
     target: Option<Value>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let mut payload = json!({ "entryId": id });
-    if let Some(t) = target {
-        payload["target"] = t;
-    }
-    let response = client.vault_release(payload).await?;
+    let response = client.vault_release_entry(&id, target).await?;
 
     // The released secret rides in a `didcomm-authcrypt` envelope; open it with
     // the caller's keys. Walk the documented shape
--- a/vta-mcp/src/server.rs
+++ b/vta-mcp/src/server.rs
@@ -245,11 +245,11 @@ impl VtaMcp {
         &self,
         Parameters(p): Parameters<VaultReleaseParams>,
     ) -> Result<CallToolResult, McpError> {
-        let mut payload = serde_json::json!({ "id": p.id });
-        if let Some(t) = p.target {
-            payload["target"] = t;
-        }
-        let response = self.client().vault_release(payload).await.map_err(to_mcp)?;
+        let response = self
+            .client()
+            .vault_release_entry(&p.id, p.target)
+            .await
+            .map_err(to_mcp)?;
         match response
             .get("sealedSecret")
             .and_then(|s| s.get("jwe"))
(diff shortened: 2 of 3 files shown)


stormer78:
Reviewed this — the fix is right. `VaultReleaseBody` (`vta-service/src/trust_tasks/vault.rs:297`) is `rename_all = "camelCase"` with a required `entry_id`, and `git log -S` shows it has been that since #420, so there's no older VTA that accepted `id` and no compatibility window to worry about.

The review turned up that the same bug was still live in the other client, so I've pushed a second commit onto this branch (`02235292`) rather than open a follow-up PR:

- **`vta-mcp/src/server.rs`** — `vault_release` built the same body with `id`, so every MCP host (Claude Desktop, agent frameworks) hit the identical `malformedRequest` reject the CLI did. Fixing only the CLI half would have left #947 half-open.
- **`vta-sdk/src/client/vault.rs`** — added `VtaClient::vault_release_entry(entry_id, target)`, which builds the body from a typed entry id, and pointed both callers at it. `vault_release(Value)` is unchanged and still there for callers shaping the whole request; this is additive, nothing breaks. Its doc comment said "entry `id` + optional `target`" — since the method takes an opaque `Value`, that comment was the only contract a caller had, and it's plausibly how two independent clients came to send the wrong key. Corrected.
- **Regression tests** — these validate the built body against the *published* `vault/release/0.1` schema from `trust_tasks_rs::schema_index`, not against a literal typed a second time. The schema is `additionalProperties: false` with `entryId` required, so it rejects both halves of the defect. A third test pins that the pre-fix `{"id": …}` body *is* rejected — without it, the other two would keep passing even if the schema stopped enforcing unknown members.

The underlying class here is the untyped `Value` boundary: the client/server pair is never exercised together, so a key mismatch is invisible to both the compiler and CI. `vault_release_entry` closes it for release; `vault_proxy_login` and `vault_sign_trust_task` sit on the same pass-through shape against `entry_id`-keyed bodies and could get the same treatment later.

Also updated the PR title, since the squash-merge subject becomes the changelog entry and the old one read as CLI-only.

@affrncsp heads-up — I pushed straight to your fork's `main`, so pull before you touch this branch again.

PR #959, opened 2026-08-12
feat(did-webvh): let a minted DID advertise TSP at the VTA's mediator

stormer78:
A VTA-minted DID could never advertise TSP, whatever the VTA's own config said.

`create_did_webvh`'s `add_mediator_service` publishes the VTA's mediator as a `DIDCommMessaging` service and nothing else (`did_webvh/document.rs`). A caller wanting `#tsp` had to hand-build the service entry and pass it through `additional_services` — which means knowing the VTA's mediator DID, the one thing `add_mediator_service` exists so a caller does not have to know. In practice nobody did, so every persona-shaped identity is DIDComm-only by construction.

## How this surfaced

OpenVTC PR [#211](https://github.com/OpenVTC/openvtc/pull/211): a join to a TSP-advertising community failed at the first hop with `400 w.m.message.deserialize` — a mediator built without its `tsp` feature feeding a CESR frame to the DIDComm JSON parser. Diagnosing it turned up the applicant persona's document:

```json
"service": [{ "id": "…#vta-didcomm", "type": "DIDCommMessaging",
              "serviceEndpoint": [{ "uri": "did:webvh:…:firstperson-mediator" }] }]
```

One entry. That is the shape this path *always* produces, so the workspace's both-ends rule — the protocol is the highest-preference one present in **both** parties' documents — can never resolve to TSP for such a DID. TSP could be on end to end and the intersection would still be DIDComm.

## What this adds

`add_tsp_service` on the create-DID wire, honoured by a new `with_tsp_service` in `did_webvh/document.rs`. When set, the minted document carries a `#tsp` (`TSPTransport`) entry pointing at the same mediator the DIDComm entry names — TSP advertises a mediator DID, not a transport URL (tsp-enablement.md D8), and the fragment and type match what the setup path and the runtime `services tsp enable` patcher emit, so a document minted here, minted at setup, or patched later are the same shape.

**Two gates, neither redundant:**

- **The caller's** (`add_tsp_service`). Opt-in, and deliberately *not* implied by `add_mediator_service`: a DID advertising a transport its holder cannot decode is unreachable over that transport, and only the caller knows whether the client behind this DID reads TSP frames. Flipping it implicitly would have minted that defect for every existing consumer at once.
- **Ours** (`[services] tsp` + a configured mediator). A VTA whose own stack does not run TSP must not mint documents claiming it does — that is the failure this change exists to prevent, not to spread. Asking without it is silently declined, and pinned by a test.

A caller who already hand-built a `TSPTransport` entry keeps theirs: two `#tsp` services would be a malformed document, and theirs is the more specific intent. Matched on the service `type`, never the `#id` fragment — the OWF reference implementation writes `#tsp-transport` where we write `#tsp`.

## Wire compatibility

Additive in both directions. The field is `skip_serializing_if` on the request and `Option` on the body, so a caller that does not set it produces a byte-identical request to one built before the field existed, and a VTA that predates it ignores an unknown key (`CreateDidWebvhBody` has no `deny_unknown_fields`). Same treatment `domain` got.

`CreateDidWebvhRequest` gains a field, so exhaustive struct literals in downstream crates need one more line — the usual 0.x breaking bump, handled by release-plz.

## Not in this PR

- **No operator flag.** `pnm did-mgmt dids create --add-tsp-service` means another positional on an already-18-argument helper plus clap wiring in both CLIs; the first consumer of this field is the SDK, so the flag belongs in its own change. Marked at the call site.
- **The mediator's `tsp` feature being off by default** (`default = ["didcomm", "redis-backend", "jemalloc"]`) is in `affinidi-tdk-rs`, not here. That build also answers a TSP frame with an error naming DIDComm and never mentioning TSP, which is what made #211 cost an hour. Worth a companion PR there.

## Verification

- `cargo test --workspace` — green
- `cargo clippy --workspace --all-targets` — clean (one pre-existing `backup_export` deprecation in `tests/e2e`)
- `cargo fmt` — clean
- Six new tests on `with_tsp_service`: the entry lands at the DIDComm mediator when asked; absent unless asked; **vetoed when `services.tsp` is off**; vetoed with no mediator configured; a caller-supplied `TSPTransport` is not duplicated (via the OWF fragment spelling, so type-matching is exercised); unrelated `additional_services` survive. The existing canonical-ordering test already pins that `#tsp` sorts ahead of DIDComm.


stormer78:
**semver checks (advisory, `continue-on-error: true`) fails here on purpose.**

```
Summary semver requires new major version: 1 major and 0 minor checks failed
  field CreateDidWebvhBody.add_tsp_service    (create.rs:147)
  field CreateDidWebvhRequest.add_tsp_service (client/types.rs:698)
```

Both structs are exhaustively constructible through the public API, so a new `pub` field breaks existing literals. That is the intended cost of putting the flag on the wire, and release-plz assigns the resulting `vta-sdk` 0.22 → 0.23. **Follow-up:** affinidi-tdk-rs then needs an "accept 0.23" chore so its `[patch.crates-io]` keeps applying — same shape as tdk #702 for 0.22.

The alternative that avoids the break is doing this client-side (resolve the VTA's own `#tsp` and pass the entry through `additional_services`), which was rejected: it puts the "does this VTA actually run TSP" gate in every client instead of once in the VTA, and leaves nothing server-side to stop a client advertising a transport the VTA does not serve.

Every other check on this run is green (12/13).

PR #962, opened 2026-08-13
feat(release): publish vta-service and its closure again

stormer78:
This reverses the publishing half of #938. Deliberately, and with the reason it was wrong.

## The audit missed a dev-dependency

#938 set `publish = false` on `vta-service` and the twelve subsystem crates behind it, on the finding that nothing external depended on them. The audit read normal dependencies. `openvtc-core` depends on `vta-service` as a **dev-dependency**, for `test_support::MockVta` — an in-process VTA that OpenVTC's end-to-end tests run against. That harness boots the real service (1,959 lines of `vta-service/src/test_support.rs`), so no client crate can substitute for it.

## Unpublishing did not freeze the crate — it broke it

`vti-common/src/acl/mod.rs:272`:

```rust
pub use vta_sdk::acl::{ActScope, ApproveScope, ContextDirection};
```

**A re-export makes the re-exported crate's version part of your public API.** Any graph combining `vti-common` with another `vta-sdk` consumer must resolve **one** `vta-sdk`. Frozen at 0.14.37, `vta-service` asks for `vta-sdk ^0.21`; `vti-common` has since moved to `^0.23`. A downstream `cargo update` resolves both, and `vta-service` fails to compile at ten call sites:

```
error[E0308]: mismatched types
  --> vta-service-0.14.37/src/messaging/handlers.rs:823
     expected `vti_common::acl::ApproveScope`, found `vta_sdk::acl::ApproveScope`
```

Identical at 0.14.24, so it is the surrounding crates moving, not any version bump. This is how it surfaced — in [openvtc#213](https://github.com/OpenVTC/openvtc/pull/213), on a routine dependency refresh. Nothing downstream can fix it: only a release that moves the requirements together can.

## What changes

Thirteen manifests return to the workspace default: `vta-service` plus its closure — `vta-audit`, `vta-backup`, `vta-config`, `vta-keys`, `vta-keyspaces`, `vta-policy`, `vta-support`, `vta-sweepers`, `vta-tee`, `vta-vault`, `vta-webvh`, `vti-webauthn`.

The cost is that closure: crates.io requires every dependency of a published crate to be published, so twelve subsystem crates go back on the registry — exactly what #938 set out to stop. Taken over the two alternatives, both worse:

- **Yank the published copies.** Honest about support status, but breaks OpenVTC's `mockvta_bootstrap_e2e` immediately, with nothing to replace it.
- **Leave them up.** Ships a crate on the registry that cannot be built.

Six crates stay internal and are untouched: `vtc-service`, `vta-enclave`, `vta-mcp`, `vta-mobile-core`, `didcomm-test`, `vti-fuzz`.

## Release ordering — expect one dry-run failure until this lands

`cargo publish --dry-run -p vta-service` **fails today**, and will until the closure is on the registry. Packaging strips path dependencies, so `vta-keys = "0.2"` resolves the *published* 0.2.1, which still asks for `vta-sdk ^0.21` — two nodes, same error. Confirmed in the packaged lockfile:

```
name = "vta-sdk"   version = "0.21.21"
name = "vta-sdk"   version = "0.23.0"
```

That resolves itself in the release, which publishes in dependency order. Every subsystem crate in this workspace **already** requires `vta-sdk = "0.23"` — release-plz kept them coherent while they were unpublished, exactly as its config comment promises — so once they upload, `vta-service` verifies against them rather than against the stale copies.

Verified that the layers behave as that predicts: `vta-keyspaces` and `vta-config`, whose dependencies are all published already, dry-run clean.

Note also that every crate here is currently at the same version as its published copy, so each needs a bump before it can upload. This PR touches all thirteen manifests, so release-plz will bump them; no version is edited by hand.

## Docs

`CLAUDE.md`, `RELEASING.md` and the `release-plz.toml` header all said 7-of-21. They now say 20-of-26, name the six that stay internal, and record the rule the audit missed: **check dev-dependencies, in sibling repos, before unpublishing anything** — and that a re-export ties you to a version.

## Verification

- `cargo metadata` parses; `cargo check --workspace` clean
- `cargo publish --dry-run` clean on `vta-keyspaces` and `vta-config`
- `cargo publish --dry-run -p vta-service` fails as described above, for the stale-registry reason, with the two-node lockfile as evidence
- `cargo fmt` clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,8 +281,34 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-semver-checks
 
+      # Excludes the twelve subsystem crates. They are on crates.io only
+      # because a published crate's whole dependency closure must be published
+      # and `vta-service` is published (for OpenVTC's `MockVta` harness — see
+      # RELEASING.md). Nothing consumes their APIs directly, here or outside.
+      #
+      # The exclusions are not cosmetic: this job builds rustdoc JSON for every
+      # crate twice, current and baseline. Including them took it from ~17
+      # minutes to over 35 — on every pull request in the repo, for a signal no
+      # consumer can act on. `release-plz.toml` carries the matching
+      # `semver_check = false` so the release decides bumps the same way.
+      #
+      # `vta-service` is NOT excluded: it has a real consumer, so a silent
+      # break there reaches somebody.
       - name: Check published crates for API breaks
-        run: cargo semver-checks --workspace
+        run: |
+          cargo semver-checks --workspace \
+            --exclude vta-audit \
+            --exclude vta-backup \
+            --exclude vta-config \
+            --exclude vta-keys \
+            --exclude vta-keyspaces \
+            --exclude vta-policy \
+            --exclude vta-support \
+            --exclude vta-sweepers \
+            --exclude vta-tee \
+            --exclude vta-vault \
+            --exclude vta-webvh \
+            --exclude vti-webauthn
 
   lockfile-self-pins:
     # Sibling guard to release-guards: the version bump can be correct and the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -868,13 +868,36 @@ These are load-bearing — know they exist before adjusting nearby code.
 the Release PR that release-plz maintains; merging that PR is what publishes.
 Merging a feature PR publishes nothing. See [`RELEASING.md`](RELEASING.md).
 
-**Only 7 of 21 crates publish** — `vta-sdk`, `vti-common`, `vti-secrets`,
-`vta-cli-common`, `vtc-client`, `pnm-cli`, `cnm-cli`. The other 14 set
-`publish = false` in their own manifest: an audit in #938 found zero external
-reverse-dependencies and no sibling repo consuming them. Before adding a crate
-back to the published set, check that everything it depends on is published too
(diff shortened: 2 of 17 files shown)


stormer78:
**`semver checks` fails here on purpose, and its content is the argument for this PR.**

```
Checking vta-service v0.14.37 -> v0.14.37 (no change; assume minor)
  field CreateDidWebvhParams.add_tsp_service   did_webvh/mod.rs:367
  field CreateDidWebvhParams.pre_derived       did_webvh/mod.rs:393
Summary semver requires new major version: 1 major and 0 minor checks failed
```

`add_tsp_service` is #959's deliberate break. **`pre_derived` is the one to look at**: it was added while the crate was unpublished, so no version consequence was ever recorded for it — the crate's API moved while its version stood still. That is the same drift that left the frozen 0.14.37 unbuildable against a current `vti-common`, and re-publishing is what makes it visible again. release-plz will now assign the compatibility bump (0.14.x → 0.15.0) from this signal.

The job is `continue-on-error: true`; the other 14 checks pass.

**Duration:** `3ad2af5a` also brings this job back to ~18 minutes (08:13:31 → 08:31:27). Before the exclusions it ran 35+ minutes, because republishing added thirteen crates to `cargo semver-checks --workspace`. Note the CI job invokes `cargo semver-checks` directly, so `--exclude` in `ci.yml` is what shortens it — the matching `semver_check = false` in `release-plz.toml` only governs how the release picks bump sizes.

PR #996, opened 2026-08-16
chore(deps)!: track trust-tasks 0.9 across the workspace

stormer78:
## What

Moves all five `trust-tasks-*` requirements plus `trust-tasks-capability-client` (0.5 → 0.8) together, and takes the `affinidi-messaging-*` releases built on 0.9 alongside them.

| dependency | from | to |
|---|---|---|
| `trust-tasks-rs` | 0.6 | **0.9** |
| `trust-tasks-https` | 0.6 | **0.9** |
| `trust-tasks-didcomm` | 0.6 | **0.9** |
| `trust-tasks-proof` | 0.6 | **0.9** |
| `trust-tasks-capability-client` | 0.5 | **0.8** |
| `affinidi-messaging-sdk` (transitive) | 0.19.7 | 0.19.8 |
| `affinidi-messaging-mediator` (transitive) | 0.18.17 | 0.18.18 |

They move as one because `trust-tasks-rs`'s core types cross the public API of `-https` / `-didcomm` / `-proof` — a graph mixing majors does not type-check.

## What cost nothing

- **No `consume_inbound` call sites in this workspace**, so 0.9.0's new required `PayloadPolicy` argument does not reach us.
- **Nothing matches on `StandardCode`**, so 0.7.0's `#[non_exhaustive]` does not either.
- The `validate` feature stays enabled and unused, exactly as before.

## What did change: the error version on the wire

`trust-task-error` moved 0.3 → 0.4 → 0.5 upstream. Each step happened for the same reason the 0.3 step did — a new standard code the older payload schema's `code` enum does not list and whose extended-code pattern does not match, so a document carrying it would not validate as the older version. `0.4` carries `idConflict`, `0.5` carries `cancelled` (SPEC §8.3).

Both services hand-write that version on their **one unrouted path** — where there is no request document to reject from, and therefore no framework call to ask. Both were left naming `0.3` while `reject_with` stamped `0.5` on every *routed* rejection. One service emitting two versions is a trap for exactly the consumer that pins one of them.

**`unrouted_and_routed_errors_agree_on_the_type_uri` exists in both services to catch this, and it did.** The bump failed those two tests rather than shipping two dialects. That is the test doing its job; constants updated and their rationale extended.

## Fixture tidy-up (and why the census count was not raised)

Two in-`src` VTC test fixtures also named `0.3`. They now take the version from `framework_error_type_uri()` rather than repeating it, so they follow the emitter on the next bump instead of stranding a version behind.

That also keeps `trust_task_manifest`'s unpublished-URI census at **one** URI for the family. The census is deliberately exact so the debt can shrink but never grow unnoticed, and its own failure message says raising the expected count is the wrong fix — so I didn't.

**Backward acceptance of an older error document keeps its coverage:** `vtc-service/tests/registry_didcomm.rs` pins `0.1` on purpose and is left alone.

## Compatibility

Per SPEC §5.2 forward-minor compatibility, a consumer still pinned to `trust-task-error/0.3` SHOULD accept `0.5`. Marked `!` because this changes the version on the wire, not because any Rust signature moved.

## Verification

- `cargo test --workspace` — all suites pass, 0 failures
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets` — clean
- `cargo tree -d -e normal,build` lists **none** of `trust-tasks-rs`, `trust-tasks-capability-client`, `vta-sdk`, `vti-common`, `affinidi-tdk`
- `scripts/check-lockfile-self-pins.sh` — no self-pin
- **No `version =` field touched**, per RELEASING.md — versions are the Release PR's job

### Unrelated pre-existing breakage found while verifying

`cargo check --workspace --all-features --all-targets` fails in `vta-sdk/tests/provision_client_e2e.rs` (5 × `E0609`). **Not caused by this PR** — it reproduces on `main`. PR #949 (`fix(provisioning)!: relay the holder's bootstrap VP as raw JSON`) changed `ProvisionIntegrationRequest::request` to `serde_json::Value` on 2026-08-12 without updating that test, and no CI job builds `--all-features --all-targets`, so it went unnoticed. Left for a separate fix; flagging it here rather than folding it in.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,9 +356,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-mediator"
-version = "0.18.17"
+version = "0.18.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95435c2e6cc192765e0dd4d7d00aa8995869cce73040852d3035c35ca8018e53"
+checksum = "01857ae34f9931c8d0335d17a8f69917f22ae8e6fb3f09e1c7bfa56bb1343b62"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-common",
@@ -470,9 +470,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-sdk"
-version = "0.19.7"
+version = "0.19.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a793b4479720c074d56ab2c0e9ed272987bf679b0baa130160718512a93079c8"
+checksum = "954fe9bbc880aa008e9f01eeac117ab45419df00fddc40541c975fc70b3de506"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-authentication",
@@ -508,9 +508,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-test-mediator"
-version = "0.2.50"
+version = "0.2.51"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "60b360ddb78e934a51d9ad7343e0fa0fbe25b60a4e1dabf4f40b25955bc6ceba"
+checksum = "1004ee041b061e9b91f440fee6db14ca003f0881933faab17ba05d2442eb8910"
 dependencies = [
  "affinidi-did-resolver-cache-sdk",
  "affinidi-messaging-didcomm",
@@ -1485,9 +1485,9 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-http-client"
-version = "1.2.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "635d23afda0a6ab48d666c4d447c4873e8d1e83518a2be2093122397e50b838e"
+checksum = "3c1c8a04cb31ba74d0115af5a890bb8c0d48fba64b52812fa13929a6ef0cc83c"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-runtime-api",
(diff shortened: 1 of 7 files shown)


stormer78:
### Note on the one red check — `semver checks`

Two separate things are in that job, and **neither is fixable before merge**. Everything else is green (Test, Feature combos, Clippy, MSRV, Format, cargo-deny, commit lint, lockfile self-pins, Enclave build, both Docker images, Mobile build).

#### 1. Three breaking changes inherited from `main`, not introduced here

```
vta_cli_common::commands::keys::cmd_key_create now takes 8 parameters instead of 6
field CreateKeyBody.internal            (vta-sdk/src/protocols/key_management/create.rs)
field CreateKeyRequest.internal         (vta-sdk/src/client/types.rs)
variant KeyOrigin:Internal              (vta-sdk/src/keys/mod.rs)
field QueryBody.oid4vp_session          (vta-sdk/src/protocols/credential_exchange.rs)
```

All of these come from **#995** (`feat(vta-keys): add non-extractable internal signing keys`) and **#993** (`feat(vta-service)!: present ISO mdoc credentials over OID4VP`), which landed immediately before this branch was cut. Nothing in this PR's diff appears in that list.

Both of those PRs **also failed `semver checks` and were merged** — release-plz carries the version bump. This branch simply inherits their unreleased breaks.

#### 2. The baseline build cannot resolve, and that is this PR's whole point

```
error[E0308]: mismatched types
   --> .../registry/src/.../vta-sdk-0.24.0/src/acl_setup.rs:147:18
    |            Some(acl),
    |                 ^^^ expected `MediatorAcl`, found a different `MediatorAcl`
note: there are multiple different versions of crate `trust_tasks_rs` in the dependency graph
   --> .../trust-tasks-rs-0.9.0/src/specs/messaging/account/update/v0_1.rs:409
   ::: .../trust-tasks-rs-0.6.5/src/specs/messaging/account/update/v0_1.rs:409
```

`cargo-semver-checks` builds `vta-service`'s rustdoc against the **published baseline** `vta-sdk 0.24.0`, pulled from crates.io — and that published copy is built on `trust-tasks-rs` **0.6.5**. Put it in a graph that also carries **0.9.0** and the two `MediatorAcl` types cannot unify.

That is exactly the "a graph mixing majors does not type-check" failure this PR exists to remove, appearing transiently in the baseline comparison because the registry does not yet have a `vta-sdk` built on 0.9 — **this PR is what produces one.** Note the path: the error is inside a *registry* copy of `vta-sdk`, not the workspace source.

It resolves on merge + publish, and cannot resolve before it. The workspace itself is clean:

```
$ cargo tree -d -e normal,build
# lists none of trust-tasks-rs, trust-tasks-capability-client, vta-sdk, vti-common, affinidi-tdk
```

PR #999, opened 2026-08-18
fix(webvh): read agent-name createdAt in both shapes the host sends

stormer78:
## Problem

A DID with at least one agent name could not be listed over DIDComm. The whole task failed with:

```
agent-name list response parse error: invalid type: string "2026-08-18T08:58:13Z", expected u64
```

which reached the operator as an internal error from a `set` that had in fact **succeeded** — the name was already in the DID document while the UI reported failure and showed an empty registry. Empty registries parsed fine, so this only appeared *after* a successful claim.

## Why both producers are right

`AgentNameEntryWire` is documented as the shape of the REST `GET /api/dids/{mnemonic}` `agentNames` array, where `createdAt` is Unix seconds as a number. `webvh_didcomm::list_agent_names` reuses it to parse the DIDComm listing — where it is not: the published payload schema for `did-management/agent-name/list/0.1` types `createdAt` as `"string"` with `"format": "date-time"`, and the host projects it that way (`affinidi-webvh-service`, `did-hosting-control/src/messaging.rs`, since 2026-07-29).

Neither side is wrong, and the mismatch is not fixable by picking one. The **same host emits both shapes**: its DID-record projection serialises the stored registry (numbers) while the list verb answers to the spec (strings). The consumer is the single place that sees both.

## Fix

The wire type accepts either and normalises to seconds. The VTA relays a `u64` outward in the canonical `AgentNameEntry`, so an RFC3339 input is converted rather than propagated — a deliberate asymmetry between what this reads and what it emits.

A third shape is a genuine contract break and still fails, quoting what arrived so it can be told apart from an unreachable host (R6.4).

`chrono` moves from dev-dependencies to dependencies for the conversion. It was already in the build graph via `vta-service`.

## Tests

* Both wire shapes decode (epoch number; RFC3339 string).
* An offset timestamp normalises to the same instant as its UTC twin.
* A two-entry listing decodes as an array — the shape `list_agent_names` actually deserialises.
* An uninterpretable timestamp fails with the offending value in the message.

`cargo test -p vta-webvh` (53) and `cargo test -p vta-service --lib` (823) pass; `cargo clippy -p vta-webvh --all-targets` is clean. (The workspace has a pre-existing `dead_code` warning in `vti-common` unrelated to this change, which trips a `-D warnings` run.)

## Not covered

No test drives the DIDComm listing end to end against a host. The parse is unit-covered at the type, but the producer/consumer pairing that broke here is still only checked by hand — the same gap that let this ship.

## Deployment

The fix reaches clients through a `vta-service` release and redeploy; consumers such as OpenVTC need no dependency change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1493,7 +1493,7 @@ dependencies = [
  "aws-smithy-runtime-api",
  "aws-smithy-types",
  "h2 0.3.27",
- "h2 0.4.15",
+ "h2 0.4.16",
  "http 0.2.12",
  "http 1.5.0",
  "http-body 0.4.6",
@@ -4216,9 +4216,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.15"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6cb093c84e8bd9b188d4c4a8cb6579fc016968d14c99882163cd3ff402a4f155"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
 dependencies = [
  "atomic-waker",
  "bytes",
@@ -4536,7 +4536,7 @@ dependencies = [
  "bytes",
  "futures-channel",
  "futures-core",
- "h2 0.4.15",
+ "h2 0.4.16",
  "http 1.5.0",
  "http-body 1.1.0",
  "httparse",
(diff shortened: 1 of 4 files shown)


stormer78:
CI triage: the **cargo-deny** failure is not from this PR.

RUSTSEC-2026-0258 (h2 < 0.4.16 — unbounded empty DATA frames, low severity, reached transitively through hyper) was published after this branch was cut, and `cargo deny check` fails identically on main today. `bans`, `licenses` and `sources` all pass, so the `chrono` move from dev-dependencies to dependencies in this PR is clean.

Worth doing as its own dependency PR rather than bundling: in this workspace `cargo update -p h2` also re-selects a set of `windows-sys` and `syn` references while rewriting the lockfile, which is a graph change that should be reviewed on its own.

PR #1002, opened 2026-08-19
ci: bound every job, and stop superseded PR runs racing the new one

stormer78:
Two gaps that surfaced together on #1000, where CI sat for an hour and reported
nothing.

## No job had a timeout

All five long jobs were stuck on the **same step** for sixty minutes:

```
in_progress  Install system dependencies     ← 60 min
pending      Run dtolnay/rust-toolchain@stable
pending      Run actions/cache@v5
pending      Run cargo test --workspace
```

That step is `sudo apt-get update && sudo apt-get install -y libdbus-1-dev
pkg-config`. They never reached the Rust toolchain, let alone the tests — an
apt hang on GitHub's runners.

With no `timeout-minutes`, that hangs indefinitely, and an indefinite hang is
indistinguishable from a slow build until someone reads the step list by hand.
I only found it because I went looking; the PR just looked slow.

Ceilings are ~2× each job's observed p100, taken from real runs rather than
guessed:

| Job | Observed | Ceiling |
|---|---|---|
| Mobile build | 22m | 45 |
| Feature combos | 20m | 45 |
| semver checks | 17–22m | 45 |
| Test | 13m | 30 |
| Clippy | 4m | 20 |
| MSRV / Enclave build / Check | 3m | 20 |
| Format, cargo-deny, lockfile self-pins, commit lint | <1m | 10 |

The headroom is deliberate. These exist to bound a **hang**, not to enforce a
performance budget — a job that fails because a runner had a slow day is worse
than no limit at all.

## No `concurrency` group

Pushing a fixup to an open PR left the superseded run executing and racing the
new one for the same runner pool. #1000 had two full runs live simultaneously,
and the older one was still churning through jobs the newer one had already
finished.

```yaml
concurrency:
  group: ci-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

**Cancellation is scoped to pull requests on purpose.** A push to `main` or
`nightly` gates releases and is the record of what that commit did, so those
runs are left to finish even when another lands on top. Only PR runs — where a
superseded result tells you nothing worth waiting for — get cancelled. This is
slightly more conservative than the unconditional `cancel-in-progress: true`
the browser-plugin repo uses.

## Scope

`ci.yml` only.

- `docker.yml` and `join-didcomm-soak.yml` already carry timeouts.
- **`publish.yml` and `release-mobile-ios.yml` have neither, and are
  deliberately untouched.** `cancel-in-progress` on a workflow that publishes
  crates could kill a publish part-way through — a worse failure than the one
  being fixed here. They want their own change, made by someone looking at the
  release path properly.

## Test

YAML parses; 12 jobs, all now carrying `timeout-minutes`, verified
programmatically rather than by eye:

```
YAML parses; 12 jobs
concurrency: {'group': 'ci-${{ github.ref }}',
              'cancel-in-progress': "${{ github.event_name == 'pull_request' }}"}
no timeout: []
```

The concurrency behaviour proves itself on this PR: pushing again should cancel
the previous run rather than leave it racing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,24 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Install system dependencies
-        run: sudo apt-get update && sudo apt-get install -y libdbus-1-dev pkg-config
+        run: |
+          # The observed failure here is a *hang*, not a clean error: apt sits
+          # on an unresponsive mirror and never returns. Three separate runs
+          # stalled on this step, one of them for a full hour before anyone
+          # looked. A bare retry loop is no use against that — it would wait
+          # forever on the first attempt — so each attempt is bounded with
+          # `timeout`, which turns the hang into a failure the loop can act on.
+          # Acquire::Retries handles the ordinary transient case underneath.
+          for attempt in 1 2 3; do
+            if sudo timeout 90 apt-get -o Acquire::Retries=3 update \
+              && sudo timeout 120 apt-get -o Acquire::Retries=3 install -y libdbus-1-dev pkg-config; then
+              exit 0
+            fi
+            echo "::warning::apt attempt $attempt failed or timed out; retrying in 10s"
+            sleep 10
+          done
+          echo "::error::apt-get failed after 3 bounded attempts"
+          exit 1
 
       - uses: dtolnay/rust-toolchain@stable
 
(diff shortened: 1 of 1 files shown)


stormer78:
## Update: bounding the hang wasn't enough

The timeouts from the first commit fired on the very next run — `Check` and
`MSRV` cancelled at **20:19** (08:18:09 → 08:38:28), both on
`Install system dependencies`. That confirms the diagnosis and simultaneously
shows the fix was incomplete: three separate runs have now stalled on that
step, one of them for a full hour before I noticed.

So this PR now also hardens the step itself.

**The failure mode drives the fix.** apt doesn't fail here — it *hangs* on an
unresponsive mirror and never returns. A plain retry loop is useless against
that, because it would wait forever on the first attempt. Each attempt is
therefore bounded with `timeout`, which converts the hang into a failure the
loop can act on:

```bash
for attempt in 1 2 3; do
  if sudo timeout 90 apt-get -o Acquire::Retries=3 update \
    && sudo timeout 120 apt-get -o Acquire::Retries=3 install -y libdbus-1-dev pkg-config; then
    exit 0
  fi
  echo "::warning::apt attempt $attempt failed or timed out; retrying in 10s"
  sleep 10
done
echo "::error::apt-get failed after 3 bounded attempts"
exit 1
```

`Acquire::Retries=3` covers the ordinary transient case underneath the
`timeout`.

**Budget.** Worst case is ~11 minutes before the step gives up with a clear
`::error::`. That fits inside the tightest ceiling this touches (20 min) while
leaving room for the work itself, which runs 3–4 minutes on those jobs.

Applied to all **seven** jobs carrying the step.

**Verified** rather than eyeballed: YAML parses, all 12 jobs still carry
`timeout-minutes`, `unhardened apt: []`, and the loop body passes `bash -n` and
retries as intended when the command fails.

The two layers are complementary — the retry should mean the timeout rarely
fires again, and the timeout is still the backstop for a hang that outlasts all
three attempts.

PR #1003, opened 2026-08-19
fix(tee): bootstrap 410 and vsock enotconn

oleksiipiliugin:
tokio-vsock can report a stream connected just before Nitro finishes the
nonblocking handshake, so the very first read on a fresh vsock:5800
connection to the parent config server can return ENOTCONN even though
the parent is listening and ready. Retry only that specific transient
error kind with a short delay; any other I/O error still fails closed
immediately, and the existing overall READ_TIMEOUT deadline still
bounds the whole fetch.

Adds positive (retries ENOTCONN then succeeds) and negative (does not
retry PermissionDenied) unit tests against the inner read loop.

`POST /bootstrap/request` returned 403 Forbidden once the TEE first-boot
carve-out had already been claimed, but the documented API contract
requires 410 Gone. A 403 tells the caller "you specifically are not
authorized", which is the wrong signal for a single-use resource that no
caller, with any credentials, can ever claim again — 410 is what tells a
client the retry itself is pointless rather than "try different
credentials".

Adds `AppError::Gone` (renders as HTTP 410) to `vti-common`, alongside the
existing `Conflict`/`NotFound` variants, and switches both closed-carve-out
branches in `mint_mode_b` (the normal "sentinel already exists" check, and
the defensive concurrent-claim-loss path) onto a single
`carve_out_closed_error()` helper that returns it. This is a wire
compatibility fix, not a security fix — the carve-out was already correctly
refusing the second request; only the status code was wrong. The SDK's
`vta_sdk::error::VtaError::Gone` (mapped from HTTP 410 via `from_http`, with
an operator-facing "provision-integration a new operator instead" hint) was
already written expecting this status, so this closes the gap on the server
side.

Adds regression tests: `vti-common` pins the 410 status + body for the new
variant, and `vta-service` drives the real `mint_mode_b` end-to-end twice
against a simulated TEE provider (first call mints an admin and closes the
carve-out; second call, with a different client key and nonce, asserts
`AppError::Gone` / HTTP 410).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -400,6 +400,16 @@ jobs:
           cargo test -p vta-service --bin vta --features config-seed setup::interactive
           cargo test -p vta-service --bin vta --features config-seed,vault-secrets,k8s-secrets setup::interactive
 
+      # The TEE Mode B carve-out. `tee` is not a default feature of
+      # `vta-service`, so `cargo test --workspace` never builds
+      # `routes::bootstrap::tests` — and the `cargo check` step above builds
+      # the lib only, so it does not even TYPE-CHECK them. That module holds
+      # the two properties CLAUDE.md calls load-bearing: exactly one admin is
+      # ever minted, and a consumed carve-out refuses with 410 rather than a
+      # code that invites a retry. Both were unrunnable in CI. Run them.
+      - name: vta-service TEE Mode B carve-out
+        run: cargo test -p vta-service --no-default-features --features tee,rest,didcomm --lib routes::bootstrap
+
       # The fleet (`BAKE_CONFIG=false`) tenant-overlay path. `tenant-overlay` is
       # not a default feature of `vta-tee` and nothing in the default workspace
       # graph turns it on, so `cargo test --workspace` never even COMPILES
--- a/vta-backup/src/ops/blob.rs
+++ b/vta-backup/src/ops/blob.rs
@@ -32,9 +32,9 @@ use vti_common::store::KeyspaceHandle;
 /// - bundle_id not found → `NotFound`
 /// - token doesn't match the stored hash → `Forbidden`
 /// - bundle is an import bundle, or already-downloaded / aborted /
-///   expired → `NotFound` / `Conflict` (see [`enforce_export_ready`])
-/// - bundle expired by clock → `Conflict` (410-style, via [`gone`])
-/// - blob bytes missing on disk → `Conflict` (already swept)
+///   expired → `NotFound` / `Conflict` / `Gone` (see [`enforce_export_ready`])
+/// - bundle expired by clock → `Gone` (410, via [`gone`])
+/// - blob bytes missing on disk → `Gone` (already swept)
 pub async fn read_export_blob(
     bundles_ks: &KeyspaceHandle,
     bundle_id: Uuid,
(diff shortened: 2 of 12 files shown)


stormer78:
## Review

Both fixes are correct, and I've pushed a follow-up commit (15c02c95) closing the surfaces they left open rather than blocking on them.

The 410 is right and the server was the side that was wrong: `vta_sdk::error.rs:224` already maps 410 → `VtaError::Gone`, `vta-cli-common/src/render.rs` already had a hint keyed off it, and `CLAUDE.md:441` already documented "subsequent calls return 410". Nothing in the workspace depended on the old 403 — no test, no doc, no CLI branch. The ENOTCONN retry is bounded by the existing `READ_TIMEOUT` and every other error kind still fails closed.

### What the follow-up commit changes

**The blob endpoints had the identical bug, already flagged in-tree.** `vta-backup/src/ops/blob.rs` carried this:

```rust
fn gone(message: String) -> AppError {
    // `AppError` doesn't have a `Gone` variant. The blob endpoints
    // want 410 specifically ... Map via `Conflict` for now —
    // a follow-on can add a typed variant if the CLI surface requires it.
    AppError::Conflict(message)
}
```

Both blob routes have declared `410` in their `utoipa` annotations since they were written, so the code was contradicting its own published contract — the same defect this PR fixes for bootstrap, and this PR is the follow-on that comment was waiting for. Switched to `AppError::Gone`, with the module doc, the `POST` 410 description and the test that pinned 409 updated. The `ImportReceived`/`Previewed`/`Committed` arm deliberately stays 409 — that bundle is progressing through the import flow, not consumed.

**The new 410 test never ran in CI, and wasn't even type-checked.** `routes::bootstrap`'s test module is `#[cfg(all(test, feature = "tee"))]`; the only `tee` step in the workflow is a `cargo check` without `--all-targets`, so it builds the lib only. Nothing in CI ran or compiled it — the same for `carveout_claim_admits_exactly_one_concurrent_minter`, which is one of the properties CLAUDE.md calls load-bearing. Added a step to the Feature combos job, alongside the existing tenant-overlay and transport-harness steps that exist for exactly this reason.

**The ENOTCONN arm applied to every read, not just the first.** The constant's doc and the commit message both say "the first read", but the arm sits inside the loop. A parent that dies mid-envelope would spin out the rest of `READ_TIMEOUT` and then report `"envelope not fully received within 10s"` instead of the errno that explains it. Now gated on `buf.is_empty()` — once a byte has arrived the connection demonstrably completed, so a later ENOTCONN is a real teardown. The mock became a scripted reader so the mid-stream case is testable.

**The retry was silent.** `connect_with_retry` warns per attempt; this loop logged nothing, so a boot that spent 9.9s spinning looked identical to one that read immediately. One latched `warn!`, not one per 10ms tick.

**`Gone` fell through to `internal-error` in four mappers** — `vta-service` and `vtc-service`, trust-task and DIDComm in each. Not reachable on either transport today, but the fallback is the wrong default: a consumed single-use resource is a terminal caller-visible outcome, and reporting it as a server fault tells the client to retry something that can never succeed. No `e.p.msg.gone` wire code invented for a variant no DIDComm surface produces; it rides with `conflict`.

Two consequences of `Gone` gaining a second producer: the CLI now runs the 410 body through `extract_human_message` (matching the 409 arm beside it) and only prints the carve-out walkthrough when the refusal is actually the carve-out — otherwise an operator whose backup bundle expired was being sent to provision an admin. `VtaError::suggested_fix`'s `Gone` string was softened for the same reason.

Also reverted the `Mode B:` → `Mint Mode B:` doc rename, which reads as a mode named "Mint Mode B".

### Checked and deliberately not changed

`vti-common/src/store/vsock.rs` does the same connect → write → `read_exact` on the storage proxy, so I expected it to need the same treatment. It doesn't: `VsockStore::send` and `VsockKeyspaceHandle::send` already catch any request error, drop the connection, reconnect and retry, so a first-I/O ENOTCONN there self-heals.

One thing to be aware of rather than act on: `vti_common::AppError` isn't `#[non_exhaustive]`, so adding `Gone` is `enum_variant_added` — semver-major for a published crate. The semver checks job reports rather than fails, and release-plz reads the same signal, so the bump lands correctly either way.

### Verification

Ran locally against the branch:

- `cargo clippy --workspace -- -D warnings` — clean
- `cargo test --workspace` — 80 suites, 0 failures
- `cargo test -p vta-tee --features tenant-overlay tenant_overlay` — 9 pass (was 8)
- `cargo test -p vta-service --no-default-features --features tee,rest,didcomm --lib routes::bootstrap` — 4 pass, which is the run CI could not do before this commit

PR #1018, opened 2026-08-20
chore(deps)!: move to trust-tasks-rs 0.11, and fix the wire drift it exposed

stormer78:
Started as "did the sibling release fix the duplicate `trust-tasks-rs`?" — yes.
But the upgrade turns on schema validation for 22 tasks that previously had
none, and **three of them were sending non-conforming payloads**.

## The graph

`trust-tasks-https` / `-didcomm` / `-proof` / `-tsp` / `-capability-client` /
`-didcomm-v1` now all require `^0.11`, and `affinidi-messaging-sdk` 0.19.9 /
`-mediator` 0.18.20 followed. Until every one agreed, bumping the direct
dependency put **two** `trust-tasks-rs` nodes in the graph and the types crossed:

```
error[E0308]: expected `MediatorAcl`, found a different `MediatorAcl`
note: there are multiple different versions of crate `trust_tasks_rs`
```

Refreshing the lock to the versions that agree collapses it to one node.

## Three live defects, all the same class as #919

They were shipping before this PR. Nothing failed only because neither end had a
schema to check against.

1. **`CreateDidWebvhRequest` serialized snake_case.** Canonical wire is camelCase
   (R3.1) and the schema is `additionalProperties: false`. It worked solely
   through `CreateDidWebvhBody`'s back-compat aliases — which is exactly what made
   it invisible.
2. **`update_context` sent `null`** for unset members, where the schema types them
   as optional *strings*. Now built from `UpdateContextBody`, which skips `None`.
3. **`WebvhPathMode` emitted `auto_assign`** where the schema pins `autoAssign`
   with a `const` per `oneOf` branch.

**Number 3 is the one worth your attention.** The drift was in a *discriminant
value*, not a member name — so it slipped past every review and every convention
that watches member casing. And it was only reachable after fixing #1: the payload
had to get past the field-name check before the value check could fail. One drift
was hiding another.

Aliases keep the retired spellings accepted, so a peer built against the old shape
still parses; only what we **emit** moves. A round-trip test pins both directions,
and the REST test that posts the old spelling is annotated as the back-compat
coverage it now is — please don't let anyone "modernise" it.

## Four census guards, all shrinking

Every debt list in the workspace fired, each in the good direction:

| Guard | Change |
|---|---|
| manifest debt (`UNPUBLISHED_CANONICAL_OK`) | 36 → 14 |
| conformance witnesses | +22, each schema-validated with the mandatory non-vacuity check |
| `UNSPECCED_DISPATCHED_URIS` | −22 |
| producer-payload `UNPUBLISHED` | −2 |

I **de-rotted one test rather than re-pinning it**. `an_unspecced_task_proceeds_by_default`
hard-coded `webvh/dids/create` as its unspecced example — which this PR publishes,
so it failed for the best possible reason. It now derives one from the debt list at
runtime, so it follows the debt down instead of breaking every time a spec lands,
and returns early with a note to delete itself when the list empties.

## Follow-up I'd propose rather than bolt on

The 22 new witnesses use hand-written `json!`, which proves *the schema accepts a
shape*. Where a typed body exists (`CreateDidWebvhBody`, `UpdateContextBody`, …),
`to_v(typed)` is strictly stronger — it asserts **our types** conform, which is
what caught all three defects here. That's a second audit pass and will likely
surface more drift, so it deserves its own PR.

## BREAKING CHANGE

`WebvhPathMode` serializes as `wellKnown` / `autoAssign` instead of `well_known` /
`auto_assign`. Deserialization accepts both, so this breaks only a consumer that
string-matches on what we emit. Title carries `!`; expect the advisory semver job
to report it.

## Pre-merge

- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] `cargo test --workspace --no-fail-fast` — **4514 passed, 0 failed** across 145 binaries
- [x] No `version =` edits, no changelog file

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,9 +356,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-mediator"
-version = "0.18.18"
+version = "0.18.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "01857ae34f9931c8d0335d17a8f69917f22ae8e6fb3f09e1c7bfa56bb1343b62"
+checksum = "ab9d65f2641d3df7aef6687abd2bd0788d7f9e725755baed7460449303b42dee"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-common",
@@ -470,9 +470,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-messaging-sdk"
-version = "0.19.8"
+version = "0.19.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "954fe9bbc880aa008e9f01eeac117ab45419df00fddc40541c975fc70b3de506"
+checksum = "6cd4c9f5aaf47ef876e858a80e1a4c396a04f19a2301f2b0eb4ec4af45ea7180"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-authentication",
@@ -2945,7 +2945,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c54e03a951783e8b327515db3f2a2fd0e3bed362a96b066f341ce66ed49b4ead"
 dependencies = [
  "data-encoding",
- "syn 3.0.3",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -4610,7 +4610,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -6487,7 +6487,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls 0.23.43",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "thiserror 2.0.20",
  "tokio",
  "tracing",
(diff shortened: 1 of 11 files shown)


stormer78:
Closing — **superseded by #1015**, which landed the same work while this was in
flight. We were both bumping to `trust-tasks-rs` 0.11 and chasing the defects the
newly-resolvable schemas exposed, and #1015 got there first and covered more (four
defects to my three).

Independently reaching the same three fixes is a reasonable signal they were the
right calls:

- `CreateDidWebvhRequest` snake_case → camelCase
- `update_context` emitting `null` for unset optional members
- `WebvhPathMode`'s **discriminant value** `auto_assign` → `autoAssign`, with
  aliases kept on intake

Same for the census bookkeeping: manifest debt 36 → 14, 22 conformance witnesses,
`UNSPECCED_DISPATCHED_URIS` trimmed, producer-payload `UNPUBLISHED` trimmed.

Rebasing this onto main would be churn — 22 witnesses rewritten to say what
#1015's already say, and comments reworded to make the same points. Not worth the
review time.

Two small things here were **not** in #1015, and I've pulled them into a focused
follow-up rather than losing them:

1. `an_unspecced_task_proceeds_by_default` derives its unspecced URI from
   `UNSPECCED_DISPATCHED_URIS` at runtime instead of naming one. #1015 re-pinned it
   to `vta/seeds/rotate/1.0` and left a comment telling the next person to move it
   again when that gets specified — this just does that mechanically.
2. An annotation on the `api_integration` case that posts the retired snake_case
   spelling, marking it as deliberate back-compat coverage so nobody "modernises"
   away the only test of the aliases.

PR #1055, opened 2026-08-23
chore(deps)!: bring every dependency to latest, collapsing two duplicates

stormer78:
`cargo outdated` reported 13 direct dependencies behind and `cargo update` had 36 compatible updates waiting. Both are clear now — `cargo outdated` reports **"All dependencies are up to date"**.

Most of this is routine. Two items were not.

## p256 was duplicated in the build

Our crates declared `0.13` while `affinidi-crypto` — reached via `affinidi-data-integrity` and the TDK — already pulled `0.14`. The graph carried **two copies of a curve implementation**. The lock now holds a single `p256 0.14.0`.

The bump brings `elliptic-curve` 0.14 and `ecdsa` 0.17, which rename the SEC1 family: `EncodedPoint` → `Sec1Point`, `From/ToEncodedPoint` → `From/ToSec1Point`. Renamed across `vta-keys`, `vta-service`, `vti-webauthn`.

## tokio-tungstenite was load-bearing, not incidental

`vta-mobile-core` depends on it *solely as a feature enabler*: iOS has no native trust store, so `rustls-tls-webpki-roots` must be on graph-wide or the mediator WebSocket fails with `no native root CA certificates found`. Features unify per major version, so the declaration only works while it matches the version `affinidi-messaging-sdk` pulls — and the SDK moved to **0.30** in this refresh.

Updating everything *except* this one would have stranded the enabler and broken iOS `wss://` **silently**. The existing comment on that dependency is what flagged it.

## The rest

| Dep | Jump | Cost |
|---|---|---|
| `rcgen` | 0.13 → 0.14 (dev) | `signed_by` takes an `Issuer`; `self_signed` borrows. Test helper now uses `Issuer::from_params`, which states its intent more directly. |
| `syn` | 2 → 3 (dev) | No source change — syn 3 was already in the graph via `trust-tasks-rs`. |
| `rmcp` | 1.7 → 3.1.4 | Two majors, one rename: `Content` → `ContentBlock`. The `#[tool_router]`/`#[tool_handler]` surface the crate is built on is unchanged. |
| lockfile | 36 updates | Includes `trust-tasks-rs` 0.11.3 (the corrected `vta/app-state` error taxonomy from dtgwg-trust-tasks-tf#253) and the AWS SDK set. **`rustls-pemfile` drops out of the graph entirely** — one of the unmaintained crates `cargo audit` flags. |

## Two places the shortest path was worse

**`vti-webauthn` keeps parse-then-validate** rather than collapsing to the one-shot `PublicKey::from_sec1_bytes`. That call merges "malformed SEC1 encoding" and "valid encoding, point not on the curve" into a single error, and those say different things to whoever reads the log — a broken client versus a point somebody chose.

**BIP-32 P-256 derivation** replaces the now-deprecated `FieldBytes::from_slice` with `TryFrom` and a real error arm. The input is a fixed 32-byte window of a SHA-512 HMAC so it cannot fail, but the length check is explicit instead of resting on a panic inside a deprecated helper. Worth noting because CI runs clippy with `-D warnings`, so the deprecation would have failed the build.

## Unchanged

The four `cargo audit` advisories ignored in `deny.toml` — all transitive through the AWS SDK's hyper 0.14 / rustls 0.21 path. `cargo deny check advisories` passes.

Separately worth someone's attention (not touched here): `cargo deny` reports the `RUSTSEC-2026-0258` ignore as **stale** — "no crate matched advisory criteria". If the AWS SDK has moved off that path, the entry can be dropped, which is the condition `deny.toml` itself names.

## Verification

- `cargo test --workspace` — **149 suites, 0 failed** (same count as the pre-refresh baseline)
- `cargo clippy --workspace --all-targets` — clean
- `cargo outdated --workspace --root-deps-only` — all up to date

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,21 +93,21 @@ dependencies = [
 
 [[package]]
 name = "affinidi-crypto"
-version = "0.2.7"
+version = "0.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00fbf84b8b1e8b813fdbb4572e8f688ab47147d8ea21738d256d39afb0ab7810"
+checksum = "3c5420ba310200f4caaa72efb95898a1202636ee40c3baeb6555be88a7dde423"
 dependencies = [
  "aes 0.8.4",
  "affinidi-encoding",
  "base58",
- "base64 0.22.1",
+ "base64 0.23.1",
  "cbc 0.1.2",
  "ed25519-dalek",
  "getrandom 0.2.17",
  "hmac 0.12.1",
  "k256",
  "multibase",
- "p256 0.14.0",
+ "p256",
  "p384",
  "p521",
  "rand 0.10.2",
@@ -149,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "affinidi-did-authentication"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6baa61c2753f158f246e9b5b5c15d2ec989c8694e7ab550bee817db8d776ade8"
+checksum = "9fd06555e5116cca844aa1d2f2e42cc75bd3bbf85069b49e3f466c8ba245e317"
 dependencies = [
  "affinidi-crypto",
  "affinidi-did-common",
@@ -159,7 +159,7 @@ dependencies = [
  "affinidi-encoding",
  "affinidi-messaging-didcomm",
  "affinidi-secrets-resolver",
- "base64 0.22.1",
+ "base64 0.23.1",
  "chrono",
  "reqwest",
  "serde",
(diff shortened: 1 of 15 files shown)


stormer78:
**Re-marked as breaking (`!`) after checking the semver result properly.**

`semver checks` is red, but its output on this branch is *identical to the pre-refresh baseline* — every field it names (`audit_sink`, `audit`, `audit_ks`) is the #1049 rename already on main, and the job is `continue-on-error: true`. Nothing p256-shaped appears.

That is not an all-clear, and the reason is worth stating: **`cargo-semver-checks` cannot see this one.** It compares rustdoc JSON, and `p256::SecretKey` renders as the same path string in 0.13 and 0.14, so a field whose type moved between incompatible versions of a third-party crate looks unchanged.

The break is real:

```rust
// vta-keys/src/derivation.rs — re-exported as vta_service::keys::…
pub struct P256Secret {
    pub secret_key: p256::SecretKey,   // 0.13 -> 0.14
}
```

A consumer holding its own p256 0.13 and passing values across that boundary stops compiling. `vti-webauthn` uses p256 privately only, and `vta-sdk` does not depend on it, so this field is the whole exposed surface.

Marked `!` because this workspace's rule — stated in the `semver-checks` job itself — is that a break is allowed but must be **known**, so the release moves the compatibility field rather than shipping it as a patch. Left as plain `chore(deps)`, release-plz would have released it as one.

The marker is hand-applied, not tool-derived. Flagging that explicitly so nobody later reads a green-ish semver job as evidence there was no break.

PR #1060, opened 2026-08-23
feat(vtc): tell a member when the community removes them

### LINKED ISSUE #1053: No removal notice reaches a member the VTC removes
Surfaced while scoping the client half of OpenVTC/openvtc#240. The client has a
`Removed` membership state with no way to ever enter it, because the VTC never
tells a member they were removed.

## Problem

There is no outbound message on any removal path. `AppState::send_to_member`
(`vtc-service/src/server.rs:241`) is the single outbound funnel, and its only
caller is `vtc-service/src/credentials/delivery.rs:176`. Every delivery call site
is a credential push on an *approve* / *present* / *renew* path:

- `vtc-service/src/messaging.rs:1339`
- `vtc-service/src/join/orchestrate.rs:303`
- `vtc-service/src/ceremony/orchestrate.rs:112`
- `vtc-service/src/routes/join_requests/present.rs:269`
- `vtc-service/src/routes/join_requests/decide.rs:161`
- `vtc-service/src/routes/members/request_vmc.rs:89`

None is on a removal path.

What exists and is **not** a removal notice:

- `spec/vtc/members/removed/0.1` — an *admin query* listing tombstoned members
  (`vtc-service/src/routes/mod.rs:568` → `routes/members/read.rs:198-220`).
- `MEMBER_SELF_REMOVE_TYPE` / `MEMBER_SELF_REMOVE_RECEIPT_TYPE`
  (`vta-sdk/src/protocols/join_requests.rs:324,330`) — the member-initiated
  leave and its receipt, i.e. a reply to the member's own request.
- `spec/vtc/members/admin-remove/0.1` — admin-facing. `admin_remove`
  (`vtc-service/src/routes/members/remove.rs:127-149`) accepts an optional
  `reason` (capped 1024, `remove.rs:69`) and returns `RemoveResponse` **to the
  admin**. The reason never leaves the audit log.

The only thing a removed member can observe is a side effect: the revocation bit
on their VMC flips (`vtc-service/src/ceremony/execute.rs:468-492`, `flip_revocation`
at `:497`). They must discover their own removal by polling a status list, and
even then learn nothing about why.

## Why it matters

Removal is the most consequential thing a community can do to a member, and it
is the one decision delivered with the least information — currently zero. The
operator sees their access stop working. Any client-side "you were removed, here
is why" UI is unreachable until this exists: in OpenVTC,
`CommunityRecord::remove()` has no production caller at all, precisely because
nothing on the wire triggers it.

## Proposed change

Send the removed member a notice on both admin removal paths, carrying:

- a typed disposition/code,
- the operator `reason` already accepted at `remove.rs:69`,
- the decision timestamp,
- the deciding authority.

On that last point — `reject_pending` knows `admin.0.did` (`decide.rs:242`) and
writes it to audit as the actor, but it never reaches any wire response. Same
shape here.

## Related

The response document is unsigned (`proof: None` in
`trust-tasks-rs/src/document.rs:387-407`), and `issuer` is echoed from the
request's `recipient` field, which the *applicant* set. So a removal notice as
things stand would not be independently verifiable — a removed member could not
prove to a third party who removed them, when, or on what grounds. Worth deciding
whether a removal notice should carry a proof; it is the case where forwardable
evidence matters most.

## Acceptance

- Admin removal and purge each emit a notice to the affected member.
- The notice carries code, reason, decision timestamp and deciding authority.
- A member with no live socket receives it on reconnect (or can recover it from
  a poll).

stormer78:
Closes #1053. Spec published upstream first as trustoverip/dtgwg-trust-tasks-tf#256 (`vtc/members/removal-notice/0.1`, in `trust-tasks-rs` 0.11.4) — the VTC's URI census requires every `trusttasks.org/spec/…` literal in its source to resolve in the published registry, so the constant could not be declared before the spec existed.

## The gap

Removal was the most consequential thing a community does to a member and the one it delivered with the least information — **none**. No outbound message existed on any removal path. The only observable signal was a side effect: the revocation bit on the member's credential flipping. They inferred their own removal from a status list and learned nothing about why.

## Signed, because the recipient is not the audience

Authcrypt already proves the sender **to the member**. But this is the one member-facing message whose value lies in showing it to somebody else — an appeal, a dispute, another community weighing a rejected applicant. Forwarded, an unsigned notice is an assertion anyone could have written.

So it is a Trust Task document with a Data Integrity proof, packed in the trust-task envelope. Note `TRUST_TASK_ENVELOPE_TYPE`, **not** the task URI — hand-building a DIDComm message with the task type is a mistake this workspace has made before, and a conformant peer rejects it silently.

## Thirty days, because there is no second route

Verified while scoping: `resolve_auth_role` returns `Forbidden("DID not in ACL")` and removal hard-deletes the ACL row, so **a removed member cannot authenticate at all**. Every authenticated route — including any poll that might have served the notice — closes the moment the removal lands. The act this reports is the act that ends their ability to ask about it.

That makes the push the only channel, which is what justifies the window. `send_to_member` already had `Delivery::Guaranteed` with a durable outbox, so this needed a deadline parameter rather than new machinery: `send_to_member_by`, with the six existing callers keeping the 24h default.

**A member offline beyond the window still never learns.** No window fixes that. The honest fix is a retrieval path not gated on an ACL row — a larger design than this issue. Stated in the spec and in the constant's doc comment rather than left to be discovered.

This also means one of the issue's acceptance criteria — *"or can recover it from a poll"* — is **not achievable as written**, and I have said so on the issue rather than satisfying the easier half and calling it done.

## Never for a self-leave

`remove_inner` serves both the admin path and the DIDComm self-leave. A member who chose to leave already has their receipt, and telling them they were "removed" is a different and worse thing to be told. Guarded on `actor_did != target_did`, and tested by asserting an absence.

## Best-effort

The notice never fails the removal. That has already happened and is durable; refusing the operator's request because the notice could not be *queued* would leave the member removed and the operator believing they were not. Failures log loudly — and log **"queued", not "sent"**, since a DIDComm `Ok` means the mediator accepted the frame and nothing more (R1.1, which names `send_to_member` as a known offender).

## Tests

**Four unit** — payload shape; a blank reason omitted rather than sent as an empty string (the route defaults `reason` to `""`, so without this the common case would ship a meaningless empty value); the two codes distinguishable; and the payload **validated against the published schema**, which is the check that catches the implementation drifting from the spec it claims to implement.

**Three end-to-end over a real mediator** (`--features didcomm-harness`), because a `send` returning `Ok` proves only that the mediator accepted the frame — the only thing that demonstrates this works is a peer holding the document:

- an admin removal reaches the member, signed, with reason and deciding admin
- a purge says it was a purge, with `reason` absent rather than empty
- a self-leave sends nothing

Two of those first failed with `no active removal policy` — informative rather than annoying. The purge test passed before them precisely because **a purge deliberately skips the removal policy**, so the failure confirmed the paths differ the way the spec says.

## Verification

- `cargo test --workspace` — **150 suites, 0 failed** (149 before; the new harness test)
- `cargo test -p vtc-service --features didcomm-harness --test removal_notice_didcomm` — 3 passed
- `cargo clippy` (including `--features didcomm-harness`) — clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2952,7 +2952,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c54e03a951783e8b327515db3f2a2fd0e3bed362a96b066f341ce66ed49b4ead"
 dependencies = [
  "data-encoding",
- "syn 3.0.3",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -4588,7 +4588,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -6425,7 +6425,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls 0.23.43",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "thiserror 2.0.20",
  "tokio",
  "tracing",
@@ -6464,7 +6464,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.5",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.61.2",
 ]
@@ -8680,9 +8680,9 @@ dependencies = [
 
 [[package]]
 name = "trust-tasks-rs"
-version = "0.11.3"
+version = "0.11.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ea804cc896dae689fa5cecc8edcfe449d8355424b86eb862fcf1f71a240dcb53"
+checksum = "fb3630f56ee78546c9ccfbb4aed780aa96fb469d758ddf832c9060d3e76c1b8b"
 dependencies = [
  "async-trait",
  "chrono",
--- a/vta-sdk/src/protocols/members.rs
+++ b/vta-sdk/src/protocols/members.rs
@@ -46,6 +46,74 @@ pub const MEMBER_VMC_TYPE: &str = "https://trusttasks.org/spec/vtc/members/vmc/0
 pub const MEMBER_VMC_RESPONSE_TYPE: &str =
     "https://trusttasks.org/spec/vtc/members/vmc/0.1#response";
 
+/// VTC → member: **unsolicited** notice that the community removed them.
+///
+/// Distinct from the self-remove receipt
+/// ([`crate::protocols::join_requests::MEMBER_SELF_REMOVE_RECEIPT_TYPE`]) in the
+/// way that matters: a receipt answers a request the member made and is
+/// correlated to it, so the member is already waiting for it. This answers
+/// nothing — the member did not ask, is not waiting, and may be offline. Sending
+/// one for a member-initiated departure would tell somebody who chose to leave
+/// that they were removed.
+pub const MEMBER_REMOVAL_NOTICE_TYPE: &str =
+    "https://trusttasks.org/spec/vtc/members/removal-notice/0.1";
+
+/// Which removal happened, as carried in [`RemovalNoticeBody::code`].
+///
+/// Two variants rather than one `removed` flag because they differ in what
+/// recourse the member has: an administrator's removal ran the community's
+/// removal policy, a super-administrator's purge deliberately skipped it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum RemovalCode {
+    /// Policy-governed removal by an administrator.
+    AdminRemoved,
+    /// Forceful super-administrator deletion; skips the removal policy.
+    Purged,
+}
+
+/// Body of a [`MEMBER_REMOVAL_NOTICE_TYPE`] notice.
+///
+/// Every field except `reason` is required, because a notice that omits any of
+/// them fails to answer one of the questions a removed member has: what
+/// happened, to what, when, and on whose say-so.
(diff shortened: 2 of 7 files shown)


stormer78:
Rebased onto #1058, which landed while this was open. All seven tests pass unchanged; workspace suite 150 / 0 failed.

## On the `semver checks` result — a correction worth recording

The first run reported a `vta-sdk` break I could not reproduce locally:

```
constructible_struct_adds_field
  field JoinRequestStatusResponseBody.code
  field JoinRequestStatusResponseBody.reason
  field JoinRequestStatusResponseBody.decided_at
```

Local `cargo semver-checks -p vta-sdk` said *"no semver update required"*. Both were correct — **they were testing different trees.** CI checks the PR merged with `main`; my branch predated #1058, which added those three fields. The break is #1058's (already marked `!` there), not this PR's.

My additions to `vta-sdk` are a new const, a new enum and a new struct — nothing existing altered, so additive. Re-verified against the current base after the rebase.

Flagging it because "it's the usual #1049 noise" was very nearly the conclusion, and earlier in this sequence that assumption would have let a real break through (#1055, where the tool genuinely could not see a p256 field-type change).

## How this relates to #1058

The two solve the same complaint — *"we decided something about you and never told you"* — and solve it differently, for a reason worth stating so they do not read as inconsistent:

| | #1058 (rejected applicant) | This PR (removed member) |
|---|---|---|
| Fix | extend the **poll** | **push** a notice |
| Why | an applicant can still poll | a removed member **cannot authenticate at all** |

`resolve_auth_role` refuses any DID without an ACL row and removal deletes it, so the recovery path #1058 relies on is destroyed by the very act this reports. That is what forces a push and the long delivery window.

Field naming is deliberately aligned — `code`, `reason`, `decided_at` — so a client reading both surfaces sees one vocabulary rather than two.

Issue #1054, opened 2026-08-23
Pairwise relationship identifiers vs. the relationships ACL policy


stormer78:
Raised from OpenVTC/openvtc#241, which proposes defaulting relationship
communication to pairwise Relationship DIDs (R-DIDs) rather than the persona
DID. That issue originates in an external RAHP/DRARM assessment finding
(`OVTC-RAHP-04`) scoring persona-DID relationship communication as stable-
identifier correlation.

The client-side half of that change is straightforward. But it collides with a
VTC policy, and the protocol question belongs here.

## The collision

`vtc-service/policies/default/relationships.rego` is the whole policy:

```rego
default allow := false
allow if {
	input.action == "publish"
	input.issuer_member.is_current
	input.subject_member.is_current
}
```

Both parties must be current, ACL-listed, non-tombstoned community members
(`vtc-service/src/routes/relationships.rs:122-134`, `:349-356`). And the issuer
is pinned to the authenticated session DID:

`vtc-service/src/routes/relationships.rs:97-107`
```rust
let issuer_did = extract_did_field(vrc, "issuer")?;
let subject_did = extract_subject_id(vrc)?;
if auth.did != issuer_did {
    return Err(AppError::Forbidden(format!(
        "session DID ({}) does not match VRC issuer ({issuer_did}) — \
         VRCs are self-issued; the VTC never mints them on a member's behalf", auth.did)));
}
```

A pairwise R-DID has no ACL row, so a VRC issued under one is rejected — 422
"subject DID … is not a current community member", or 403
`RelationshipPolicyDenied`. **The only enforcement that exists on this path
forbids pairwise identifiers rather than requiring them.**

`GET /v1/relationships/graph` (`routes/relationships.rs:446`) then deliberately
exposes the full correlatable member-to-member edge set to admins.

This is not a bug report — the ACL check is doing what it was designed to do.
It is a design question that the client cannot resolve on its own.

## The question

Should relationship communication use pairwise identifiers, and if so, how does
that reconcile with a trust graph whose entries are required to be attributable
to current members?

Some shapes worth weighing:

1. **Separate the two.** Pairwise identifiers govern the *communication channel*;
   the published VRC stays persona-attributed. This is roughly what OpenVTC does
   today — the R-DID carries post-handshake messages, the VRC `issuer` is the
   persona DID. It bounds the exposure to the graph, but the graph is exactly
   where the correlation the assessment describes lives.
[...]

PR #1061, opened 2026-08-23
feat(vtc)!: let a member publish a relationship edge without naming themselves in it

stormer78:
Closes the protocol half of #1054. Design note:
`docs/05-design-notes/vrc-publish-proof-of-possession.md`.

## The problem

`POST /v1/relationships` required the VRC's `issuer` to equal the caller's
session DID:

```rust
if auth.did != issuer_did { return Err(AppError::Forbidden(...)) }
```

That is why a member's membership DID ends up inside the durable, publishable
credential, and why `GET /v1/relationships/graph` is a correlatable
member-to-member edge set. DTG Credentials asks for the opposite — R-DIDs
RECOMMENDED, a new unique R-DID per counterparty "even within the same
community", and M-DID edges as a bootstrapping allowance to migrate away from.

## The change

The pin conflated two properties:

- **The VRC was made by the party it names as issuer.** Provided by the
  data-integrity proof. Untouched here.
- **The publisher is the issuer.** What the pin actually provided, and worth
  keeping — issuance and publication are different disclosures, and appearing
  in the community graph should be the issuer's to make rather than that of
  whoever holds a copy of the credential.

So the pin is *replaced*, not removed. The session proves membership; a publish
authorization signed by the issuing key proves control of it. Neither requires
them to be the same string.

```json
{
  "type": "VrcPublishAuthorization",
  "vrc": "<sha-256 of the VRC>",
  "aud": "<this VTC's DID>",
  "sessionId": "<caller's session>",
  "issuedAt": "2026-08-23T07:40:00Z"
}
```

Signed with `eddsa-jcs-2022`, verified through the existing `DidVmResolver` +
`check_issuer_binding` path. Every field covers a distinct replay — signatures
made over other objects, other credentials, other communities, other members'
sessions, unbounded reuse within a live session — and there is a test per field.

**The authorization is verified and dropped.** It carries `sessionId`, which is
attributable to a membership DID; persisting it would rebuild exactly the
linkage pairwise identifiers exist to remove. This is the point the ZKP task
force made on trustoverip/dtgwg-cred-spec#9 — "adding a visible field to make
the association checkable would recreate exactly the correlation surface you are
flagging." Two tests assert it directly against the stored row and the audit
store, because it is the kind of property a later debug log undoes in silence.

## Subject membership check dropped on the pairwise path

Not merely unanswerable once the subject is an R-DID — DTG Credentials
§Community-Anchored ZKP is explicit that "community membership is **not** a
precondition for issuing, holding, or presenting a VRC". The subject's consent
to the edge is their publication of the reciprocal VRC (the two-VRC edge model),
not this community's assertion that they exist. This is the same reasoning the
spec editors gave on trustoverip/dtgwg-cred-spec#8 for the member-issued
reciprocal VMC being the member's consent artifact — a pattern this codebase
already implements at join.

## Breaking: policy input shape

`relationships.rego` input gains `authenticated_member`, `issuer.pop_verified`
and `subject`. `issuer_member` / `subject_member` are still supplied; for a
pairwise VRC both report `is_current: false`, so **an operator policy written
against the old shape denies the publish** rather than being silently loosened.
That is deliberate — this change should not quietly relax a policy someone
wrote. The shipped default allows both forms.

The deprecated form (`issuer` equal to the session DID, no authorization) is
accepted for one release and logs a deprecation line, so existing clients keep
working while they move. OpenVTC/openvtc#254 already defaults the client to
pairwise R-DIDs.

## Tests

11 new integration tests in `vtc-service/tests/relationships.rs::pairwise`,
against a live `did:key` resolver:

- publishes under a relationship DID; asserts the stored row contains neither
  the member's DID nor the session id
- the authorization never reaches the audit trail
- subject need not be a member
- idempotent re-publish
- rejects: missing authorization, one signed by another key, one bound to a
  different VRC, one from another session, one for another community, a stale
  one, and one of the wrong `type`

Plus a policy unit test for the pairwise rule. `769` lib tests unchanged.

One existing test earned its keep: an early draft moved the resolver ahead of
caller validation, so a missing resolver returned 500 for what was really a 403.
`publish_rejects_caller_not_issuer` caught it, and the ordering comment in the
handler now says why the gate stays where it is.

## Not in scope, flagged

- **Audit attribution.** `audit_writer.write(&issuer_did, ...)` now records the
  R-DID, so publications are no longer attributable to a named member. That is
  the intended privacy property and a real loss for moderation. This PR takes
  the privacy-preserving option; the design note lists the alternatives. **Worth
  an explicit decision in review rather than acceptance by default.**
- **`GET /relationships/graph`** still stores and returns an adjacency list —
  now of R-DIDs, so no longer member-correlatable, but the design note argues it
  should become derived rather than stored, and should distinguish half-edges
  from complete edges.
- **`canonicalise`** is a recursive key sort, not RFC 8785. Unchanged here, but
  it now has a second consumer via the `vrc` hash binding. Same class of problem
  as trustoverip/dtgwg-cred-spec#6.
- **Accepted DID methods** should be community policy rather than whatever
  `DIDCacheConfigBuilder::default()` supports. `vta-config` already has the shape
  to copy (`allowed_did_methods`); the VTC's `validate_did` hook is still the
  no-op default.
- **The only prose spec for this endpoint** —
  `trust-tasks/relationships/publish/1.0/spec.md` — is marked `status: retired`
  and superseded by a `0.1` document that does not exist in this repo. I have
  not edited a retired spec; the design note is the current reference. Worth
  resolving separately.

## Still open upstream

The residual transient linkage — the VTC observes the session DID and the R-DID
in the same request, unstored — needs the community-anchored ZKP to remove.
That needs the identity linkages of trustoverip/dtgwg-cred-spec#9 defined first;
the ZKP TF's position there is "a real dependency, not a blocker". This design
is the shape that takes that proof when it lands: `issuer.pop_verified` is
replaced by a ZK proof of the same predicate, and neither the policy input nor
the stored row changes.

trustoverip/dtgwg-cred-spec#21 tracks the glossary/body contradiction that made
this ambiguous in the first place.

### DIFF
--- a/docs/05-design-notes/vrc-publish-proof-of-possession.md
+++ b/docs/05-design-notes/vrc-publish-proof-of-possession.md
@@ -1,0 +1,267 @@
+# VRC publish: proof-of-possession in place of the issuer pin
+
+Status: implemented — see `vtc-service/src/routes/relationships.rs`
+Tracking: OpenVTC/verifiable-trust-infrastructure#1054, OpenVTC/openvtc#241
+Upstream: trustoverip/dtgwg-cred-spec#21, trustoverip/dtgwg-cred-spec#9
+
+## Why
+
+`POST /v1/relationships` pins the VRC's `issuer` to the authenticated session
+DID (`vtc-service/src/routes/relationships.rs:97-107`):
+
+```rust
+if auth.did != issuer_did { return Err(AppError::Forbidden(...)) }
+```
+
+That equality is what forces a member's membership DID (M-DID) into the durable,
+publishable credential. Removing it is the change #1054 is blocking on. But the
+pin is doing real work, and dropping it without a replacement removes a property
+we want to keep.
+
+## What the pin actually protects
+
+Two properties are conflated in one line, and they need separating before
+anything is replaced.
+
+**P1 — the VRC was made by the party it names as issuer.** Provided already by
+`verify_vc_proof` + `check_issuer_binding` (`relationships.rs:316-343`): the
+data-integrity proof is verified against a verification method bound to the
+`issuer` field. This holds whatever identifier the issuer used and is untouched
+by anything below.
+
+**P2 — the party publishing the VRC is the party that issued it.** This is what
+the pin provides, and it is the property that needs a replacement.
+
+P2 is worth keeping. Issuance and publication are distinct acts. A VRC handed
+privately to a counterparty should not become an edge in the community graph
+because the *counterparty* chose to upload it. Appearing in the graph is a
+disclosure, and it should be the issuer's disclosure to make. Without P2 any
+authenticated member could publish any VRC that ever reached them.
+
+## Design
+
+Replace the identity equality with a **publish-time proof of possession** of the
+issuer's key, bound to the request. The session proves membership; the PoP proves
+control of the issuing identifier; neither requires the two to be the same
+string, and neither requires disclosing that they belong to one member.
+
+### The authorization object
+
+The client signs, with the private key of the VRC's `issuer`:
+
+```json
+{
+  "type": "VrcPublishAuthorization",
+  "vrc": "<sha-256 of the VRC, hex>",
+  "aud": "<this VTC's C-DID>",
+  "sessionId": "<the caller's authenticated session id>",
+  "issuedAt": "2026-08-23T07:40:00Z"
+}
+```
+
+Each field earns its place:
+
+| Field | Prevents |
+|---|---|
+| `vrc` | replaying a captured PoP to authorize a *different* credential |
+| `aud` | replaying a PoP made for one community at another |
+| `sessionId` | replaying another member's PoP — this is the load-bearing binding |
+| `issuedAt` | unbounded replay within a live session (accept a narrow window) |
+
+`sessionId` is the field that does the real work. It ties "controls this R-DID"
+to "is making this request" *inside the request*, rather than by naming the
+member inside the credential. `AuthClaims.session_id`
+(`vti-common/src/auth/extractor.rs:30`) is already carried through to handlers
+for exactly this class of session-targeted operation, so nothing new is needed
+to reach it.
+
+The PoP is carried as a second field on `PublishBody`, which today is a single
+`vrc` (`relationships.rs:57-65`).
+
+### Canonicalization
+
+The authorization object is signed with an `eddsa-jcs-2022` data-integrity
+proof, so the signature is over RFC 8785 (JCS) canonical form and is reproducible
+by any conforming implementation. This is the same `DataIntegrityProof` path the
+VRC itself and the credential-exchange verifiers use, so no new machinery.
+
+The `vrc` field of the authorization binds to the hash the handler computes with
+its own `canonicalise` (`relationships.rs`), which is a recursive key sort only —
+no number normalization, no string-escaping rules. That is adequate for the
+idempotency key it was written for, and this change does not alter it, but it is
+not a canonicalization a second implementation could reproduce from a
+specification. It is a latent interop problem of the same kind
+trustoverip/dtgwg-cred-spec#6 was filed about, and it now has a second consumer.
+Worth a follow-up; deliberately out of scope here to keep the change reviewable.
+
+### Verification order
+
+1. Authenticate the session (unchanged). `auth.did` is a current member.
+2. Extract `issuer` and `credentialSubject.id` from the VRC (unchanged).
+3. **New, and it must stay here:** if no PoP was supplied and `issuer` is not the
+   session DID, reject. This gate is before the resolver is touched, because the
+   resolver is a *daemon-configuration* prerequisite — putting it first lets a
+   missing resolver return 500 for what is really a 403, masking a caller error
+   with an operator error. The original code ordered it this way deliberately
+   ("Daemon-config prerequisites surface after caller validation"); an early
+   draft of this change lost that, and the existing
+   `publish_rejects_caller_not_issuer` test caught it.
+4. Verify the VRC's data-integrity proof (unchanged) — **P1**.
+5. Compute the VRC hash. Moves ahead of policy evaluation, since the PoP binds
+   to it.
+6. Verify the PoP signature against the `issuer` DID's verification method,
+   reusing `DidVmResolver` and the same `check_issuer_binding` rule as step 4.
+   Then check `type`, `vrc`, `aud`, `sessionId`, and the `issuedAt` freshness
+   window — **P2**.
+7. Evaluate policy against the new input shape (below).
+8. Store, audit, respond (unchanged).
+
+The `type` guard on the authorization object is not decoration: without it, any
+object the member legitimately signed with the same key and which happens to
+carry the right field names could be replayed as authorization to publish.
+
+### The PoP is verified and discarded
+
+**It must not be persisted, logged, or written to the audit trail.** The PoP
+contains `sessionId`, and the session is attributable to an M-DID; storing it
+would create exactly the durable M-DID-to-R-DID linkage this change exists to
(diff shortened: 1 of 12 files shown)


stormer78:
Pushed a second commit (11e10916) after a conformance pass against DTG
Credentials, prompted by a simple correction: **there is no such thing as a
recognition credential — only a relationship credential.**

`VerifiableRecognitionCredential` is not one of the seven concrete subtypes the
specification defines, and never was. It survived in this repo because the
publish path **never inspected `type` at all** — the string lived only in
fixtures, doc comments and an architecture table, describing a credential
nothing produces.

Everything else in the stack was already right:

- `dtg-credentials` has `DTGCredential::new_vrc`, emitting the correct
  `["VerifiableCredential", "DTGCredential", "RelationshipCredential"]`.
- openvtc has always minted VRCs through it.
- The VTC accepted those only because it accepted anything.

Which means **the 11 tests in the first commit were exercising a wire form no
client sends.** That is a fair hit on this PR: the tests would have passed even
if the handler rejected every real VRC. They now use the actual wire form, and
the gap is closed rather than documented — `check_vrc_shape` requires both DTG
`@context` entries, the `DTGCredential` base type, and exactly the
`RelationshipCredential` subtype, classified through
`dtg_credentials::DTGCredentialType` so issuer and verifier share one definition
instead of string-matching independently.

Four new tests cover it, including a regression guard that names the invented
type directly so it cannot drift back in.

The recognition *protocol* — cross-community session minting from a foreign
VEC + VMC pair — is a genuinely different concept and is untouched.

Full suite green: 53 suites, clippy clean.

### Related, not fixed here

The same sweep turned up #1062: `/v1/auth/recognise` routes by the literal
`"VerifiableEndorsementCredential"`, which no catalog-minted VEC carries, and
its test hand-rolls the same fictional type — so the suite agrees with the bug.
Different subsystem, so it is filed rather than folded in.

### Still open on this PR

- **R-DID uniqueness is unenforced.** The spec requires a new unique R-DID per
  counterparty, "even within the same community". This PR makes pairwise edges
  possible without making them pairwise: a client reusing one R-DID everywhere
  gets back the correlation we removed, and the VTC stores it happily. The VTC
  is the only party that can see this, and `list_for_did` already makes it one
  lookup. I would add it before merge — as policy with deny as the shipped
  default, since bootstrapping communities may need a grace period.
- **Audit attribution** still needs the explicit decision flagged above.
- **Validity windows** are never checked on publish; an expired VRC becomes a
  permanent edge. Informative-section requirement, so lower priority, but real.

stormer78:
Third commit (b8bbb40f) implements **F5**, and corrects a framing error the
Linux-community case exposed in the second one.

## Both forms are permanent

The previous commit marked the membership-DID path *deprecated* and logged a
deprecation warning on every publish. That was wrong. Take a public open-source
community: everyone is known, the graph is attributed on purpose, and that is
the point of it rather than a compromise. DTG Credentials permits this
directly —

> The member may also assert the M-DID in any VRC where the member wishes to
> assert a VTC relationship.

— so the warning was telling a perfectly conforming community that its normal
operation was technical debt. Removed. The two forms are now named and
symmetrical:

- **attributed** — issued under the member's membership DID; the edge names
  them and the graph is correlatable by design.
- **pairwise** — issued under a relationship DID scoped to one counterparty,
  with the publish authorization proving control of it.

## Per-member choice, community-declared default

The member chooses per relationship. The community *declares* what it expects
in the profile as `relationshipIdentifierDefault` — a published fact a client
reads **before** minting, which is what makes it a declaration rather than a
preference buried in policy. Defaults to `pairwise`, matching the
specification's recommendation, so a community that has not considered the
question declares the more conservative expectation.

It is not a gate: both forms are accepted. A community that wants to *require*
one enforces it in `relationships.rego`, which now receives `identifier_form`
on every publish — the rego comments carry the three-line rule for doing so.

## Uniqueness is type integrity, not privacy policy

The spec MUST — "a new, unique R-DID for every single entity they connect with,
even within the same community" — is now enforced on publish, and
unconditionally, not as something a community can switch off. The reasoning
matters for review:

A verifier reading a pairwise edge is entitled to conclude the identifier says
nothing beyond that one relationship. A reused R-DID breaks that inference for
**every reader of the graph**, not just for whoever reused it. And the community
is the only party who can see the violation at all — each counterparty sees only
its own edge. So it is "a UUID must be unique", not "you must be anonymous".

Nothing is lost by holding the type to its meaning, because a member who wants
to be recognised across their relationships has the attributed form. The rule
does not apply there: a membership DID is not claiming to be pairwise.

Three tests, including `attributed_edges_may_share_one_identifier_across_counterparties`
— the Linux case, asserting that an attributed edge is *never* rejected by the
uniqueness rule.

## One real break the tests caught

Adding a required field to `CommunityProfile` made every config export taken
before it unreadable on import. `admin_config`'s round-trip tests failed on it.
Fixed in the code with `serde(default)` rather than in the tests, with a comment
saying why it is load-bearing.

Full suite green: 53 suites, 0 failures, clippy clean.

## Raised upstream

trustoverip/dtgwg-cred-spec#21 now carries the implementation report: the
bootstrapping-vs-terminal-state framing problem, the fact that the uniqueness
MUST has no assigned enforcer while the only party who can see violations is
not the party bound by it, and the general observation that catalog-mediated
issuance plus hand-rolled test credentials produces exactly the ingress blind
spot this PR has been closing.

## Still open

- **Audit attribution** — publications attribute to the R-DID on the pairwise
  path, so they are no longer traceable to a named member. Intended privacy
  property, real moderation loss, and still your call.
- The wider conformance audit turned up nine further findings; #1062 is filed,
  the rest are held pending a read.

stormer78:
Fourth commit (c03a845d): **audit attribution decided — option B**. The last open item on this PR is closed.

The trail attributes a publication to the **authenticated member**, not to the VRC issuer. Under the pairwise form those differ, and the issuing relationship DID names nobody — so the trail could not answer "which member published this edge" for anyone, at any access level. Not redacted; never captured.

Checking the writer changed my read of the trade, and is worth recording. `AuditEnvelope` stores **two** actor fields:

- `actor_did_hash` — HMAC-SHA256 under a rotating key, covered by the tamper-evidence chain
- `actor_did_plain` — plaintext, deliberately **excluded** from the chain digest so RTBF can null it without breaking verification

So the store already separates correlation handle from plaintext identity, and already has redaction machinery. This is not "persist a permanent plaintext linkage".

Three reasons for drawing the exception here and nowhere else:

1. **The store is the right one.** What #1054 removes is *public, permanent, unavoidable* correlation — a membership DID welded into a credential anyone can retain and republish. The audit store is private, admin-gated, tamper-evident and redactable. That is what a privileged accountability record is for.
2. **Refusing it buys little.** `VrcPublishedData` carries `vrc_id`, `vrc_id` resolves to the row, and the row holds the relationship DIDs. Any trail that references the edge *and* names the member creates the mapping transitively — there is no variant that attributes without linking.
3. **Moderation needs the reverse lookup.** "This edge is abusive, who made it?" is the question operators actually have. Attributing to a hash alone answers only "this member is suspect, what have they done?".

**The residual, stated rather than buried:** an operator with audit access can map every pairwise edge to its member. That is the cost and it is accepted. It is also why the linkage stops there — the `info!` on the publish path still carries the relationship DID and not the member, because logs have neither the redaction machinery nor the access controls that make the trade defensible.

A test asserts the actor is the member and never the relationship DID, and the design note records the reasoning rather than just the outcome.

Worth flagging for review: **this decision is not VRC-specific.** It answers "when a member acts under a pairwise identifier, what does the audit trail record?" and should hold for every pairwise-capable operation added later. If you disagree with it, disagree now — it is cheaper than unpicking it from three subsystems.

Full suite green: 53 suites, 0 failures, clippy clean. All four commits DCO-signed.

This PR now has nothing open on it from my side.

Issue #1059, opened 2026-08-23
No conformance witness for VTC Trust Task responses, and two already diverge


stormer78:
Surfaced while fixing #1052. The VTC binds 61 published `spec/vtc/*` Trust Task
URIs and **nothing checks that any of them speaks its URI's wire shape.** Two
of them provably do not.

## What is checked, and what is not

`vtc-service/tests/trust_task_manifest.rs` runs two parity checks, and both are
about *identity*, not *shape*:

- `every_bound_task_is_published_or_excepted` — coverage
- `every_bound_canonical_task_exists_in_the_registry` — the URI resolves in
  `trust_tasks_rs::schema_index`

Neither looks at a payload. There is no runtime validation either — a grep for
`validate_payload` / `schema_for` across `vtc-service/src` returns nothing but a
single unrelated comment. So a VTC response can diverge arbitrarily from the
schema its URI names and every gate in the repo stays green.

This is exactly the gap `vta-service` closed in #857 with
`vta-service/src/trust_tasks/conformance.rs`, whose own module docs name the
defect class: *"Nothing checked **correctness**: that a task bound to a
published URI actually speaks that URI's schema."* The VTC never got the
equivalent.

## Two confirmed divergences

Both in the `join-requests` family, both against
`additionalProperties: false` response schemas.

**1. `vtc/join-requests/submit/0.1`** — the published Response is:

```json
{ "requestId": "…", "status": "pending" }
```

`required: [requestId, status]`, `additionalProperties: false`.

The service returns `VerdictResponse`:

```json
{ "requestId": "…", "verdict": { "effect": "…", "with": { … } } }
```

No `status` field at all (a *missing required* member), plus an unspecced
`verdict`. The ceremony verdict envelope
(`docs/05-design-notes/vtc-ceremony-protocol.md` §3) superseded the published
response shape and was never taken upstream.

**2. `vtc/join-requests/status/0.1`** — the published Response allows
`requestId`, `status`, `needs`, `presentationDefinition`, `ext`. #1058 added
`code`, `reason` and `decidedAt` on top. That was flagged in the PR under R3.*
and merged deliberately, on the grounds that the family already diverged — but
it widens the gap rather than closing it, and nothing will notice if it widens
again.

I have not audited the other 59. The point of the issue is that no one can say
whether they conform, because nothing asks.

## Why it matters

The VTC's whole claim in binding `https://trusttasks.org/spec/vtc/<slug>` is
[...]

PR #1111, opened 2026-08-26
test(vtc)!: make response conformance fail the build

stormer78:
The response-conformance layer now **fails the build**. Until this, a green
suite was not evidence of conformance — #1107's own memory note said to grep
the run rather than trust the exit code, which is a signal nobody has to obey.

Also retargets `relationships/{list,graph}` to the `0.2`s published in
trustoverip/dtgwg-trust-tasks-tf#266, closing the last VTC-family violations.

**The VTC family is now clean. Every remaining violation is `auth/*`.**

## The gate

A violation outside the allowlist returns `500` with the schema error, so the
test that provoked it fails on its own status assertion and prints the reason.

Chosen over panicking in middleware deliberately: a panic there surfaces at the
call site as a transport error — "the request failed" — which says nothing
about which task, or why.

## The allowlist, and why it can only shrink

Eight `auth/*` tasks, each listed with `ALLOWED_COUNT` asserted beside them.
Same discipline as `KNOWN_DRIFT_COUNT`, for the same reason: **the cheapest way
to make a failing check pass is to add a line to a list nobody is watching.**
Closing an entry means deleting it *and* lowering the count; adding one needs a
reason in review.

An allowlisted violation is still *reported*, just not fatal — pinned by a
test, because a list that suppressed the evidence would mean rediscovering each
entry before it could be closed.

## Two guards caught me, both correctly

**The allowlist named `login/{start,finish}/0.1`; the service binds `0.2`.**
Those two entries protected nothing, and the gate failed six step-up tests on
the first run. I had built the list from the violation *output* — which prints
whatever the request header said — rather than from the route mounts. Now taken
from `routes/mod.rs`.

**`every_bound_canonical_task_exists_in_the_registry` flagged a fake URI in my
own unit test.** I had used `trusttasks.org/spec/not/a/task/9.9` to exercise
the unknown-task path; that guard scans the crate for bound canonical URIs and
was exactly right that *"binding a `trusttasks.org/spec/` URI asserts the
registry serves it"*. Now `example.invalid`, with a comment saying why — a
guard that scans source for a URL pattern will find your fixtures too, and the
fix is to make the fixture obviously-not-real rather than to weaken the guard.

## A seeder that could never have caught a digest bug

`relationships/list` still failed after the `0.2` retarget, and it was not the
service: `seed_relationship` wrote `seed-{uuid}` into `vrcDigestMultibase`.
Unique per row, which is all the storage layer needed — and it meant **no test
ever exercised a digest that looked like a digest**, so the whole suite was
structurally blind to the format. It now computes a real multibase multihash
over the seeded credential.

## What remains

Eight `auth/*` tasks, all in the shared families both daemons serve. The models
genuinely differ (`passkeys` against `credentials`; `whoami`'s session needs an
`issuedAt` this service does not currently record).

The direction is settled: **follow the spec**, and where the spec cannot serve
one of the daemons, amend the spec — carefully — rather than bend an
implementation to fit. Each of the eight gets that question asked individually;
`whoami` is not a case of the spec being wrong but of this service not
recording something it should.

## Gates

- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test -p vtc-service --no-fail-fast` — 54 suites, no failures
- `cargo fmt --all --check`

`trust-tasks-rs` 0.11.14.

Refs #1059

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8681,9 +8681,9 @@ dependencies = [
 
 [[package]]
 name = "trust-tasks-rs"
-version = "0.11.13"
+version = "0.11.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c43a3909671dfa59719e98a836207b51d83bc2f53dfc82281207473354f83aa7"
+checksum = "09d00a1209aa2b7c0d5a012371ac356355b67511af00da974050aa8e61743987"
 dependencies = [
  "async-trait",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -244,7 +244,7 @@ aws-lc-rs = "1.16"
 # than fail quietly — but it would fail as "you are serving unspecced URIs",
 # which reads as an implementation fault instead of a stale dependency. The
 # pin makes the resolver say the true thing instead.
-trust-tasks-rs = { version = "0.11.13", default-features = false, features = ["validate"] }
+trust-tasks-rs = { version = "0.11.14", default-features = false, features = ["validate"] }
 trust-tasks-capability-client = "0.9"
 trust-tasks-https = { version = "0.10", default-features = false, features = ["server", "client", "jwt"] }
 # The DIDComm binding. Carries `ENVELOPE_TYPE` — the one `type` every
--- a/vtc-service/src/routes/mod.rs
+++ b/vtc-service/src/routes/mod.rs
@@ -19,7 +19,7 @@ pub(crate) mod members;
 pub(crate) mod policies;
 pub mod recognise;
 mod recognition_admin;
-mod relationships;
+pub(crate) mod relationships;
 mod schemas;
 pub(crate) mod status_lists;
 pub mod trust_tasks;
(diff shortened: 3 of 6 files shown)


stormer78:
Closing as **superseded, not abandoned** — every line of this PR is already on `main`, and merging it now would move the repository backwards.

This was the bottom of a stack. #1112 was built on top of it and merged first, carrying this PR's content in with it, so the branch is left holding an older copy of work that has already landed.

Verified file by file against `origin/main`:

| file | state |
|---|---|
| `vtc-service/src/routes/mod.rs` | byte-identical |
| `vtc-service/src/trust_tasks/conformance.rs` | byte-identical |
| `vtc-service/tests/relationships.rs` | byte-identical |
| `vtc-service/src/test_support/response_conformance.rs` | `main` is **ahead** |
| `Cargo.toml` | `main` is **ahead** |

The two where `main` is ahead are why this cannot simply be merged:

- **The gate is already live.** This branch sets `ALLOWED_COUNT = 8`; `main` sets it to **0**. The `INTERNAL_SERVER_ERROR` response — the actual "fail the build" change this PR is named for — is present in both. Merging would reopen eight allowlist entries that #1112 closed.
- **The dependency pin would regress.** This branch pins `trust-tasks-rs` at `0.11.14`; `main` is on `0.11.15`, which carries the upstream `extensions` amendment (trustoverip/dtgwg-trust-tasks-tf#267) that `auth/passkey/enroll/start/0.2` needs to conform at all.

So the outcome this PR set out to produce is in place: response conformance fails the build, with an empty allowlist. Nothing is lost by closing it.

Refs #1112

PR #1172, opened 2026-08-28
fix(ci): stop Dependabot emitting chore(deps)(deps) titles, and let a fixed title clear the check

stormer78:
Every cargo PR Dependabot opened in its first hour is red on `commit lint`:

```
#1170 chore(deps)(deps): bump tower-http from 0.6.11 to 0.7.0
#1169 chore(deps)(deps): bump azure_identity from 0.35.0 to 1.0.0
...
```

`include: scope` appends the scope itself, so a `prefix` that already carries one doubles it. The commit-lint job checks the **PR title** against the conventional-commit regex, which rejects `chore(deps)(deps):` — a check failing for a reason unrelated to the bump, on every PR, which is the fastest way to teach people to ignore it.

My bug, from #1160 an hour ago. The `github-actions` and `docker` entries were already correct — they set a scoped prefix and *don't* set `include`, which is the other consistent way to spell it. I'll retitle the six open ones by hand, since the check reads the title rather than the branch's commit messages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,12 @@ updates:
     # advisory is not subject to this cap.
     open-pull-requests-limit: 5
     commit-message:
-      prefix: "chore(deps)"
+      # `chore`, not `chore(deps)`: `include: scope` appends the scope itself,
+      # so a prefix that already carries one yields `chore(deps)(deps): …`,
+      # which the commit-lint regex rejects. It checks the PR *title*, so every
+      # Dependabot PR opened under the first version of this file was red on a
+      # check that had nothing to do with the bump.
+      prefix: "chore"
       # `!` is added by hand when a bump is breaking — the commit lint accepts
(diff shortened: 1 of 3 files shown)


stormer78:
Extended: the config fix stops Dependabot *emitting* the bad title, but the six PRs already open could not be repaired by fixing their titles.

`on: pull_request` in ci.yml carries no `types:`, so it defaults to opened/synchronize/reopened — an `edited` event never fires. The check reads `github.event.pull_request.title`, and re-running the job replays the original event payload, so it re-reads the *old* title. I retitled all five cargo PRs and every one came back red again.

So a PR whose only problem is its title can never go green by fixing the title. The only escape is pushing a commit that changes nothing.

Adding `edited` to ci.yml's trigger would re-run the docker builds, the enclave build and the mobile build on every description tweak. The job moves to its own `pr-title.yml` instead — the one check that depends on PR metadata becomes the one that reruns when PR metadata changes. Check name unchanged (`commit lint`), so branch protection still matches, and nothing `needs:` it.

PR #1222, opened 2026-08-31
feat(pnm-cli): add memory command group

marat-affinidi:
Adds a `pnm memory` command group wrapping the VTA's per-context agent-memory Trust Tasks (`spec/vta/memory/{put,list,delete}/0.1`).

Subcommands:
- `plant <key> <value>` — upsert an entry
- `recall [key]` — list entries, or one by key
- `forget <key>` — delete one entry
- `wipe` — delete all entries in a context (confirmation unless `--force`)

All are context-scoped via `--context` (default `default`) and support `--json`.

### DIFF
--- a/docs/02-vta/personal-ai-agents.md
+++ b/docs/02-vta/personal-ai-agents.md
@@ -226,11 +226,14 @@ in the agent runtime, the memory survives the runtime being replaced, is audited
 (`memory.put` / `.list` / `.delete`), and is included in the VTA's encrypted
 backups.
 
-The gate is **context access**, not a capability: `require_context(contextId)`,
-the same check the context-scoped key tasks use. That makes the trust context
-the isolation boundary — a context-A agent cannot read, write, or delete
-context-B memory — and it means the least-privilege grant for a memory-only
-agent is an ACL entry with role `application` naming exactly one context.
+Two gates apply, in this order. **Capability**: `MemoryRead` for
+`memory/list/0.1`, `MemoryWrite` for `put` and `delete` — so a `reader` can be
+given the memory without being given the ability to rewrite it. **Context
+access**: `require_context(contextId)`, the same check the context-scoped key
+tasks use, which makes the trust context the isolation boundary — a context-A
+agent cannot read, write, or delete context-B memory. The least-privilege grant
+for a memory-only agent is an ACL entry with role `application` naming exactly
+one context; `application` carries both memory capabilities.
 
 ```bash
 # One context per agent (or per domain) — this is the isolation boundary.
@@ -245,6 +248,23 @@ From a client, `VtaClient::{memory_put, memory_list, memory_delete}` drive the
 three tasks through the generic Trust-Task dispatcher, so they ride TSP, DIDComm
 or REST according to what both DID documents advertise — the agent writes no
 transport code.
+
+From the operator side the same three tasks are `pnm memory`:
+
+```bash
+pnm memory plant favourite-colour green --context my-agent-memory
+pnm memory recall --context my-agent-memory          # or `recall <key>`
+pnm memory forget favourite-colour --context my-agent-memory
+pnm memory wipe --context my-agent-memory            # confirms; --yes to skip
+```
+
+`--context` is required on every subcommand: a super-admin's access check
+passes for any id and these tasks do not require the context to exist, so a
+defaulted or mistyped id would silently read and write a context that isn't
+there — indistinguishable from an empty one. `wipe` needs both capabilities
+(it lists before it deletes) and is N round-trips, not atomic — there is no
+bulk-delete task — so re-running it after a partial failure is safe and
+finishes the job.
 
 Two boundaries worth stating before anything is stored here:
 
@@ -285,7 +305,8 @@ Lower priority unless your agents do credential work.
 | `pnm device …` / `pnm vault …` CLI     | **Added** (this change) — `VtaClient::{device_*,vault_*}` + `pnm` commands, both transports |
 | Sealed vault upsert/release            | **Added** — `seal_vault_secret` / `open_sealed_secret` on the DIDComm session |
 | `pnm acl --capabilities` flag          | **Gap** (deferred Phase 3) — capabilities derive from role / are set at `device/register` |
-| Agent memory Trust Tasks               | Exists (`trust_tasks/memory.rs`, `VtaClient::memory_*`) — context-gated, audited |
+| Agent memory Trust Tasks               | Exists (`trust_tasks/memory.rs`, `VtaClient::memory_*`) — capability- and context-gated, audited |
+| `pnm memory …` CLI                     | **Added** — `plant` / `recall` / `forget` / `wipe` over the same three tasks |
 | Agent-side connect ladder              | Exists (`vta_sdk::agent_connect::AgentConnect`) — one way in for every agent bridge |
 
 ## References
--- a/pnm-cli/src/cli.rs
+++ b/pnm-cli/src/cli.rs
@@ -278,6 +278,95 @@ pub(crate) enum Commands {
         #[command(subcommand)]
         command: DidTemplateCommands,
     },
+
+    /// Play with the VTA's per-context agent memory — the key/value notes the
+    /// hosted agent reads before it answers.
+    ///
+    /// The one-command demo that memory drives the model: `plant` a fact,
+    /// ask the agent, then `wipe` it and watch the answer change. Backed by
+    /// the `spec/vta/memory/{put,list,delete}/0.1` Trust Tasks; every
+    /// operation is gated on access to `--context` (default `default`).
+    #[command(name = "memory")]
+    LlmMemory {
+        #[command(subcommand)]
+        command: LlmMemoryCommands,
+    },
+}
+
+/// CRUD over the agent's memory. `plant` creates/updates, `recall` reads,
+/// `forget` deletes one entry, and `wipe` deletes every entry in the context.
+#[derive(Subcommand)]
+pub(crate) enum LlmMemoryCommands {
+    /// Plant a memory: store `value` under `key` so the agent recalls it.
+    ///
+    /// Upsert — re-planting the same key overwrites the old value.
+    Plant {
+        /// Memory key, unique within the context (e.g. `favorite-color`).
+        key: String,
+        /// The value to store (e.g. `green`).
+        value: String,
+        /// Context whose memory to write.
+        #[arg(
+            long,
+            short = 'c',
+            alias = "context-id",
+            value_name = "ID",
+            default_value = "default"
+        )]
+        context: String,
+    },
+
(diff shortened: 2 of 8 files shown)


stormer78:
Nice shape overall — the four verbs map cleanly onto the three Trust Tasks, and
CI is green. Four things before it comes out of draft, one of which I'd call
blocking.

### 1. `--json` silently skips the wipe confirmation (blocking)

```rust
if !force && !render::is_json_output() {
    // …confirm…
}
```

So `pnm memory wipe --json` deletes every memory in the context with no prompt
and without `--force`. An output-format flag isn't consent — someone piping to
`jq` to *read* the result gets the destructive path unguarded, and that is the
invocation a script is most likely to use.

The intent is presumably "don't block on a prompt when nobody can answer it",
which is right, but the conclusion should be to **refuse**, not to proceed:

```rust
if !force {
    if render::is_json_output() {
        return Err("`wipe` needs `--force` in --json mode: there is no one to \
                    confirm to".into());
    }
    // …confirm…
}
```

Same reasoning as `--force` elsewhere in the CLI: the flag is where the operator
says "yes, I mean it", and nothing else should be able to say it for them.

### 2. `wipe` loses its progress on a partial failure

```rust
for (key, _) in &items {
    client.memory_delete(context, key).await?;
    wiped.push(key.clone());
}
```

The `?` discards `wiped`, so a failure on entry 7 of 20 leaves six memories
deleted and reports only the error. The operator can't tell what went without
re-running `recall` and diffing by hand.

For a destructive bulk op that's worth the few lines: collect failures, keep
going or stop, and report both halves. The workspace rule this touches is R2.1
(*"process dies on the next line — then what?"*) — the answer here should be
"re-running `wipe` is safe and finishes the job", which it is, but only if the
operator can see where it stopped.

Also worth a line in the `wipe` doc comment: there is no bulk-delete Trust
Task, so this is N round-trips and not atomic. You say it in the code comment;
the operator-facing help is where it matters.

### 3. `extract_items` silently drops entries — which makes `wipe` under-report

```rust
/// Tolerant of a missing/oddly-shaped `items` array — it just yields fewer
/// pairs rather than erroring, so a demo never dies on a rendering technicality.
```

Fine for `recall`. Not fine for `wipe`, which derives *what to delete* from the
same function: an entry whose `value` isn't a string is skipped, so `wipe`
reports "wiped N — total amnesia" while leaving memories behind. The one
operation that must not under-count is the one promising there's nothing left.

Either make `wipe` count what it skipped and say so, or have it key off `key`
alone (it never uses the value).

### 4. The enum doc comment describes verbs that don't exist

```rust
/// CRUD over the agent's memory, with names that fit the demo. `plant`
/// creates/updates, `recall` reads, `wipe` deletes one, `amnesia` deletes
/// the lot.
```

`amnesia` isn't a subcommand, and `wipe` deletes *all*, not one — `forget` is
the single-key delete. Looks like a rename that didn't reach the comment.

### Not yours

The failing `semver report` reports `VtaError::UnsupportedTaskType` from #1220,
merged to `main` but not yet released. It stays red on every branch until the
Release PR publishes `vta-sdk`.

Happy to push 1–4 if you'd like, though it's your fork so I can't from here.

stormer78:
Second pass, this time against the checked-out branch rather than the diff. My
earlier four findings all hold; three more came out of reading it beside the
rest of the CLI, and one of them is the root cause of two others.

## New: the wire types already exist — `extract_items` shouldn't

`vta_sdk::protocols::memory` defines exactly what this is digging out by hand:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct MemoryListResponse { pub items: Vec<MemoryItem> }

pub struct MemoryItem { pub key: String, pub value: String }
```

So `extract_items` can be:

```rust
let resp: MemoryListResponse = serde_json::from_value(client.memory_list(context).await?)?;
```

This is the fix for two of my earlier points rather than a stylistic preference:

- the "tolerant of a missing/oddly-shaped `items` array" behaviour only exists
  because the value is untyped. Typed, a malformed response is an error instead
  of a silently-shortened list — which matters most for `wipe`, which derives
  *what to delete* from it and would otherwise report "total amnesia" over
  entries it never saw;
- the `--json` arm re-serialises a hand-built `{items:[{key,value}]}` to keep
  the shape stable. With the typed value that shape *is* the SDK's, and can't
  drift from it.

(`memory_list` returning `Value` rather than the typed body is the SDK's
inconsistency, not yours — `list_webvh_servers` and friends go through `rpc_tt`
and come back typed. Deserialising at the call site is the right move here; a
typed `memory_list` would be a separate SDK change.)

## New: this belongs in `vta-cli-common`

Every other command group is a thin dispatcher — `pnm-cli/src/commands/acl.rs`
is 40 lines of `match` calling `vta_cli_common::commands::acl::cmd_acl_*`, and
the implementation, rendering included, lives in the shared crate. CLAUDE.md
states the rule: *"`vta-cli-common` — Shared CLI command implementations; both
CLIs are thin wrappers."*

`llm_memory.rs` carries its whole implementation in `pnm-cli`, so `cnm` can
never have `memory` without a copy-paste — and a community operator wanting
agent memory is not an odd request. Moving it also makes it testable the way the
shared modules are (`acl.rs` has 17 tests; this PR has none).

## New: no test coverage

280 lines, no `#[cfg(test)]`. `extract_items` / the typed decode and the
one-vs-many pluralisation are pure functions and cheap to cover; the wipe
guard's decision (`force`, JSON mode, empty context) is worth a test precisely
because getting it wrong is destructive — see below.

## Still blocking: `--json` skips the wipe confirmation

```rust
if !force && !render::is_json_output() {   // …confirm…
```

`pnm memory wipe --json` deletes every memory in the context with no prompt and
no `--force`. `is_json_output()` is documented as an *output format* selector —
"commands that render tables check this and dispatch to a JSON serializer
instead of their human-readable renderer" — it carries no claim about
interactivity, let alone consent.

Refuse rather than proceed:

```rust
if !force {
    if render::is_json_output() {
        return Err("`wipe` needs `--force` in --json mode: there is no one to \
                    confirm to".into());
    }
    // …confirm…
}
```

## Still open from the first pass

- **`wipe` loses its progress on partial failure.** The `?` in the delete loop
  discards `wiped`, so failing on entry 7 of 20 leaves six deleted and reports
  only the error. Re-running is safe and finishes the job, but only if the
  operator can see where it stopped.
- **The enum doc comment describes verbs that don't exist** — it names
  `amnesia`, and says `wipe` "deletes one" when `forget` does and `wipe` deletes
  all. A rename that didn't reach the comment.
- Worth a line in `wipe`'s help that there is no bulk-delete Trust Task, so it
  is N round-trips and not atomic. It's in the code comment; the operator-facing
  text is where it matters.

## Not yours

The red `semver report` is #1220's `VtaError::UnsupportedTaskType`, merged but
unreleased — it stays red on every branch until #1218 publishes `vta-sdk`.

Happy to push any of this if you'd rather hand it over; it's your fork so I
can't from here.

marat-affinidi:
Thanks — addressed in the latest push.

- **`wipe --json` skipped confirmation** — now refuses without `--force` in JSON mode (no prompt to confirm to); the decision is a pure `wipe_guard(force, json)` with tests.
- **Partial-failure reporting** — the delete loop no longer discards progress; on a mid-loop failure it returns an error naming the count and the key it stopped on. Re-running resumes. Noted in the code and the `wipe` help.
- **`extract_items` silently dropping entries** — removed; the list response now decodes into the SDK's typed `MemoryListResponse`, so a malformed body errors instead of shortening. The `--json` arm re-emits that same body.
- **Stale enum doc comment** — corrected (`forget` deletes one, `wipe` deletes all; no `amnesia`).
- **Non-atomic `wipe`** — called out in the operator-facing help.
- **Belongs in `vta-cli-common`** — moved; `pnm-cli` is now a thin dispatcher, and the implementation is reusable by `cnm`.
- **Tests** — added for the wipe guard, the typed decode (incl. the malformed case), and pluralisation.

`semver report` stays red on the unreleased `VtaError::UnsupportedTaskType` as you noted — not from this change.

stormer78:
Reviewed, and I've pushed `fd3225ea` to the branch rather than hand you a list — everything below is already applied, so read it as a changelog and push back on anything you disagree with.

The shape is good and the earlier round is properly addressed: typed decode through `MemoryListResponse`, the `wipe_guard` refuse-in-JSON fix with tests, partial-failure reporting that names where it stopped, the implementation in `vta-cli-common` where a shared command belongs. Two real findings, then a consistency pass.

### 1. `--context` defaulting to `"default"` (the one that mattered)

Nothing in `vta setup` creates a context with the id `default` — context ids are operator-chosen slugs — and a super-admin's context check is `ActScope::Unrestricted.covers(_)`, which is true for *any* string. The memory handlers don't require the context to exist either. So the no-flag path was a silent miss:

```
pnm memory plant favourite-colour green      # writes into a context that need not exist
pnm memory recall                            # "Context 'default' has no memories."
```

That last line reads as a legitimate empty answer, not "you forgot to scope this" — and a mistyped `--context` behaved identically. `--context` is now required on all four subcommands, the same shape as `pnm keys create --context`, with clap refusing as the guard and a test pinning it.

### 2. The gate description went stale under you

#1234 landed `MemoryRead` / `MemoryWrite` and put them *in front of* the context check, and it's in this branch's history — so the help text ("every operation is gated on access to `--context`") and the module doc were describing the pre-#1234 world. Corrected in both, plus `docs/02-vta/personal-ai-agents.md`, which still carried the same sentence from before the split. Two consequences worth stating in operator-facing text: `wipe` needs *both* capabilities because it lists before it deletes, and a `reader` now holds `MemoryRead` only, so every write is refused for that role.

That doc also gained the `pnm memory` walkthrough it was missing — it previously told operators the only client was `VtaClient::{memory_put, …}`.

### 3. Consistency with the rest of the CLI

- **`--force` → `--yes`.** `--force` already means "hard-delete instead of soft-delete" on `pnm vault delete` / `pnm cred-vault delete`; the skip-the-prompt flag elsewhere is `--yes` (`pnm keys create`). Two meanings for one flag in one binary is the kind of thing that gets someone once. `--force` stays as a hidden alias, and both spellings are tested.
- **Dropped `-c` / `-f`.** No other pnm subcommand has command-level shorts.
- **`LlmMemory` / `llm_memory.rs` → `Memory` / `memory.rs`**, matching the command, the shared module, and the workspace's "agent memory" vocabulary. Each subcommand also takes a hidden alias under the Trust Task's own verb — `put` / `list` / `delete` / `clear` — so someone reading the task names finds the command.
- Help text de-demoed: `pnm memory --help` is operator documentation, so "Play with…" and "the one-command demo that memory drives the model" are gone.

### Not yours

The red `semver report` is `constructible_struct_adds_field` on the `ext` members #1231 added across ten SDK bodies (`MemoryPutBody.ext` among them). It's informational and stays red on every branch until the Release PR publishes `vta-sdk`. Don't bump anything.

Nothing outstanding from me beyond CI going green.

PR #1229, opened 2026-08-31
test(trust-tasks): fix doc() helper ignoring issued_at, cover the unboundable-window case

marat-affinidi:
## What

Fixes a bug in the `freshness_bounds::doc()` test helper: it always seeded a fresh default `issuedAt` regardless of its `issued_at: Option<&str>` argument, so `doc(None, None)` never actually built a document without one, and the test that name implied went unexercised for as long as it existed.

## What this PR does

With the helper fixed to genuinely omit `issuedAt` when `None` is passed, replaces the one test that relied on the broken helper with three that cover the actual behavior:

- a truly timestamp-less document is refused under a bounded `FreshnessPolicy` (`max_age.is_some()`) with `RejectReason::Stale { detail: Unboundable }` — SPEC §7.2 *Bounding the record*: a consumer retaining a bounded window has nowhere to place a document it can't measure the age of.
- `expiresAt` alone is sufficient without `issuedAt`.
- the same timestamp-less document is accepted under a policy with no window at all.

All six `freshness_bounds` tests pass with nothing ignored. `cargo test`, `cargo clippy`, and `cargo fmt --all --check` are clean.

## History

The first version of this PR mischaracterized the wire-level `expired` code on this path as a §7.2 item 13 violation. [@stormer78 corrected this](https://github.com/OpenVTC/verifiable-trust-infrastructure/pull/1229#issuecomment-5490325599): the `expired` code here is `trust-tasks-rs`'s mapping of an internal `Stale { detail: Unboundable }` reason (there's no narrower standard code for it), governed by §7.2's *Bounding the record* clause, not item 13 — and it's driven by this VTA's own `freshness_policy()` choosing a bounded 10-minute window, not by a defect in the crate. Verified directly against the `trust-tasks-rs` 0.17.3 source (`freshness.rs:216-228`, `error.rs:525`). No upstream report needed.

### DIFF
--- a/vta-sdk/src/session.rs
+++ b/vta-sdk/src/session.rs
@@ -2981,6 +2981,11 @@ fn now_epoch() -> u64 {
 ///
 /// `issuedAt` alone is the right bound to carry: an `expiresAt` would put the
 /// consumer's retention obligation in this client's hands.
+///
+/// Gated with its three callers: `TspPingSession` and `TspSession` are both
+/// `tsp`-only, so under `session` alone this is dead code and `-D warnings`
+/// fails the `vta-service rest only` feature combo.
+#[cfg(feature = "tsp")]
 fn ping_document(
     client_did: &str,
     vta_did: &str,
@@ -2999,7 +3004,7 @@ fn ping_document(
     Ok(doc)
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "tsp"))]
 mod ping_document_tests {
     use super::ping_document;
 
--- a/vta-sdk/src/session.rs
+++ b/vta-sdk/src/session.rs
@@ -2321,17 +2321,16 @@ impl TspPingSession {
         timeout: std::time::Duration,
     ) -> Result<u128, Box<dyn std::error::Error>> {
         use std::time::Instant;
-        use trust_tasks_rs::TrustTask;
-
-        let id = format!("urn:uuid:{}", uuid::Uuid::new_v4());
-        let nonce = uuid::Uuid::new_v4().to_string();
-        let type_uri = crate::trust_tasks::TASK_MESSAGING_PING_0_1
-            .parse()
-            .map_err(|e| format!("messaging/ping type URI parse: {e}"))?;
-        let mut doc: TrustTask<serde_json::Value> =
-            TrustTask::new(id.clone(), type_uri, serde_json::json!({ "nonce": nonce }));
-        doc.issuer = Some(self.client_did.clone());
-        doc.recipient = Some(vta_did.to_string());
+        let doc = ping_document(&self.client_did, vta_did)?;
+        let id = doc.id.clone();
+        // Read back off the document rather than kept alongside it: the two
+        // must be the same value for `correlates` to mean anything.
+        let nonce = doc
+            .payload
+            .get("nonce")
+            .and_then(|v| v.as_str())
+            .ok_or("messaging/ping payload carries no nonce")?
+            .to_string();
         let body = serde_json::to_vec(&doc)?;
 
         let start = Instant::now();
@@ -2412,18 +2411,7 @@ impl TspPingSession {
     /// [`ping`](Self::ping) does) always times out and masks the send success.
     /// `Ok(())` means the relationship-free routed send worked (3c).
     pub async fn probe_send(&self, vta_did: &str) -> Result<(), Box<dyn std::error::Error>> {
-        use trust_tasks_rs::TrustTask;
-
-        let id = format!("urn:uuid:{}", uuid::Uuid::new_v4());
-        let nonce = uuid::Uuid::new_v4().to_string();
-        let type_uri = crate::trust_tasks::TASK_MESSAGING_PING_0_1
-            .parse()
-            .map_err(|e| format!("messaging/ping type URI parse: {e}"))?;
-        let mut doc: TrustTask<serde_json::Value> =
-            TrustTask::new(id, type_uri, serde_json::json!({ "nonce": nonce }));
-        doc.issuer = Some(self.client_did.clone());
-        doc.recipient = Some(vta_did.to_string());
-        let body = serde_json::to_vec(&doc)?;
+        let body = serde_json::to_vec(&ping_document(&self.client_did, vta_did)?)?;
 
         self.identity
             .hub
@@ -2616,18 +2604,7 @@ impl TspSession {
         vta_did: &str,
         mediator_did: &str,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        use trust_tasks_rs::TrustTask;
-
-        let id = format!("urn:uuid:{}", uuid::Uuid::new_v4());
-        let nonce = uuid::Uuid::new_v4().to_string();
-        let type_uri = crate::trust_tasks::TASK_MESSAGING_PING_0_1
-            .parse()
-            .map_err(|e| format!("messaging/ping type URI parse: {e}"))?;
-        let mut doc: TrustTask<serde_json::Value> =
-            TrustTask::new(id, type_uri, serde_json::json!({ "nonce": nonce }));
-        doc.issuer = Some(self.client_did.clone());
-        doc.recipient = Some(vta_did.to_string());
-        let body = serde_json::to_vec(&doc)?;
+        let body = serde_json::to_vec(&ping_document(&self.client_did, vta_did)?)?;
 
         self.send_document(vta_did, mediator_did, &body).await
     }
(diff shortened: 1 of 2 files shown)


marat-affinidi:
Checked this against the framework spec (trustoverip/dtgwg-trust-tasks-tf, `SPEC.md`) and the "question for maintainers" above has a clear answer: **the observed behavior is a spec violation, not an open interpretation question.**

- §8.3 (Standard error codes) defines `expired` narrowly: *"The document's `expiresAt` was in the past at the time of evaluation."* Nothing about `issuedAt`, and nothing about an absent timestamp.
- §7.2 item 13 defines exactly two malformed conditions on timestamps (future-dated `issuedAt` beyond skew; `expiresAt` at-or-before `issuedAt`) and states outright: *"Both are `malformedRequest` rather than `expired`, because `expired` names a document that was once acceptable and neither of these ever was."*
- §7.3 item 17 only requires `issuedAt` for *consequential* Trust Tasks (mutating/destructive, or disclosing secret material, or acting as the subject) — it stays **SHOULD** for everything else.

The repro document in this PR carries neither `issuedAt` nor `expiresAt`. Per the spec, that's not `expired` (no `expiresAt` is even present to be "in the past"), and it isn't `malformedRequest` under item 13 either (neither of its two clauses applies to an absent timestamp). So `validate_freshness` returning `Expired` here doesn't match any defined outcome for this input.

Since `validate_freshness` is `trust-tasks-rs`'s own method (called directly, with nothing VTI-specific in between), this likely needs a matching report against that crate rather than a fix in this repo — flagging here since this is where I found it and where the affected test lives.

marat-affinidi:
Correction — the links above 404 because I linked the *annotated tag object*'s SHA (\`e9c74c4\`) instead of the commit it points to. Dereferenced properly this time; verified all three resolve (HTTP 200) before posting:

- [§8.3 Standard error codes](https://github.com/trustoverip/dtgwg-trust-tasks-tf/blob/27401b7d66ffc2855a3119b8de073db14cd38fa2/SPEC.md#83-standard-error-codes) — the \`expired\` row: *"The document's \`expiresAt\` was in the past at the time of evaluation."*
- [§7.2 Consumer requirements](https://github.com/trustoverip/dtgwg-trust-tasks-tf/blob/27401b7d66ffc2855a3119b8de073db14cd38fa2/SPEC.md#72-consumer-requirements) — item 13 and its explanatory paragraph: *"Both are \`malformedRequest\` rather than \`expired\`, because \`expired\` names a document that was once acceptable and neither of these ever was."*
- [§7.3 Specification requirements](https://github.com/trustoverip/dtgwg-trust-tasks-tf/blob/27401b7d66ffc2855a3119b8de073db14cd38fa2/SPEC.md#73-specification-requirements) — item 17, scoping the \`issuedAt\` **MUST** to *consequential* Trust Tasks only.

Still pinned to the commit tagged \`trust-tasks-rs-v0.17.3\` (now the correct underlying commit, \`27401b7d\`), confirmed identical to current \`main\`.

stormer78:
The helper bug is real and worth having: `doc()` seeded a default `issuedAt`
whatever it was passed, so `doc(None, None)` never built a timestamp-less
document and that case went unexercised for as long as it has existed. Good
catch, and the repro is exact.

The diagnosis it leads to doesn't hold up, though, and I'd rather not merge the
`#[ignore]` — an ignored test is a claim nobody re-checks, and this one asserts
a spec violation against `trust-tasks-rs` that isn't there.

**It isn't returning `expired`.** Probed directly on this branch:

```
doc(None, None)              -> Err(Stale { detail: Unboundable })
doc(None, Some(expiresAt))   -> Ok(())
doc(None, None), no max_age  -> Ok(())
```

`Stale` maps onto the `expired` *standard code* because the framework has no
narrower one, so the wire says `expired` while the reason says something else
entirely. Reading the code alone is what makes this look like a contradiction.

**`Unboundable` is deliberate, and cited.** `freshness.rs:221` says so two
lines above the branch:

> No `issuedAt`. A policy with a window cannot place the document in it unless
> the producer supplied an `expiresAt` instead (SPEC §7.2, *Bounding the
> record*).

That's the rule that fires here, not §7.2 item 13. Item 13 is quoted correctly
in the PR description — it defines two *malformed* timestamp conditions and
neither applies — but it isn't the only clause in §7.2, and *Bounding the
record* is the one that governs a document with nothing to bound it by. This
VTA's `freshness_policy()` sets `max_age = 10 minutes`, so a document carrying
neither timestamp genuinely has nowhere to sit in that window.

The second and third probes settle it: `issuedAt` is **not** required — an
`expiresAt` alone is accepted — and the same timestamp-less document is
accepted under a policy with no window. So the refusal belongs to the policy,
not to the document, and the original comment ("item 13 bounds the timestamps a
document carries; it does not require one") was right all along. It just wasn't
the whole rule.

So: no upstream report needed against `trust-tasks-rs`.

**Suggested change** — keep your helper fix, drop the `#[ignore]`, and replace
the one test with three that assert the behaviour and show why:

```rust
/// A document carrying **no timestamp at all** is refused — and the rule is
/// the *window*, not `issuedAt`. SPEC §7.2, *Bounding the record*.
#[test]
fn a_document_with_no_timestamps_cannot_be_placed_in_the_window() {
    let err = doc(None, None)
        .validate_freshness(Utc::now(), &freshness_policy())
        .expect_err("an unbounded document cannot sit in a bounded window");
    assert!(
        matches!(
            err,
            RejectReason::Stale { detail: trust_tasks_rs::StaleReason::Unboundable }
        ),
        "got {err:?}"
    );
}

/// `issuedAt` is not required: `expiresAt` alone bounds the record. This is
/// the assertion the ignored test was reaching for.
#[test]
fn an_expiry_alone_bounds_a_document_with_no_issued_at() {
    let expires = (Utc::now() + chrono::TimeDelta::minutes(5))
        .to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
    assert!(
        doc(None, Some(&expires))
            .validate_freshness(Utc::now(), &freshness_policy())
            .is_ok(),
        "an `expiresAt` is enough to bound a document with no `issuedAt`"
    );
}

/// And with no window to place it in, a timestamp-less document is fine — so
/// the refusal belongs to the policy, not the document.
#[test]
fn without_a_window_a_timestampless_document_is_accepted() {
    assert!(
        doc(None, None)
            .validate_freshness(Utc::now(), &trust_tasks_rs::FreshnessPolicy::default())
            .is_ok()
    );
}
```

I ran these locally against your branch — all six `freshness_bounds` tests pass
with nothing ignored. Happy to push the change if you'd rather I did; it's your
fork so I can't from here.

Two notes unrelated to the substance:

- The failing `semver report` on this PR is not yours. It reports
  `VtaError::UnsupportedTaskType` from #1220, which is merged to `main` but not
  yet released, so it will stay red on every branch until the Release PR
  publishes `vta-sdk`.
- The helper fix on its own is worth keeping regardless of how the assertions
  land — a helper that ignores its own argument will mislead the next person
  the same way.

marat-affinidi:
You're right, and I verified it against the actual source rather than taking it on faith — thank you for the precise correction.

Checked \`trust-tasks-rs-0.17.3\`'s \`freshness.rs:216-228\`: the \`Unboundable\` branch fires exactly as you described, gated on \`policy.max_age.is_some() && self.expires_at.is_none()\` with no \`issuedAt\` present — the comment two lines above it cites SPEC §7.2 *Bounding the record* verbatim. And \`error.rs:525\` confirms \`RejectReason::Stale { .. } => StandardCode::Expired\` — so the wire code really is just the closest available standard code for an internal reason the framework's fixed vocabulary doesn't have a narrower slot for. My mistake was reading the wire-level \`expired\` string as if it necessarily meant §8.3's narrow definition applied, instead of checking the actual \`RejectReason\` variant.

Pushed your suggested three tests in place of the \`#[ignore]\`d one — all six \`freshness_bounds\` tests pass now, nothing ignored. Updated the PR description to drop the "spec violation" framing. Appreciate you running it locally and citing the exact lines rather than just asserting it — made this fast to verify.

stormer78:
Reviewed this and pushed two commits addressing what came out of it. The helper fix here is correct — `TrustTask::issued_at` is `#[serde(default)]`, so `doc(None, None)` now genuinely omits the member, and the three `doc(Some(..), ..)` call sites are unaffected. What the fix exposed is that the case it uncovers is answered wrongly, and that a live producer in this workspace has been tripping over it since #1127.

## 1. An unboundable document was answered `expired`, and the spec's own rule never ran

`3673e317` — `fix(vta)!: require issuedAt, and cap replay retention at the acceptance window`

`validate_freshness` runs at the top of `dispatch_trust_task_validated`, roughly 250 lines before `schema_index::spec_policy_for(..).enforce(..)`. Under a bare `with_max_age`, a document carrying neither timestamp is refused there as `Stale { detail: Unboundable }` → wire code `expired`. That means for the **105 of 209** indexed specs that declare `IS_ISSUED_AT_REQUIRED`, `SpecPolicy::enforce`'s `malformedRequest` + `ISSUED_AT_REQUIRED_BY_SPEC` — the code that actually names the missing member — was unreachable.

The consequence lands on the producer: `expired` names a document that was once acceptable and tells it to wait. This one was never acceptable, so it waits, reissues the identical shape, and loops. That is the principle `a_future_dated_document_is_malformed_not_expired` already pins three tests above, and `trust-tasks-rs` says the same thing in its own docs on `IS_ISSUED_AT_REQUIRED`: `expired` "would misdescribe a document that was never acceptable".

There was a second gap alongside it. A document with **only** `expiresAt` was accepted — an `expiresAt` does bound the record — but the bound was then whatever instant the producer picked. `expiresAt = now + 10 years` is accepted for ten years, and §7.2 makes the replay record's retention that same bound, so the producer decided unilaterally how long this VTA must remember its `id`.

`freshness_policy()` now adds `.requiring_issued_at()`, which closes both and makes the last instant any accepted document can return provable — `issuedAt + max_age + skew`.

## 2. Replay retention was uncapped

Same commit. `record_expiry` takes a producer-supplied `expiresAt` **verbatim**, so that ten-year document pinned its `id` in `REPLAY_GUARD` for ten years. Bounded by `InMemoryReplayGuard`'s capacity, so not exhaustion — but an entry held long past the last moment it could be needed crowds out the ones that are, which is exactly the load-dependent eviction horizon `freshness_policy()`'s own doc comment says #1127 exists to close.

A new `retain_until` helper caps at the acceptance window. It only ever moves retention **earlier than the producer asked for**, never earlier than the window — shortening it below the window is the direction §7.2 forbids, because a replay arriving while the document is still acceptable with its record already dropped executes twice.

## 3. `messaging/ping` has been refused by every bounded VTA since #1127

`4120b3b4` — `fix(sdk): stamp issuedAt on messaging/ping, which a bounded VTA refuses`

This is the one worth reading closely, because it is a live defect rather than a code-quality point, and this PR's helper fix is what surfaced it.

`TspPingSession::ping`, `TspPingSession::probe_send` and `TspSession::announce` each built a `messaging/ping/0.1` document carrying **neither** `issuedAt` nor `expiresAt`. `messaging/ping/0.1` dispatches through the spine (`trust_tasks/mod.rs:1351`), so from #1127 onward the liveness probe has been refused at `validate_freshness` before reaching the handler — and refused as `expired`, which reads as a transport delay rather than a malformed document. The failure looked exactly like the timeout it is not. A liveness probe that a live peer refuses is worse than no probe: it reports the transport broken when the transport is fine.

The three sites are now one `ping_document` builder, since the thing they had in common was the member they all left out. `ping` reads its correlation nonce back off the built document rather than keeping it alongside — the two have to be the same value for `correlates` to mean anything.

## On the tests

Rewrote the three new ones so they assert against this service's own policy:

- `a_document_with_no_timestamps_is_malformed_not_expired` — now asserts the `malformedRequest` code and the `ISSUED_AT_REQUIRED` reason, not `Stale { Unboundable }`.
- `an_expiry_alone_bounds_the_record_but_not_this_consumers_window` — keeps the old assertion but against `FreshnessPolicy::default()`, and adds the refusal under `freshness_policy()`. The difference between the two policies is the point.
- `without_a_window_a_timestampless_document_is_accepted` is gone. It validated against a `FreshnessPolicy::default()` this service never uses, so it passed regardless of what `freshness_policy()` did — including its outright deletion, which is the same "cannot fail on a regression in the code under review" class this PR set out to remove. Replaced with `the_services_policy_bounds_both_the_window_and_the_member`, which pins the service's own policy.
- New `record_retention` module covers the cap in both directions, and the SDK test asserts against `FreshnessPolicy::consequential()` rather than `issued_at.is_some()` — the latter passes on a document the consumer still refuses, so it would not have caught the ping bug.

I reverted each fix in turn and confirmed the corresponding test fails. `cargo test --workspace` is green (155 suites), `cargo clippy --workspace --all-targets` and `cargo fmt --all --check` clean.

## Two things for you to decide

**The `requiring_issued_at()` posture is stricter than most specs, and that is a judgement call worth pushing back on.** 104 of the 209 indexed URIs — the read-shaped ones: `acl/list`, `auth/whoami`, `config/show`, `device/heartbeat` — leave `issuedAt` OPTIONAL, so a conforming producer of one may omit it. For all but one shape that costs nothing, because the refusal already happens and only the code changes. The single shape moving from **accepted to refused** is `expiresAt` present with `issuedAt` absent. No in-repo producer emits that, but an external client might.

The justification is that this is the posture `FreshnessPolicy::consequential` describes, which the crate scopes to "any specification whose execution is consequential, which is exactly the set for which item 11 applies" — and this spine applies item 11 to **every** document it dispatches, `whoami` included. If you would rather not take that, the narrower fix is to leave the policy permissive and translate `Stale { Unboundable }` into `MalformedRequest` at `reject_with`. That fixes the misleading code with zero behaviour change, but leaves the producer-controlled-retention hole open. Happy to switch if you prefer it.

**The PR title needs to change before squash.** It is `test(trust-tasks): …`, and since the title is the changelog entry, a behaviour change would land in the changelog as a test change. Suggest something like `fix(vta)!: require issuedAt, so an unboundable Trust Task is malformed and not expired`.

stormer78:
CI caught one thing I missed, fixed in `b277995d`.

`Feature combos` failed on `vta-service rest only` with `error: function ping_document is never used`. `TspPingSession` and `TspSession` are both `tsp`-gated, so under `session` alone the shared builder has no consumer and `-D warnings` rejects it. The helper and its tests are now gated on `tsp` with their callers.

Worth noting how it got past me: `cargo test --workspace` unifies features across the graph, and several crates turn `vta-sdk/session` and `tsp` on, so every local run had the callers compiled and was green. Only a narrowed combo builds the function without them — which is exactly what that CI step says it exists to catch.

Re-ran the three combos touching these files before pushing, all clean: `rest only`, `tsp without didcomm`, `workspace --all-features --all-targets`, plus `cargo test -p vta-sdk --all-features` (714 tests, both new ones present).

**CI is now green.** The one remaining red is `semver report (informational — never blocks)`, and it is not from this branch: the two findings are `constructible_struct_adds_field` in `vta-sdk` and `enum_variant_added` in `vti-common`, neither of which these commits touch — no public struct field or enum variant is added here. Pre-existing against the published versions, and per CLAUDE.md the bump belongs to the Release PR, not this branch.

PR #1231, opened 2026-09-01
fix(sdk): accept the `ext` member every payload schema declares

stormer78:
SPEC §4.5.1 gives every Trust Task payload an `ext` slot for ecosystem-defined extension members, and the published schemas declare it. Sixteen `deny_unknown_fields` structs had no field for it, so a producer doing exactly what the schema permits had its whole document rejected.

## The failure

```
malformed request: payload parse: unknown field `ext`,
expected one of `role`, `scope`, `direction`, `subjectPrefix`, `pageSize`, `cursor`
```

`specs/acl/list/0.1/payload.schema.json` lists `ext` among its properties. So does `policy/list/0.2`. The client was told it could send one; the agent rejected the document for containing it.

## Scope

| Family | Structs |
|---|---|
| `acl/*` | `CreateAclBody`, `DeleteAclBody`, `GetAclBody`, `ListAclBody`, `AclEntry` |
| `vta/memory/*` | `MemoryPutBody`, `MemoryListBody`, `MemoryDeleteBody` |
| `policy/*` | `ListPoliciesBody`, `GetPolicyBody`, `DeletePolicyBody` |
| `vta/credentials/*` | `IssueCredentialBody`, `RevokeCredentialBody` |
| `vta/app-state/*` | `AppStateWrite` |
| `config/*` | `GetConfigBody`, `UpdateConfigBody` |

Seven sibling structs already carry `ext`, with the reasoning written out on each — `AppStateGetBody` and `UpsertPolicyBody` among them. **This completes that work rather than starting it**, and reuses their wording verbatim.

`deny_unknown_fields` stays. Carrying `ext` explicitly is what keeps a *typo* refused, which is the guard that clause was there for, while letting through the one member the spec says is always allowed. Relaxing the clause instead would trade a real defect for a worse one.

## How it was found, and why nothing caught it

A browser-based VTA management console. Its Access and Policy panes died outright, and the operator was shown a parse error naming a field the spec had told the client it could send.

Whether a caller trips this is decided entirely by whether it populates `ext` — so the same agent looks healthy to one client and broken to another. The conformance table did not catch it because it exercises the members its fixtures *set*, and this defect lives in the member they leave unset. `keys/list/0.1` worked fine throughout, which is why only two panes failed rather than all of them.

## The guard

Another fixture would have the same blind spot, so the guard is a census over the source, matching the existing `payload_null_census.rs`:

`vta-sdk/tests/payload_ext_census.rs` fails on any `deny_unknown_fields` type under `protocols/` that carries no `ext`, naming each one and printing the field to add. Exceptions go in `NO_EXT_BY_DESIGN` and must state a reason — there is no way to pass by silence. It asserts it inspected a plausible number of types, so a broken walk fails rather than passing vacuously.

Parsed with `syn`, not grepped: the attribute is routinely written across several lines, and a line-oriented regex misreads that in the direction that lets a violation through.

**Verified to fail:** reverting `ListAclBody` alone produces

```
1 wire type(s) deny unknown fields but carry no `ext` …
  acl_management/list.rs::ListAclBody
```

## Checks

- `cargo test -p vta-sdk` — 300 + all integration suites pass
- `cargo test -p vta-service --lib trust_tasks::conformance` — 2 passed
- `cargo clippy --workspace --all-targets` — no warnings

## Not in this PR

Two adjacent gaps found from the same console, both needing spec work rather than a code fix — happy to open either if wanted:

1. **`vta/credentials` has no list task.** Only `issue` and `revoke` are specified, so a client can mint and revoke a credential but cannot enumerate what it has issued. The console renders an explicit "there is no list, and this is not an empty one" notice rather than a blank table it has no basis for.
2. **`vault/credentials/*` is implemented but unspecified.** `vta-sdk/src/client/vault.rs` dispatches `TASK_VAULT_CREDENTIALS_QUERY_0_1` and friends (query, get, receive, archive, unarchive, delete, restore, purge), and `pnm cred-vault` drives them, but there is no entry under `specs/vault/credentials/` and so no generated bindings. Any TypeScript client wanting the holder-side credential store has to hand-transcribe the shapes from the Rust — a copy that drifts.

### DIFF
--- a/vta-cli-common/src/commands/config.rs
+++ b/vta-cli-common/src/commands/config.rs
@@ -60,7 +60,10 @@ pub async fn cmd_config_update(
 
     let resp = client
         .update_config(UpdateConfigRequest {
-            patch: UpdateConfigBody { overrides },
+            patch: UpdateConfigBody {
+                overrides,
+                ext: None,
+            },
         })
         .await?;
 
--- a/vta-cli-common/src/commands/policy.rs
+++ b/vta-cli-common/src/commands/policy.rs
@@ -28,6 +28,7 @@ pub async fn cmd_list(
             enabled_only,
             cursor: None,
             page_size: None,
+            ext: None,
         })
         .await?;
 
@@ -129,6 +130,7 @@ pub async fn cmd_delete(
             id: id.to_string(),
             expected_version,
             reason,
+            ext: None,
         })
         .await?;
     println!("Deleted policy {} at {}", result.id, result.deleted_at);
--- a/vta-sdk/src/client/mod.rs
+++ b/vta-sdk/src/client/mod.rs
@@ -2698,7 +2698,10 @@ mod tests {
         let mut overrides = std::collections::HashMap::new();
         overrides.insert("vta_name".to_string(), serde_json::json!("Test"));
         let req = UpdateConfigRequest {
-            patch: UpdateConfigBody { overrides },
+            patch: UpdateConfigBody {
+                overrides,
+                ext: None,
+            },
         };
         let json = serde_json::to_value(&req).unwrap();
         assert_eq!(json["overrides"]["vta_name"], "Test");
--- a/vta-sdk/src/client/policy.rs
+++ b/vta-sdk/src/client/policy.rs
@@ -43,7 +43,10 @@ impl VtaClient {
 
     /// `policy/get/0.1` — read one policy module by id.
     pub async fn get_policy(&self, id: &str) -> Result<GetPolicyResultBody, VtaError> {
-        let req = GetPolicyBody { id: id.to_string() };
+        let req = GetPolicyBody {
+            id: id.to_string(),
+            ext: None,
+        };
         self.rpc_tt(
             crate::trust_tasks::TASK_POLICY_GET_0_1,
             serde_json::to_value(&req)?,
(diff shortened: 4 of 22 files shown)


stormer78:
Flagging this before a reviewer has to find it: **the `semver report` job fails, and it is right to.**

Adding a field to a `pub struct` whose fields are public is a breaking change in Rust — every struct-literal construction site has to be updated. That is not incidental to this PR, it is the shape of it: 30-odd construction sites across `vta-sdk`, `vta-cli-common` and the service tests needed `ext: None`, and they are in the diff.

The job is marked "informational — never blocks", so nothing is being overridden. But the call is worth making explicitly rather than by silence:

- `vta-sdk` is `0.32.2`. Under Cargo's semver rules a `0.x` minor bump is where breaking changes go, so this is expressible without a major version.
- The alternative that would avoid the break is `#[non_exhaustive]` on the affected structs. That trades this one-time break for a permanent one — downstream crates lose struct-literal construction entirely and must go through a builder or `..Default::default()`. For 16 payload types whose whole purpose is to be constructed literally, that seems the worse trade, but it is a call for whoever owns the crate's compatibility policy rather than mine.
- Doing nothing is not neutral. The structs currently reject a member their own published schemas declare, so any conforming producer that populates `ext` gets its document refused — which is the defect, and it does not go away by deferring the field.

Happy to switch to `#[non_exhaustive]`, split the SDK changes from the service changes, or hold this behind a version bump — whichever fits the release plan.

PR #1234, opened 2026-09-01
feat(acl): add MemoryRead/MemoryWrite and gate the memory tasks on them

stormer78:
Closes the orthogonal gap called out in `docs/05-design-notes/data-rooms.md` §11.1
(design PR #1233). Independent of that design — the gap exists today.

## The gap

**There is no read-only grant on agent memory.** The gate in
`vta-service/src/trust_tasks/memory.rs` is `auth.require_context(&req.context_id)`
and nothing else, and a context is binary: any DID that can reach one can also
write and delete every memory in it. An operator cannot give an agent read-only
access to their own memory. The `--read-only` flag in `vta-mcp/src/guard.rs` is a
client-side glob filter on the MCP bridge — a caller talking to the VTA directly
never encounters it.

**The published spec already assumes otherwise.**
`specs/vta/memory/delete/0.1/spec.md` reasons at length about *"a VTA whose write
capability is granted more freely than its read capability"* and about a caller
holding write "without ever holding the read capability that `vta/memory/list`
requires". There was no read capability and no write capability. There was a
context.

The ACL supplied nothing finer either: the act axis is
`(role, allowed_contexts)` decoded to a three-valued `ActScope` — a *where* with
no *what*.

## The change

`Capability::{MemoryRead, MemoryWrite}`, wired through
`derived_capabilities_for_role`, checked in all three handlers. Legacy rows carry
no explicit capability set and fall back to the derived mapping, so the roles
that write memory today keep writing it.

| role | before | after |
|---|---|---|
| `admin`, `initiator` | read + write (via context) | read + write |
| `application` | read + write (via context) | read + write |
| `reader` | **read + write** (via context) | **read only** |
| `monitor` | **read + write** (via context) | **none** |

**Two deliberate tightenings**, both of which are the point of the change:

- **`reader` loses memory write.** A read-only consumer of a context should not
  be able to rewrite the memories in it.
- **`monitor` loses memory access entirely.** It is the least-privileged role
  and the `Default` for `AuthClaims` — *"a test fixture that leaks past its
  expected reach now lands on the most-restricted role"* — which it did not,
  while memory was context-gated alone.

**`application` keeps both, deliberately**, with a test saying why:
`vta-agent-memory` grants exactly that role so *the memory service is not you*,
and every existing install would otherwise stop being able to save.

The capability is checked **before** the context, so a caller missing it cannot
use the reason text to probe which contexts exist.

## Tests

Four new, in the existing slice: reader-can-list-not-write,
monitor-reaches-nothing, application-keeps-full-access (the regression guard for
the plugin), and gate-ordering. `cargo test -p vti-common -p vta-service --lib`
1,439 passing; workspace builds clean; clippy clean.

## Follow-up, not done here

The canonical `Capability` enum in the trust-tasks registry
(`specs/device/_shared/0.1/device-binding.schema.json`) is **already behind this
crate** — it carries neither `sign-trust-task` nor `credential-write`, both of
which shipped here without it. A spec PR reconciling all four values follows
separately. Flagging rather than quietly widening the gap.

### DIFF
--- a/vta-service/src/trust_tasks/memory.rs
+++ b/vta-service/src/trust_tasks/memory.rs
@@ -1,6 +1,18 @@
 //! Agent-memory trust-task slice (`spec/vta/memory/{put,list,delete}/0.1`).
 //!
 //! A per-context key/value store for AI-agent memory. Each handler is:
+//! - **Capability-gated** — read tasks require [`Capability::MemoryRead`],
+//!   mutating tasks require [`Capability::MemoryWrite`]. Until this pair
+//!   existed the only gate was the context, which is binary: any DID that
+//!   could reach a context could also write and delete every memory in it, so
+//!   there was no way to grant an agent read-only access to your own memory.
+//!   The published specification already assumed otherwise —
+//!   `vta/memory/delete/0.1` reasons about "a VTA whose write capability is
+//!   granted more freely than its read capability". Legacy ACL rows carry no
+//!   explicit capability set and fall back to
+//!   [`derived_capabilities_for_role`], which grants both to `admin`,
+//!   `initiator` and `application` (the role a memory agent runs as), read
+//!   only to `reader`, and neither to `monitor`.
 //! - **Context-gated** — the caller must be permitted to act in
 //!   `payload.contextId`, enforced via [`AuthClaims::require_context`], the
 //!   exact ACL gate the context-scoped key tasks use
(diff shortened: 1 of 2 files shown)


stormer78:
Two notes for review, both found while raising the spec-side counterpart
([dtgwg-trust-tasks-tf#344](https://github.com/trustoverip/dtgwg-trust-tasks-tf/pull/344)).

**These two PRs are independent — no local patch or merge order needed.** Nothing
in this workspace consumes the *generated* `Capability` from `trust-tasks-rs`;
`vti_common::acl::Capability` is hand-written and merely documented as matching
`device/_shared/0.1`. So this can land before, after, or without #344.

**And that is exactly how the drift happened.** Nothing enforces agreement
between the hand-written enum and the schema it claims to match, which is why
`sign-trust-task` and `credential-write` reached production without ever
reaching the registry. This PR does not widen that gap (#344 closes all four
values at once), but the mechanism that allowed it is still there.

Worth a follow-up: a small census test — the same shape as
`DISPATCHED_URIS`/`UNSPECCED_DISPATCHED_URIS` in `vtc-service` — asserting that
every `Capability` variant's kebab-case wire form appears in the published
schema enum, and vice versa. It would have caught this class the day it
appeared. Happy to raise it separately rather than growing this PR.

**On the red `semver report`:** expected, and the report working. Two new
variants on a public enum change the public API of `vti-common`, so per
`CLAUDE.md` the fix is not a version bump in this branch — release-plz reads the
same signal and puts the bump in the Release PR.

stormer78:
Correcting my note above on the semver report — it carries **two** failures, and
only one of them is this PR's. A reviewer shouldn't have to work out which.

**1. `enum_variant_added` — this PR, and intended.**

```
variant Capability:MemoryRead  in vti-common/src/acl/mod.rs:153
variant Capability:MemoryWrite in vti-common/src/acl/mod.rs:157
```

Two new variants on an exhaustive public enum. The report is doing its job; per
`CLAUDE.md` the fix is not a version bump in this branch — release-plz reads the
same signal and puts the bump in the Release PR. The practical consequence for
consumers is in the PR description: exhaustive `match` on `Capability` needs one
arm each; wildcard arms are unaffected.

**2. `constructible_struct_adds_field` — pre-existing on `main`, not from here.**

```
field MemoryDeleteBody.ext, MemoryListBody.ext, MemoryPutBody.ext,
      CreateAclBody.ext, ListAclBody.ext, IssueCredentialBody.ext, … (vta-sdk)
```

These are the `ext` members added by #1231, which is the commit this branch is
based on (`a6b52d01`, *"fix(sdk): accept the `ext` member every payload schema
declares"*). This branch touches no `vta-sdk` file — `git diff origin/main...HEAD
-- vta-sdk/` is empty — and the same failure is on `main`. Flagging it here only
so it isn't misread as fallout from this change.

PR #1235, opened 2026-09-01
feat(vta): implement vta/credentials/list, and check the vault/credentials family

stormer78:
Implements `vta/credentials/list/0.1`, specified upstream in [trustoverip/dtgwg-trust-tasks-tf#342](https://github.com/trustoverip/dtgwg-trust-tasks-tf/pull/342) and shipped in `trust-tasks-rs` 0.17.4.

`vta/credentials` served `issue` and `revoke` and nothing else, so an issuer could not ask its own agent what it had issued. `revoke` is keyed on a `credentialId` that `issue` returns exactly once — a caller that did not record it at that moment could not recover it at all.

## A read over records that already exist

`IssuedCredentialRecord` already carries the id, holder, both instants and the revocation instant and reason, and revocation is a **tombstone rather than a delete** — the comment on it says that is precisely so the audit and verifier trail survives. So a revoked credential is still there to list. No new storage.

- **Bodies are never returned.** `summarise` is the one place the projection happens, so "a summary never carries the credential" is enforced in a single spot rather than remembered at each call site. `vault/list/0.1`'s rule — *list enumerates, release uses* — is what this follows.
- **`status` is derived at read time**, `revoked` beating `expired`. A stored status is a copy of a fact about the clock and is wrong one second after it is written; reporting a revoked credential as merely expired would hide that somebody acted.
- **Gated on `require_manage`**, not the Admin-plus-step-up its mutating siblings use. An operator who may read the ACL and the policy set may read what their own agent issued — same category of question — and a step-up firing on every page of a list is one people learn to clear without reading. It is audited anyway: *who enumerated the issuance log* is what an incident review asks, and nothing else would record it.

## The dep bump found two real defects

Bumping `trust-tasks-rs` to 0.17.4 made the eight `vault/credentials/*` URIs *published*. They have been **dispatched since before they had a specification**, so the conformance sweep could not see them — `resolved_uris()` filters dispatched URIs through `schema_for()`. Specifying the family (#338) is what finally let it look, and it found:

**1. `ReceiveBody` serialized `credentialBase64: null`.**

```
vault/credentials/receive/0.1: request is not canonical:
data did not match any variant of untagged enum Payload
  { …, "credentialBase64": null, "format": null, … }
```

`#[serde(default)]` without `skip_serializing_if` leaves an unset member as `null`, and the schema's `oneOf` counts a null member as *present* — so the body matched neither branch. Same defect class as the sibling registry's `payload_null_census`.

**2. `force` was accepted by four verbs that ignore it.**

`CredLifecycleBody` was shared across archive, unarchive, delete, restore and purge, but only `handle_delete` reads `force`. A caller asking for something stronger than the verb it named got the weaker thing **and a success**. `delete` now has its own `CredDeleteBody`; the other four refuse the member, which is what their schemas always said.

Neither breaks anything today — these are request bodies the service parses and never sends — but both would bite the first typed client built on them, and #2 is a silent-wrong-outcome rather than an error.

## Three debt ratchets moved down

Each discharged by specification rather than deletion:

| Ratchet | Change |
|---|---|
| `UNSPECCED_DISPATCHED_URIS` | −8 entries |
| producer-payload census `UNPUBLISHED` | −1 (that payload is now *validated*, not skipped — and it passes) |
| vtc-service bound-URI count | 12 → 4 |

What remains in the last one is the four *secrets*-store lifecycle verbs, which still have no spec — named in the comment so the residue is not mistaken for the whole.

## Tests

`page_rows` is split out of `list_issued` and unit-tested, because the cursor is where a bug hides. It is the last storage key of the previous page and resumption is **strictly after it**, not at an offset — so a credential issued mid-walk cannot shift the window and skip a row nobody has seen. That is a test, not a comment:

```
a_row_issued_mid_walk_does_not_displace_an_unseen_one
a_cursor_resumes_strictly_after_it
a_full_page_reports_truncation_and_the_key_to_resume_after
a_short_page_is_not_truncated_and_offers_no_cursor
```

Plus status precedence, an unreadable expiry reading as **active** rather than expired (guessing "expired" would report a live credential as dead and send someone to reissue something that works), and that a serialized summary contains no credential.

The conformance table gains nine witnesses — the new task, and the eight `vault/credentials/*` that had none.

`IssuedCredentialSummary` is `payload_ext_census`'s first `NO_EXT_BY_DESIGN` entry, with the reason recorded: it is a list row rather than a payload root and its published schema declares no `ext` slot, so adding the field would make this crate emit documents the schema rejects — the inverse of the defect that census catches.

## Checks

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets` (no warnings), `cargo test -p vta-sdk` (18 suites), `cargo test -p vta-service --lib` (1009 passed), `producer_payload_conformance` (5), `vtc-service --test trust_task_manifest` (8). The full-workspace run is slow locally; CI is the arbiter.

## Downstream

This is what lets the browser VTA console replace its "there is no list of issued credentials, and this is not an empty one" notice with a real table. That notice stays until an agent can actually answer — a console claiming a list before one existed would be the failure it was written to avoid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8769,9 +8769,9 @@ dependencies = [
 
 [[package]]
 name = "trust-tasks-rs"
-version = "0.17.3"
+version = "0.17.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "41d841bf16638473fed22f94ce446cadcc932afc60ca3f623eef184d546d945b"
+checksum = "84ea8a0b30f49df3227e7f1adaf52ca2bad7591740059be7a629906a743f8dfa"
 dependencies = [
  "async-trait",
  "chrono",
--- a/vta-sdk/src/protocols/credentials_issuance.rs
+++ b/vta-sdk/src/protocols/credentials_issuance.rs
@@ -1,5 +1,5 @@
 //! Wire payloads for the issued-credential lifecycle Trust Tasks
-//! (`spec/vta/credentials/issue/0.2` and `spec/vta/credentials/revoke/0.1`).
+//! (`spec/vta/credentials/{issue/0.2, revoke/0.1, list/0.1}`).
 //!
 //! These mint / revoke a VTA-signed W3C Verifiable Credential addressed to a
 //! holder DID, distinct from the credential-vault slice (`vault/credentials/*`)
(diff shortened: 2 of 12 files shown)


stormer78:
Flagging the `semver report` failure before a reviewer has to chase it. It is informational and never blocks, but it is reporting something real, and it is a different break from the one on #1231.

**`CredLifecycleBody` lost its `force` field.** That is a removal, not an addition, and removals are the sharper kind: anything reading `body.force` on an archive/unarchive/restore/purge body stops compiling.

That is the point of the change rather than a side effect of it. Only `handle_delete` ever read the field; the other four verbs accepted `force` and silently dropped it, so a caller asking for something stronger than the verb it named got the weaker thing **and a success**. Their published schemas declare no such member. `delete` now has its own `CredDeleteBody` with the field, and the other four refuse it.

The structs are `pub(super)` within `vta-service` — not part of any published crate's API — so the break is contained to this repo. The report is picking up `vta-sdk`'s side of the change, which is purely additive: a new `IssuedCredentialStatus` enum, `ListCredentialsBody`, `IssuedCredentialSummary`, `ListCredentialsResponse`, and the `TASK_VTA_CREDENTIALS_LIST_0_1` constant.

So: additive in the published crate, deliberately breaking in the private one. If the release plan would rather the `force` split landed separately from the new task, it splits cleanly — the two touch different files and neither depends on the other.

---

Separately, on the first CI run here: `Check`, `MSRV`, `Enclave build` and both Docker images failed with `cannot find function handle_list in module credentials`, and they were right. `mod tests` in that file carries two stacked attributes:

```rust
#[cfg(any(test, feature = "test-support"))]
#[cfg(test)]
mod tests {
```

Moving the handler above the test module matched the second, so it landed between the two and inherited the first — compiling only under `test`. I had verified with `cargo check --workspace --all-targets`, which has `test` active and therefore compiled it. For cfg-gating the superset command is strictly *weaker* than the plain one, because it turns on the cfg that hides the mistake. Fixed in 5193a201, and verified against CI's exact `cargo check --workspace` this time.

PR #1237, opened 2026-09-02
feat(rooms): data rooms end to end — storage, dispatch, verification, MLS, and a host

stormer78:
Data rooms, end to end: shared credential-governed spaces for people **and their AI agents**, with the host trusted for availability and nothing else.

Design note: `docs/05-design-notes/data-rooms.md`. Schemas: `dtgwg-trust-tasks-tf` #344 / #346 / #349, released in `trust-tasks-rs` 0.17.7.

## What works

`cargo run -p room-host --example data_room` runs all of it against a real host over HTTP:

- Alice registers a room and writes two records.
- Her **agent** reads them under a chain one link longer, conferring `read` for four hours — and is refused a write.
- Alice **replays her agent's presentation** and is refused: the chain is bound to whoever signed the request.
- A stranger presents a perfectly valid chain from their own room and gets nothing here.
- A **private** room registers and then refuses to be read (see below).
- Alice and Bob form an MLS group, she mints the epoch the membership change produced, seals a record under the exporter key and stores it. Bob fetches it from the host and opens it. The host held every byte, could not read one, and cannot move one — the record fails to open at another key, another version, or another room.

## The shape

**`vti-rooms`** (published) — storage, wire types, and the *shape* half of authorization: presence, depth, membership, the private-tier binding. Depends only on `vti-common`, so a room host can reuse it without a credential library or a DID resolver.

**`vti-rooms-dtg`** (`publish = false`) — the cryptographic half, behind a `ChainVerifier` trait. Two independent checks, and neither substitutes for the other: every credential's proof must verify against the key its `verificationMethod` names, *and* the chain must reach a root the room issued. Doing only the first is the classic mistake — anyone can mint a well-formed VAC naming any scope and any action.

The seam is not squeamishness about dependencies. Verification needs a resolver, and the resolver is a different thing on a VTC than on a standalone host; pinning either choice into `vti-rooms` would make the storage layer un-reusable for the other. It is what lets **one** decision about what is safe to serve be shared by hosts that resolve DIDs differently — and a host that configures no verifier gets `RefusesEverything` and serves nothing, on every tier.

**`room-host`** (`publish = false`) — a host with no roster, no policy engine, no credential issuance and no admin surface. That absence is structural: a host keeping its own record of who belongs would become part of the room's membership, and the room could no longer move without reissuing credentials.

**`vtc-client`** — `RoomSession`, the five room methods (none of which reads `self.token`), and the MLS + sealing layers behind an `mls` feature.

## Two findings, both from tests that failed the first time they ran

**The pooling defence compares the chain's *root*, not its leaf.** The first version compared the leaf and refused every agent — correctly by its own rule, because *an agent is not a member of anything*. Its human is. The root is the grant the room made, so its subject is the member whose standing the chain descends from. Comparing the root admits the agent and still refuses the attack: a chain rooted at Bob cannot be presented with Alice's membership, whoever holds the leaf.

**A presentation is bound to its presenter, not bearer.** `dtg_credentials::authority::verify_chain` takes a `presenter` but uses it only for the `audience` check on links that name one — binding the leaf is the verifier's job, and a verifier assuming otherwise would authorize every captured presentation. Checked twice, in the verifier and again in `authorize`, so the property does not depend on every future implementation remembering it.

## Defects this branch found and fixed

- **`room-host` answered with bare JSON** where `vtc-client` parses a routed Trust Task document — two hosts of the same protocol disagreeing about its wire form, invisible to room-host's own tests because they asserted the shape they sent.
- **`updatedAt` was a unix integer**; the published schema types it `string`/`date-time`. Also `null` for absent optionals under `additionalProperties: false`. Both in `Record::metadata()`, both invisible to serde — there is now a schema-conformance suite for the rooms family, since the VTC's sweep is scoped to `spec/vtc/`.
- **The nitro image** copies a hand-maintained member list; both new crates were missing, so cargo-chef's stub survived and the build died six minutes in on `vti-common = "^0.0.1"`. Fixed, plus a census so the next one fails in a second naming the line to add.
- **Every backup test failed** on `no AppState handle for keyspace 'rooms'` — `BACKED_UP` and `backed_up_handle` are two lists of the same thing with nothing making them agree. Fixed, and pinned in both directions.
- **cargo-deny** on RUSTSEC-2026-0173 — build-time only, unavoidable through OpenMLS's only crypto provider, behind an off-by-default feature. Scoped exception with the drop condition named.

## What is deliberately not done

**A `private` room is refused, and the demo shows the refusal.** Its subject is withheld by design, so the same-subject property has to be proved in zero knowledge — and *which* proof is a profile the DTG cred-spec puts explicitly out of scope. `SubjectBindingVerifier` is the seam, with no implementation shipped. A private room whose pooling defence nobody checked is worse than one that will not open.

**`dtg-credentials` is a pinned git rev.** 0.6.0 is merged and unreleased; that repository had no release path at all until OpenVTC/dtg-credentials#17. It lives in the one crate that is `publish = false`, and swaps to `dtg-credentials = "0.6"` on one line when the `v0.6.0` tag publishes.

**The `rooms/keys/*` oracle is spec'd (#349, released) but not implemented.** It is what lets an agent open a sealed record without ever holding the key, and it belongs in `vta-service` — its own PR, and it wants a decision about whether the MLS and sealing layers move from `vtc-client` into `vti-rooms` first, since a VTA needs them too.


stormer78:
Extended: the open tier is now complete **end to end**, client included.

## Added since the first push

**Dispatch** (`rooms::{wire,authz,handlers}`) — and a correction. I had said this was blocked
on `rooms/*` publishing. It wasn't, and checking rather than repeating the claim is what
unblocked it: the schema gate is a `#[cfg(test)]` conformance sweep scoped to `spec/vtc/`,
and this family is top-level `spec/rooms/`, while `parse_payload` is generic over
`DeserializeOwned` and needs no generated type. Wire types are hand-written in `rooms::wire`
against [dtgwg-trust-tasks-tf#346](https://github.com/trustoverip/dtgwg-trust-tasks-tf/pull/346)
(now merged) — **swap them for the generated ones when the crate publishes, don't keep both**.

**Client** (`vtc-client/src/rooms.rs`) — `create_room`, `put_record`, `get_record`,
`list_records`, `mint_epoch`, and a `RoomSession`.

## The property worth reviewing for

Every other method on `VtcClient` carries an operator token. **A room call carries none.**
`RoomSession` holds credentials, not a session, and `room_task` is the single place a room
call is made so that "never reads `self.token`" is checkable in one function rather than
across five.

On the service side the same invariant has two structural defences: handlers take **no
`AuthClaims`**, and `AuthorizedAction` is constructible only by `authorize()`.

The agent case is modelled by construction rather than a flag — a member's session and their
agent's are built identically and differ only in the chain they carry (depth 1 vs 2). There
is a test pinning that, because if the difference ever becomes a *parameter* on the type
rather than a property of the credentials, the design has drifted.

## Still deliberately refused

Sealed tiers. Chain verification needs `dtg_credentials::authority::verify_chain`
([dtg-credentials#15](https://github.com/OpenVTC/dtg-credentials/pull/15), merged but not yet
on crates.io — the workspace pins `dtg-credentials = "0.4"`). Serving a sealed room before
that is wired means serving one whose chain nobody checked, so it refuses, with a test
pinning the refusal.

**That is the one line to change when 0.6.0 publishes**, and it is the whole of what
`attributed` needs on this side.

`cargo test`: vtc-service **950 lib + 104 integration**, vtc-client **13**. Clippy clean.

stormer78:
Two more layers since the last update. The branch now carries the open tier complete, plus
the two pieces of `attributed` that do not depend on an upstream publish.

## DID templates — `room` and `room-host`

Neither is a new mechanism: provision-integration already mints DIDs from templates, so a
room and a room host are **integration kinds**.

`room` is a data room's own identity — it holds a DID, issues the VIC/VMC/VAC credentials
that govern it, and is messageable. **`did:webvh` rather than `did:peer`, and that is
load-bearing**: transferring a room is a controller change, and `did:peer` encodes its keys
in the identifier so it can never have one.

`room-host` exists so that hosting your own rooms (topology T1) does not mean adding
stranger-facing surface to the VTA — the process guarding the master seed should not also
terminate presentations from arbitrary DIDs.

## MLS group layer (`vtc-client`, `mls` feature)

Off by default, matching how `vtc-service` gates its `bbs` verifier: a caller that only
reads an `open` room carries none of OpenMLS.

The mapping is the design's own shape, which MLS arrived at independently — DTG as
Authentication Service, the host as Delivery Service trusted for availability only, an MLS
epoch as the room epoch the host stores. **One leaf per member (their VTA)**, so devices and
agents hang off it through the oracle model rather than joining the group, which sidesteps
MLS's multi-device complexity.

Storage keys come from the **exporter** under a room-specific label rather than the group's
own message keys — the `draft-sullivan-mls-attachments` pattern after SFrame — so a change
to record sealing cannot weaken the group's messaging, or the reverse.

`epoch_authenticator` is exposed because it is what gets **anchored in the room's witnessed
DID log**. A host acting as Delivery Service can fork a group, and members cannot detect
that by comparing *through* the host. An anchor it cannot forge is what gives each member
something to check against.

**Two tests carry the argument**: two members derive an identical storage key without it
ever crossing the host, and removing a member provably changes that key — forward-only
removal demonstrated rather than asserted.

## What is left, and what it waits on

Wiring `dtg_credentials::authority::verify_chain` into `rooms::authz` — the last thing
`attributed` needs here. [dtg-credentials#15](https://github.com/OpenVTC/dtg-credentials/pull/15)
is merged but crates.io still has 0.5.0 and **that repo has no publish workflow**, so it
needs a manual `cargo publish`. Until then `authz` refuses the sealed tiers, with a test
pinning the refusal.

`cargo test`: vtc-service **950 lib + 104 integration** · vtc-client **13 default / 19 with
`mls`** · vta-sdk **300**. Clippy clean, including with the feature on.

stormer78:
Sealed-tier client integration added — the MLS layer now meets the task surface.

## `SealedRoom`

Pairs a `RoomSession` (the credentials a host authorizes against) with a `RoomGroup` (the
key material a host never sees). Those are two different things, and this type is the only
place they meet: **authorization travels to the host; keys do not.**

Each record's AEAD associated data commits to `roomId|key|version|epoch`, so a host that
relocates a sealed record produces an authentication failure rather than a readable record.
It holds every byte and still cannot move one — the same class of defence
`vti_common::store::encryption` already applies by binding values to their
`(keyspace, key)` location, repeated deliberately because the reasoning was paid for once.

## One wrinkle worth review

**A record's version is assigned by the host, so a writer does not know it when sealing.**
`seal_record` binds the version the writer *intends*, and a caller that lets the host assign
a different one will find the record does not open.

That is the correct failure — accepting whatever came back would mean the binding commits to
nothing — but it does constrain usage to create-only writes (`expected_version: Some(0)`) or
a read-before-rewrite. Both are already supported by the task surface, and the module docs
say so rather than leaving it to be discovered.

## Tests

Six, three of which carry real properties:

- **A relocated record does not open** — checked four ways: another key, another version, a
  relabelled epoch, another room.
- **A non-member cannot open one**, however identical their session looks — a different
  group is a different key.
- **Sealing the same plaintext twice uses a fresh nonce**, because a reused one would break
  ChaCha20-Poly1305 outright.

`cargo test -p vtc-client --features mls`: **25 passing**. Default build unchanged at 13.
Clippy clean both ways.

PR #1249, opened 2026-09-04
fix(tee): make allow_kms_reinit explicit authorization for reinit regardless of KMS failure type

yaroslava-kurash:
## Summary

Make `allow_kms_reinit` the sole authorization for clearing existing TEE bootstrap ciphertexts. `AccessDenied` and all other KMS failures now preserve the existing identity and fail closed unless explicit reinitialization is enabled.

## Motivation

Previously, any KMS `AccessDenied` during a subsequent boot automatically cleared the sealed seed, JWT, and data-key ciphertexts and generated a new VTA identity.

That behavior created a destructive denial of service: launching the wrong EIF could permanently erase the persisted VTA identity, and the correct EIF could no longer recover it. The fix makes this fail closed and preserves the sealed identity material.

## Changes

- Removed the implicit `AccessDenied` reset path.
- Made `allow_kms_reinit` the only authorization for clearing bootstrap ciphertexts.
- Updated KMS configuration and security documentation.
- Preserved the tenant-overlay restriction preventing runtime injection of `allow_*` flags.
- Kept KMS error classification for diagnostics only.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p vta-tee kms_bootstrap --lib`
- `cargo test -p vta-config overlay_rejects_allow_flags_admin_did_and_mode --lib`
- `cargo clippy -p vta-tee -p vta-config --all-targets -- -D warnings`

Manual Nitro validation confirmed:

- `AccessDenied` fails closed without identity reset.
- A temporary KMS denial preserves the existing identity.
- A non-`AccessDenied` proxy transport failure also fails closed, and recovery restores the same DID.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10537,6 +10537,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.11.0",
+ "tempfile",
  "tokio",
  "tokio-vsock",
  "tracing",
--- a/deploy/nitro/README.md
+++ b/deploy/nitro/README.md
@@ -1262,10 +1262,13 @@ The template format follows the did:webvh spec. Examples:
 | `did:webvh:{SCID}:example.com%3A8080:vta` | `https://example.com:8080/vta` |
 
 > **Note on ciphertext deletion:** If an attacker deletes the ciphertext files,
-> the next boot will generate a new identity. This is a denial-of-service (the old
-> identity and data are lost), not a privilege escalation — the attacker still
-> can't authenticate to the new VTA without admin credentials. Back up the
-> ciphertext files and monitor for unexpected identity changes.
+> the next boot takes the first-boot path and mints a new seed — but the TEE
+> integrity manifest is MAC'd with a key derived from the *old* seed, so it no
+> longer verifies and the boot fails closed rather than silently coming up under
+> a new identity. Either way this is a denial-of-service (the old identity and
+> data are unreachable), not a privilege escalation — the attacker still can't
+> authenticate to the VTA without admin credentials. Back up the ciphertext
+> files and monitor for unexpected boot failures.
 
 **The mnemonic is never displayed.** To export it for backup:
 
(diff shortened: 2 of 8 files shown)


stormer78:
Reviewed this and pushed a follow-up commit (1c35f81) onto the branch rather than leaving it as a list of review notes. The core change here is right and closes a real destructive DoS — the notes below are all downstream of it. Happy to drop any of them if you disagree, particularly the third.

### The reinit path didn't produce a bootable VTA

This is the one that mattered. The clear removed the four KMS-protected rows, but `tee:integrity-manifest` lives in the same bootstrap keyspace and is MAC'd with `derive_mac_key(&storage_key)` — and `storage_key` is derived from the seed the reset discards. So first boot minted a new seed, `boot_verify_and_install` loaded the surviving manifest, `stored.mac_valid(&mac_key)` was false, and startup aborted with:

> TEE integrity manifest MAC verification failed — the manifest was tampered with or the storage key changed.

`allow_anchor_init` doesn't help; it only covers the *missing*-manifest arm. Concretely: an operator follows the new error message ("set `tee.kms.allow_kms_reinit = true` for a one-time reset"), reboots, and the identity is destroyed **and** the enclave still refuses to start — with an error pointing at parent-host compromise rather than the reset they just authorized.

The manifest is now cleared alongside the ciphertexts it's sealed against. I pulled the clear out into a named `clear_reinit_state()` with a test over a real store, so a row can't silently drop out of that list again.

This is pre-existing (the old `AccessDenied` auto-clear had the same hole) — it just becomes load-bearing once this PR makes reinit the *only* recovery route.

### An in-place reinit still isn't a full reset

Worth being explicit about, because it changes what the flag is for. The discarded seed is what every other keyspace is encrypted under, and the store is strictly fail-closed on decrypt — so `compute_state()`'s `acl_root` read fails before the manifest check even runs, on any VTA that has written state. `allow_kms_reinit` only yields a bootable VTA on a store holding nothing but the bootstrap rows (e.g. a VTA re-imaged before it ever provisioned an admin). Otherwise the reset is a wipe of the data volume.

I've said so in the log lines, the `TeeKmsConfig` doc comment, `security-model.md` and the Nitro README, and noted that an anchored deployment additionally needs `allow_anchor_init` (plus `allow_unanchored` to re-anchor the counter), since the external counter still holds the old version.

### Transient classes now fail closed even with the flag — please sanity-check this one

This narrows the PR's "regardless of KMS failure type" slightly, so flagging it explicitly. `KMS_INTERNAL` and `NETWORK` mean KMS never answered — the ciphertexts have not been *shown* to be undecryptable, so there's nothing to reset from. Destroying a recoverable identity on "no answer" reopens the same DoS by another route: operator sets the flag for a deliberate reset, reset succeeds, they forget to rebuild the EIF without it, and a later vsock-proxy hiccup on reboot wipes the freshly-minted identity.

It's in the same direction as the PR (the flag is still the sole *authorization*; it's now necessary rather than sufficient), but it is a semantic change to your title, so say the word if you'd rather keep the flag unconditional.

### Test coverage

Grepping for `allow_kms_reinit` found no test anywhere in the workspace, and `mod tests` in this file only covers CMS/DER round-trips and key-string drift — so `cargo test -p vta-tee kms_bootstrap --lib` exercised none of the changed branch. Since the P3.6 plan bullet you rewrote asks for exactly this, I extracted `resolve_reinit(class, allow_kms_reinit) -> ReinitDecision` and asserted every cell of the truth table across all six classes. Plan bullet updated to record what's done and what's left (folding `ciphertexts_present` in so the whole `BootDecision` is one resolver).

### Docs

- **`deploy/nitro/README.md`** wasn't updated by the PR, and it's the file operators actually run. §"Auto-recovery (forgot to update KMS policy)" still told them the VTA would "Clear the old bootstrap data / Do a fresh first boot with a new identity", and §"Fresh deployment" ended with "Or simply deploy without updating the KMS policy — the auto-recovery will handle it". Both now describe the fail-closed behaviour and the real recovery (add the new PCR0 to the policy and boot again — the original DID comes back). The adjacent ciphertext-deletion note was wrong for the same reason and is corrected.

- **`tee-architecture.md`**: "their delete-and-reinit is gated by the explicit `allow_kms_reinit` reset flag" is only true for the *decrypt-failure* path. A parent host that just deletes the three ciphertext rows misses the `if let (Some, Some, Some)` entirely and takes the unguarded first-boot path — no flag consulted (the README's own note describes that attack). What catches it is the manifest: a new seed means a new storage key, hence a new MAC key, so the surviving manifest fails verification. Using the flag to justify excluding the ciphertexts *from* the manifest was circular. Rewritten to separate the two cases — which is also why an authorized reinit has to clear the manifest, so the two halves now explain each other. Also fixed the missing article.

### Validation

`cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, `cargo clippy -p vta-tee -p vta-config --all-targets -- -D warnings`, `cargo test -p vta-tee -p vta-config --lib` (35 + 22 pass).

One dependency added: `tempfile` under a new `[dev-dependencies]` in `vta-tee`, for the store-backed clear test.

PR #1292, opened 2026-09-07
feat: pnm clients self-provision their own mediator ACL (health + key rotation)

abqdr-aff:
Fresh/rotated pnm clients now open their own mediator account automatically, so
DIDComm reply-bearing calls (e.g. `pnm health` trust-ping) work without a manual
mediator ACL edit.

- Health: provision over the live TrustPingSession socket before the VTA ping.
- Rotation: `connect_with_transport` now honours `needs_rotation` for DIDComm
  (previously dead code — DIDComm-only VTAs never rotated); rotated DID's
  mediator account opened on both rotate paths.

Verified live against a TEE VTA. Unit tests added for the ACL-hash convention
and allow-all ACL shape.
[2-before-fix-no-rotation.log](https://github.com/user-attachments/files/31916893/2-before-fix-no-rotation.log)
[3-after-fix-rotation-evidence.log](https://github.com/user-attachments/files/31916894/3-after-fix-rotation-evidence.log)

### DIFF
--- a/pnm-cli/src/commands/health.rs
+++ b/pnm-cli/src/commands/health.rs
@@ -208,6 +208,22 @@ pub(crate) async fn run(
         println!("  {DIM}No transport advertised{RESET}");
     }
 
+    // Re-read the session before anything below uses its key material.
+    //
+    // The `session` above is a snapshot taken at the top of `run`, but the
+    // Authentication section just called `ensure_authenticated`, which rotates a
+    // `needs_rotation` temp did:key: it mints a fresh DID, moves the ACL entry
+    // onto it and *deletes the temp entry at the VTA*. The snapshot's
+    // `client_did` / `private_key_multibase` are the dead temp identity from
+    // that point on.
+    //
+    // Reading a stale identity used to cost only a failing ping. It now costs a
+    // write: the DIDComm probe below provisions an allow-all mediator account
+    // keyed on `sha256(client_did)`, so a stale DID here leaves a permanently
+    // open account for a throwaway credential — exactly the litter
+    // `vta_sdk::acl_setup`'s module docs say the SDK avoids.
+    let session = auth::loaded_session(keyring_key).or(session);
+
     // ── Mediator + DIDComm pings ──────────────────────────────────
     print_section("Mediator");
 
--- a/pnm-cli/src/commands/health.rs
+++ b/pnm-cli/src/commands/health.rs
@@ -294,6 +294,13 @@ pub(crate) async fn run(
                 .await
                 {
                     Ok(Ok(session)) => {
+                        // Open this client's own mediator account (allow-all)
+                        // over the session's live socket before the forwarded
+                        // VTA trust-ping. A freshly bootstrapped or rotated
+                        // client is closed for forwarded delivery, so the VTA's
+                        // pong would otherwise be dropped by the mediator.
+                        session.provision_client_acl("pnm").await;
+
                         // Ping mediator (steady-state: warm-up + measured)
                         match tokio::time::timeout(
                             std::time::Duration::from_secs(20),
--- a/vta-sdk/src/acl_setup.rs
+++ b/vta-sdk/src/acl_setup.rs
@@ -110,16 +110,8 @@ async fn set_client_acl_internal(
     .await
     .map_err(|e| format!("failed to create ATM profile: {e}"))?;
 
-    // Hash the client's DID for the mediator's ACL record (self-reference).
-    // SHA-256 hex to match the mediator's account-key convention
-    // (`sha256::digest(did)` in affinidi-messaging-sdk).
-    let mut hasher = Sha256::new();
-    hasher.update(client_did);
-    let hash_bytes = hasher.finalize();
-    let client_did_hash = hash_bytes
-        .iter()
-        .map(|b| format!("{:02x}", b))
-        .collect::<String>();
+    // SHA-256 hex of the DID — the mediator's account-key convention.
+    let client_did_hash = client_acl_hash(client_did);
 
     // Build an "allow all" ACL that accepts every message type. Fields left
     // `None` (e.g. the self-manage flags) keep the mediator's existing value.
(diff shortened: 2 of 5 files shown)


marat-affinidi (approved):

stormer78:
## Review — changes needed before merge

Reviewed `6099600d` against `origin/main`. The core idea is right: a client opening its own mediator account over its live socket is the correct shape, and `set_client_acl_with_profile` reusing the caller's profile is the right primitive. Two ordering problems on the rotation path need fixing before this is safe to merge, plus a handful of smaller items.

Verified independently: the `client_acl_hash` test vector is correct (`shasum -a 256` matches the hard-coded hex), and `allow_all_acl_opens_everything_but_blocked` does assert the `*Forwarded` flags rather than just round-tripping — a good guard against the wire-casing drift class.

---

### 1. Blocking — mediator I/O sits between the irreversible ACL swap and the local key write

`vta-sdk/src/session.rs:1234` (REST) and `:1101` (DIDComm).

In `rotate_key`, by the time we reach

```rust
if let Ok(Some(mediator_did)) = resolve_mediator_did(&session_vta_did).await {
    provision_rotated_client_acl(&new_did, &new_private_key, &mediator_did).await;
}
```

the VTA has already swapped the temp DID's ACL entry onto `new_did` and deleted the temp entry — but the caller (`ensure_authenticated`, `session.rs:499`) has **not yet** run `self.save_session(key, &session)`. So `new_private_key` exists only in memory.

Both `resolve_mediator_did` (a network DID resolution) and `provision_rotated_client_acl` → `TrustPingSession::new` (TDK init + ATM build + `profile_add` + `profile_enable_websocket`) carry no timeout at any layer, unlike every other mediator connect in this file, which goes through `connect_didcomm_bounded`.

Failure path: the VTA advertises a mediator that is down or black-holing. An operator runs `pnm --transport rest acl list` right after `pnm setup`. Rotation swaps the key server-side, then the process hangs indefinitely inside `TrustPingSession::new`. Ctrl-C — or a CI job timeout — kills it before `save_session`. The temp DID is gone from the VTA ACL and the new key was never persisted: the operator is locked out with no recoverable credential.

`rotate_key_didcomm:1101` has the identical window — the call sits after `probe.delete_acl(&session.client_did)` but before `ensure_authenticated_didcomm` reaches `self.save_session(key, &rotated)` at `:601`.

This is R2.1 (Remote-First): "process dies on the next line — then what?" And the work is explicitly documented as best-effort ("a failure only means the account opens lazily on the next connect"), so it has no business being on the critical path at all.

**Fix:** move the provisioning out of `rotate_key` / `rotate_key_didcomm` and into the callers, *after* the session is persisted. `ensure_authenticated` and `ensure_authenticated_didcomm` both already have the mediator DID in scope at that point. Belt-and-braces, wrap the call in a `tokio::time::timeout` regardless — nothing best-effort should ever be able to hang a CLI.

### 2. Blocking — the rotation delegation drops the DIDComm connect deadline

`vta-sdk/src/session.rs:929`.

The new `if session.needs_rotation { return self.ensure_authenticated_didcomm(key).await; }` is a good catch, but it routes around `connect_didcomm_bounded`. `ensure_authenticated_didcomm` calls `crate::client::VtaClient::connect_didcomm` directly, twice (`:580` and `:605`), with no timeout.

`connect_didcomm_bounded`'s own doc comment says it exists so "an unreachable mediator surfaces as an error naming the recovery flag rather than as an indefinite hang". So the moment a session is flagged `needs_rotation`, `pnm <cmd>` against a dead mediator regresses from "times out in 30s with the `--transport rest` hint" to "hangs forever with no output" — exactly what `DIDCOMM_CONNECT_TIMEOUT_DEFAULT` was introduced to prevent.

**Fix:** switch both call sites in `ensure_authenticated_didcomm` to `connect_didcomm_bounded`.

---

### 3. `pnm health` provisions an allow-all account for an already-revoked DID

`pnm-cli/src/commands/health.rs:302`.

`let session = auth::loaded_session(keyring_key)` at `health.rs:23` is a snapshot taken **before** the Authentication section calls `auth::ensure_authenticated` at `:170` — which is what performs the `needs_rotation` key roll. On a pending-rotation session, `info.client_did` / `info.private_key_multibase` at `:290-293` are therefore the *pre-rotation temp DID*, whose VTA ACL entry `rotate_key` has just deleted.

Before this PR that only meant a failing ping — harmless diagnostic noise. Now it is a **write**: an allow-all `account/update` gets persisted on the mediator keyed on `sha256(temp_did)`, leaving a permanent open account for a dead throwaway DID. That is precisely the outcome `acl_setup.rs`'s module doc says the SDK deliberately avoids ("litter the mediator with allow-all entries for throwaway probe DIDs").

**Fix:** re-load the session after the Authentication section, or skip `provision_client_acl` when the loaded session had `needs_rotation` set.

### 4. "DIDComm-only VTAs never rotated" is only half fixed

The `needs_rotation` check was added only to the `VtaEndpoint::DIDComm` arm. Two sibling paths in the same function still connect with the temp key and return without rotating:

- **`session.rs:749-761`** — an explicit `mediator_did_hint` (`[vta] mediator_did` in pnm/cnm config) short-circuits before DID resolution and calls `connect_didcomm_bounded` directly.
- **`session.rs:790-806` and `:880-891`** — the `VtaEndpoint::Tsp` arm. A VTA advertising *both* TSP and DIDComm never reaches the `DIDComm` arm at all, so both `Auto` and a forced `--transport didcomm` still ignore `needs_rotation` there.

Dual-transport is the direction the workspace is explicitly moving (`docs/05-design-notes/tsp-enablement.md`), so on those VTAs the session stays flagged `needs_rotation` forever and the temp `did:key` is never retired — the same latent state this PR is fixing, just relocated.

**Fix:** hoist the `needs_rotation` check to the top of `connect_with_transport`, before the endpoint match, rather than repeating it per arm.

### 5. `provision_rotated_client_acl` opens a third socket, defeating the helper it calls

`vta-sdk/src/session.rs:1145`.

`set_client_acl_with_profile`'s doc justifies itself as "reuses the caller's live profile/socket — so no second websocket contends for the DID's one-socket-per-DID slot". But `provision_rotated_client_acl` builds a brand-new exclusive `SessionHub` + websocket for `new_did`.

In `rotate_key_didcomm` that gives three open/close cycles on the same DID's single mediator slot in quick succession: `probe` connect → `probe.shutdown()` → `TrustPingSession` connect → shutdown → the caller's `connect_didcomm` at `:605`. Because these are separate ATMs, `SessionHub.attached` cannot detect an incompletely torn-down predecessor, so a slow mediator teardown surfaces as a `duplicate-channel` eviction of the final, authoritative client.

**Fix:** call `set_client_acl_with_profile` on the **probe's** live profile before `probe.shutdown()`. That removes the extra cycle entirely and matches the helper's stated design.

### 6. Redundant field, and a `dead_code` warning in one feature combination

`vta-sdk/src/session.rs:2107`. `TrustPingSession.client_did` duplicates `self.identity.did`, which `AttachedIdentity` (`session_hub.rs:316`) already carries.

`acl-setup` is an optional feature requiring `session` (`vta-sdk/Cargo.toml:192`), and `provision_client_acl` is the field's only reader — so `cargo check -p vta-sdk --features session` (without `acl-setup`) warns `field 'client_did' is never read`. Reading `self.identity.did` removes both the duplication and the warning.

### 7. Two comments make contradictory claims about reply delivery

`set_client_acl_internal` (`acl_setup.rs:128-131`) states the `account_update` response "cannot arrive until this very ACL grants `receive_forwarded` — so an `Err` here (typically a timeout) does NOT mean the request was dropped".

`set_client_acl_with_profile` (`acl_setup.rs:195`) asserts the opposite: "The `account_update` reply rides the same live socket (direct delivery, not forwarded), so it returns even while the account is still closed."

Only one can hold for a given mediator. If the older comment is right for an `ExplicitAllow` mediator whose account is still closed, every first-time call burns the full `ACL_TIMEOUT` — a guaranteed silent 10s stall added to `pnm health` and both rotation paths, with nothing printed to explain the pause.

**Fix:** confirm against the mediator's actual delivery path and delete whichever comment is wrong. If the pessimistic case is reachable, `pnm health` should print a line for it rather than stalling mutely.

---

### Not blocking, noted for the record

`pnm health` overwriting a deliberately-restrictive per-account ACL with allow-all is a behaviour change on what is otherwise a read-only command — but it matches what `set_client_acl_on_connection` already does on every DIDComm connect, so it is not new policy from this PR.

---

Items 1 and 2 share one root cause and one fix: nothing unbounded should run between the remote ACL mutation and the local key write. Happy to push these directly if you tick **Allow edits by maintainers** on the PR — the branch is on your fork with that off, so I can't currently write to it.

stormer78:
Rebased onto latest `main` (was 24 commits behind) and pushed `a113d821` addressing all seven items. Your commit is untouched apart from the rebase.

| # | Item | Fix |
|---|---|---|
| 1 | Mediator I/O inside the rotation commit window | REST: moved out of `rotate_key` into `ensure_authenticated`, **after** `save_session`, under a 25s deadline. DIDComm: moved to **before** `delete_acl`, while the temp entry is still authoritative and a failure costs nothing |
| 2 | Rotation delegation dropped the connect deadline | Both connects in `ensure_authenticated_didcomm` now go through `connect_didcomm_bounded` |
| 3 | `pnm health` writing an ACL for a revoked DID | Re-reads the session after the Authentication section, before anything uses its key material |
| 4 | Only the `DIDComm` arm rotated | One gate hoisted above every mediator connect, gated on `didcomm_mediator_advertised` so REST-only and TSP-only VTAs fall through unchanged |
| 5 | Third socket for `new_did` | Provisions over the probe's live socket via the new `VtaClient::provision_client_acl` — which is what `set_client_acl_with_profile` was written for |
| 6 | Redundant `client_did` field | Dropped; reads `self.identity.did`. `cargo check -p vta-sdk --features session` is warning-free again |
| 7 | Contradictory reply-delivery comments | The older one was wrong. The mediator checks `Capability::ReceiveForwarded` only in `routing.rs`, against the *next hop's* ACL — a direct reply on the requesting socket isn't gated by it, so no `ACL_TIMEOUT` stall. Comment corrected with the reference |

On item 1, the DIDComm ordering is worth a second look since it's the non-obvious half: step 3 already creates the ACL entry for `new_did`, so from that point the new DID is live and the temp is only *also* live. Provisioning between the probe connect and `delete_acl` therefore sits entirely outside the danger window — crash there and the temp is still authoritative, so the caller just retries. Doing it after `delete_acl` was what put optional, unbounded work on the one path that can't tolerate it.

**On testing.** The invariant is *where an await sits* relative to `save_session` and `delete_acl`, spread across three functions with nothing type-level holding it — no return value observes it, and reproducing it live needs a mediator that hangs rather than refuses. So it's pinned with four source-level guards over this file's own `include_str!`. That's an unusual shape, so each was mutation-checked rather than trusted:

| Mutation | Test that failed |
|---|---|
| Move DIDComm provisioning back after `delete_acl` | `didcomm_rotation_opens_the_mediator_account_before_deleting_the_temp_entry` |
| Call `open_rotated_mediator_account` inside `rotate_key` | `rest_rotation_opens_the_mediator_account_only_after_the_key_is_saved` |
| Restore the unbounded `VtaClient::connect_didcomm` | `didcomm_rotation_path_bounds_its_connects` |
| Move the rotation gate back into the `DIDComm` arm | `every_didcomm_capable_transport_path_rotates` |

Each mutation failed exactly one guard and no others — so they're load-bearing, not vacuous. `cargo check --workspace --all-targets` and clippy are clean, and the guard against a second session in `rotate_key_didcomm` will fail the build if a `TrustPingSession` reappears there.

Nothing here changes the feature you built — it's the same self-provisioning, moved off the one path where blocking is unsafe.

PR #1300, opened 2026-09-07
fix(rooms)!: a room stays readable across a membership change

stormer78:
Adding or removing one member made **every record already in the room fail to open** — for every member, including the one who wrote it. `vti-rooms` 0.1.1 is on crates.io with this behaviour.

## The defect

Records are sealed under the MLS exporter for the epoch current when they were written. `SealedRoom::open_record` used the group's **current** key regardless of the record's own epoch. Every membership change is a commit and every commit advances the epoch, so the first add or remove silently destroyed read access to the room's entire contents.

It surfaced as `DidNotOpen` — "sealed under a different key, epoch, or location" — which reads like corruption or a hostile host, not like a key the caller no longer has.

Nothing caught it because **every test in the crate sealed and opened inside one epoch**, and the end-to-end demo never advanced one either. `tests/epoch_backfill.rs` is the regression, and it fails on the parent commit in both directions.

Found while comparing the design against [Encrypted Spaces](https://encryptedspaces.org/) (Orrù–Perrin–Trapp–Zaverucha, Aug 2026), whose retention system exists precisely to solve this.

## The fix: an epoch key chain

At each commit the committer seals the outgoing epoch's storage key under the incoming one — an `EpochLink`. A holder of the current key walks the chain backwards to any retained epoch.

```
  epoch 4 key ──opens──▶ link(4) ──yields──▶ epoch 3 key
                                                 │
                                      ──opens──▶ link(3) ──yields──▶ epoch 2 key …
```

**Backwards only**, and that asymmetry is the whole design: a member holding epoch 2's key derives nothing at epoch 3, so removal stays forward-only and a removed member still reads nothing written after them. That third property is what stops the first two being fixed the easy and wrong way, and it has its own test.

This is the upper tier of the Encrypted Spaces retention system (§4.2 — *"each rekey produces a new group key that encrypts its predecessor"*), linearised: we have no directory hierarchy for a lower tier to align to.

Links are ciphertext, so a host stores them and learns only that an epoch happened — which `Room.epoch` already told it.

## What it costs, and why it is a policy not a default

Post-compromise security **for record content**: once the chain exists, a compromised current key reaches every retained epoch. MLS's own PCS is untouched — a compromised leaf still heals at the next commit — but the storage layer above it no longer inherits that healing.

That is the trade a **library** makes and a message stream does not, so it is `RetentionPolicy::{Chained, FromJoin}`, fixed at creation beside `visibility`. Deserialising defaults to `FromJoin` because a room stored before this change genuinely has no links, and no policy field can conjure keys that were never sealed.

## Cryptographic deletion falls out of it

`storage::prune_epoch_links_before` severs the chain below an epoch. Unlike `purge_record`, which asks a host to erase bytes and trusts it against backups and snapshots, this destroys the only path from a key any member holds to the keys that material was sealed under.

## What this does NOT finish

A **newly joined member still reads only from their joining epoch.** Their Welcome carries the current epoch and nothing below it, and the links reach them over a task that does not exist yet. The crate supports the backfill (`SealedRoom::{links, add_links}`, exercised in tests and in the demo); the wire does not.

Same for choosing `FromJoin` at creation and for a prune verb. All three need `dtgwg-trust-tasks-tf` first, because every `rooms/*` request schema is `additionalProperties: false` — the same constraint that keeps `retention_days` a per-host constant. Design note §12.2 has the table; the chain-fetch task is the one that matters.

## Breaking

- `Room` gains `retention_policy`
- `MembershipChange` gains `link`; `RoomGroup::apply_commit` returns it
- **`SealedRoom::group_mut` is gone** — it let a caller advance the epoch behind the chain's back, reintroducing this defect one call site at a time. Replaced by `add_member` / `remove_member` / `apply_commit` on `SealedRoom`
- `SealedRoom::open_record` takes `&mut self` (memoises the walk)
- `RoomGroupRecord` gains `links`

## Verification

- `cargo test --workspace --all-features` — **170 suites, 5528 tests, 0 failures**
- `cargo clippy --workspace --all-features --all-targets` — clean
- `cargo run -p room-host --example data_room` — step 13 now adds a third member through a real host and shows Alice, Bob and a joining Carol reading a record sealed before the change

Docs: design note §5.5, §11 (prior art), §12.2 (remaining wire work), §14.7–8 (verified reads; rooms created before the chain). Operator guide §2 and the commits section.

> Note for CI: `vta-service --test mock_vta` stack-overflows in debug builds independently of this change — `services_write_paths_against_a_hosted_vta_did` passes under `RUST_MIN_STACK=16777216`. Pre-existing, not introduced here.


stormer78:
### Semver report triage

Red, and it should be — but only some of it is this PR. Recording the split so nobody re-derives it, and **no version is being bumped here** (release-plz reads the same signal and puts the bump in the Release PR).

**From this change, all documented in the PR body:**

| Crate | Finding |
|---|---|
| `vti-rooms` | `field MembershipChange.link` |
| `vti-rooms` | `field Room.retention_policy` |
| `vti-rooms` | `inherent_method_missing` — `SealedRoom::group_mut` removed |
| `vti-rooms` | `method_receiver_ref_became_mut` — `SealedRoom::open_record` takes `&mut self` |
| `vta-service` | `field RoomGroupRecord.links` |

**Already on `main`, not introduced here** — the baseline is the published crate, so unreleased breaks from earlier merges show up on whichever PR runs next:

| Crate | Finding | From |
|---|---|---|
| `vti-rooms` | `field Room.mirror_of` | #1289 (read mirrors) |
| `vta-persona` | `field Finding.shared_with_profile_count` | persona work |
| `vti-secrets` | `Gcp/Azure/Aws/K8sSeedStore` no longer `UnwindSafe`/`RefUnwindSafe` | #1290 (seed-read cache — an added lock does this) |

The `vti-secrets` one is the only entry I'd give a second look independently of this PR: losing `UnwindSafe` is the kind of break that is invisible until a downstream `catch_unwind` stops compiling. It is correct and expected for a cached store, just worth being deliberate about in the release notes.

Every other check is green: Test (workspace), Test (VTC), Feature combos, Clippy, Format, MSRV, cargo-deny, DCO, commit lint, Enclave build.

PR #1306, opened 2026-09-07
fix(persona): give the disclosure step-up context a type an approver can render

stormer78:
A defect in #1304, found by reading the approver side rather than the agent
side.

`initiate_disclosure_step_up` builds an authorization context so the approver
can see **what would leave** — the verifier, the claim types, the purpose. It
emitted a flat bag:

```json
{ "operation": "persona/disclosure/present", "previewId": "…",
  "verifierDid": "…", "claimTypes": ["payment.card"], "purpose": "checkout" }
```

That is not the shape an approver renders. `vta-mobile-core`'s
`parse_step_up_request` surfaces the context as raw JSON "for the native layer
to decode and render as the approval card", and the card discriminates on a
**`type`** URI — the Cierge share ask
(`https://openvtc.org/cierge/authorization-context/0.1`) is the only exemplar
and carries `{type, summary, risk, action}`.

So a context with no `type` is a card the approver cannot choose. The member
that exists precisely so the holder is not answering "approve a disclosure?"
blind would have shown them nothing.

## The fix

```json
{ "type": "https://openvtc.org/persona/authorization-context/0.1",
  "summary": "Approve disclosing 1 fact to did:key:z…",
  "risk": "high",
  "action": { "kind": "disclose", "previewId": "…", "verifierDid": "…",
              "claimTypes": ["payment.card"], "purpose": "checkout" } }
```

- **`summary` is the same string as `reason`, deliberately.**
  `reason_and_context` reads `summary` back out as the reason for any context
  that carries one, so two different sentences would put two accounts of one
  act in a single signed document.
- **`risk: "high"`** is the claim-type registry's judgement restated, not a
  fresh one: everything reaching this gate is a type the registry marked
  `release: stepUp`.
- The specifics move under `action` keyed by `kind`, so one renderer serves
  every context.

`vta-mobile-core` needs no change — it passes the context through untouched,
and its test pins the Cierge shape, not this one.

## Cuts over with the plugin

OpenVTC/vta-browser-plugin#190 reads the new shape, and adds the check this
change makes possible: **a context whose `type` is not the disclosure one is
refused**, so a share ask travelling under the same `ext` key can never be
shown to a holder in a disclosure's words. Per that repo's "nothing is
deployed — do not write compatibility folds" rule, the two land together and
neither carries a dual-accept arm.

## Tests

The end-to-end assertion in `persona_trust_task.rs` now pins the `type`, the
`risk`, and **that `summary` equals the request's `reason`** — the last being
the one a future edit is most likely to break by rewording one of them. The
existing assertion that the card number never reaches the approver's device is
unchanged and still passes.

`persona_trust_task` 22/22; `cargo clippy --workspace --all-targets --locked
-- -D warnings` clean.

## Guide checklist (§9)

- [x] R1.2/R1.3/R1.4/R1.5/R1.6 — no client, lock, retry, loop or ack touched
- [x] R2.1 — no mutation added
- [x] **R3.* — a wire shape changed**, and both consumers cut over in the same
      cycle: the plugin in #189, `vta-mobile-core` unaffected (pass-through).
      camelCase throughout; the `type` URI is the discriminator
- [x] R5.* — a context with no recognised `type` is refused by the consumer
      rather than guessed at
- [x] R6.* — `risk` restates the registry rather than asserting a fresh judgement
- [x] Deviations — none

### DIFF
--- a/vta-service/src/trust_tasks/step_up.rs
+++ b/vta-service/src/trust_tasks/step_up.rs
@@ -569,6 +569,14 @@ const DEFAULT_STEP_UP_REASON: &str = "this operation requires a stepped-up (AAL2
 /// the same key.
 const EXT_KEY_AUTHZ_CONTEXT: &str = "org.openvtc.authorization-context";
 
+/// `type` of the authorization context a `release: stepUp` disclosure carries.
+///
+/// The context is passed through to the approver's device untouched, and the
+/// card that renders it discriminates on this member — the Cierge share ask
+/// (`https://openvtc.org/cierge/authorization-context/0.1`) is the exemplar.
+/// A context without one is a card the approver cannot choose.
+const PERSONA_AUTHZ_CONTEXT_TYPE: &str = "https://openvtc.org/persona/authorization-context/0.1";
+
 /// Pick the reason string + optional structured authorization context from a
 /// gated request's payload. A request MAY carry a `payload.authorizationContext`
 /// (e.g. a Cierge share/spend/tool ask); when it does, its human `summary`
@@ -1087,15 +1095,35 @@ pub(super) async fn initiate_disclosure_step_up(
             n => format!("{n} facts"),
         }
     );
-    let mut context = json!({
-        "operation": "persona/disclosure/present",
+    // The shape an approver's card already knows how to render:
+    // `{type, summary, risk, action}`, with the specifics under `action` keyed
+    // by `kind`. The first version of this emitted a flat bag of members with
+    // no `type`, which a native layer that discriminates on `type` cannot pick
+    // a card for — so the context that exists to show the approver what would
+    // leave would have shown them nothing.
+    //
+    // `summary` is the same string as `reason` on purpose, not by coincidence:
+    // `reason_and_context` reads `summary` back out as the reason for any
+    // context that carries one, so two different sentences here would put two
+    // different accounts of the same act in one document.
+    let mut action = json!({
+        "kind": "disclose",
         "previewId": preview_id,
         "verifierDid": verifier_did,
         "claimTypes": claim_types,
     });
     if let Some(p) = purpose {
-        context["purpose"] = json!(p);
+        action["purpose"] = json!(p);
     }
+    let context = json!({
+        "type": PERSONA_AUTHZ_CONTEXT_TYPE,
+        "summary": reason,
+        // Every disclosure that reaches this gate is one the claim-type
+        // registry marked `release: stepUp`. That is what `high` means here —
+        // it is the registry's judgement restated, not a fresh one.
+        "risk": "high",
+        "action": action,
+    });
 
     match mint_pending_step_up(
         &state.sessions_ks,
--- a/vta-service/tests/persona_trust_task.rs
+++ b/vta-service/tests/persona_trust_task.rs
@@ -1830,11 +1830,24 @@ async fn a_step_up_claim_needs_an_approval_bound_to_that_preview() {
     // second device, and the approver authorises a release rather than reading
     // one.
     let context = &approve_request["payload"]["ext"]["org.openvtc.authorization-context"];
-    assert_eq!(context["operation"], "persona/disclosure/present", "{body}");
-    assert_eq!(context["previewId"], preview_id, "{body}");
-    assert_eq!(context["verifierDid"], "did:key:z6MkVerifier", "{body}");
-    assert_eq!(context["claimTypes"][0], "payment.card", "{body}");
-    assert_eq!(context["purpose"], "checkout", "{body}");
+    // The `{type, summary, risk, action}` shape an approver's card renders. The
+    // `type` is the half a native layer discriminates on, so a context without
+    // one is a context that shows the approver nothing.
+    assert_eq!(
+        context["type"], "https://openvtc.org/persona/authorization-context/0.1",
+        "the context carries no type, so no approval card can be chosen for it: {body}"
+    );
+    assert_eq!(context["risk"], "high", "{body}");
+    assert_eq!(
+        context["summary"], approve_request["payload"]["reason"],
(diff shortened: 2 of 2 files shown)


stormer78:
The commit message references `vta-browser-plugin#189`; the paired PR is **#190**. The PR body is corrected — leaving the commit as-is rather than force-pushing over a green run for a number.

stormer78:
Thanks. I've gone through each of these against the code. **One produced a real fix — in the paired PR, not here** (OpenVTC/vta-browser-plugin#190, commit 4a31167: the consumer now checks `action.kind`, not only the context `type`). The rest I'm declining, and several rest on a factual error worth correcting because it recurs across four of the findings.

## The recurring error: the context *is* cryptographically bound

STRIDE-1 and STRIDE-5, and the "previewId carried as plaintext, unsigned field" finding, all assume the authorization context travels unsigned or unbound. It does not.

`mint_pending_step_up` builds the whole document — `type`, `summary`, `risk`, `action`, and every field inside it — and then signs it:

```rust
let proof = DataIntegrityProof::sign(
    &doc, secret,
    SignOptions::new()
        .with_proof_purpose("assertionMethod")
        .with_cryptosuite(CryptoSuite::EddsaJcs2022),
).await
```

The approver verifies that proof and requires the signer to be an enrolled executor **before rendering anything** — `vta-mobile-core`'s `parse_step_up_request`, and `verifyStepUpApproveRequest` on the wallet side. The spec rule is explicit that a consumer verifies the proof *before* surfacing the reason.

So STRIDE-1's mitigation "introduce integrity checks (e.g. HMAC over the full context object) so a downstream verifier can detect tampering between context construction and card rendering" describes a control that already exists and is stronger than an HMAC — an asymmetric signature over the whole document, verified against an enrolment list. The plugin test `a tampered context does not survive the proof` demonstrates it: appending one claim type after signing fails verification.

## STRIDE-5 / "previewId ... no visible cryptographic binding to approval" — declined

The recommendation is "ensure `preview_id` is bound server-side to the exact disclosure request ... verified at approval time (not solely trusted from client-echoed payload)". **That is exactly the design, and it is the requirement the spec imposes** (`persona/_shared/0.1/CONVENTIONS.md` §5: *"The approval is the maintainer's state, not a member of the request"*).

At approval time the handler does not read a `previewId` from the request at all:

```rust
let pending = consume_pending_step_up(&state.sessions_ks, &challenge, now_epoch()).await …
if let Some(preview_id) = pending.bound_to.as_deref() { … }
```

`bound_to` is server-side state in the agent's own keyspace, written at mint time. The request carries only the `challenge` (256 bits of entropy, two UUIDv4s, base64url), which is the lookup key. **A producer cannot name which preview gets approved.** The copy in `action.previewId` is display material for the card, and the wallet cross-checks it against the refusal's own copy (#190) precisely so a mismatch is caught rather than trusted.

## STRIDE-1's attack scenario — declined on inputs

Step 1 posits an attacker who "controls or influences `claim_types`, `verifier_did`, `purpose`, or `preview_id`". Those are read from the **stored `Preview` record** — `preview.claims`, `preview.verifier_did`, `preview.purpose` — fetched from the agent's keyspace by `peek_preview`. They were fixed when the preview was minted. The only request-supplied value on this path is the `previewId` used to fetch that record.

Also declining the sub-recommendation to "validate that `claim_types` match the registry's `release: stepUp` set before minting". The list is deliberately *every* claim in the preview, not the gated subset: the approver should see everything that would leave, not only the field that triggered the gate. Narrowing it would make the card understate the disclosure.

## "Untyped/unbounded `purpose`" — declined

`purpose` is bounded at **512 characters** by `persona/disclosure/preview/1.0`'s payload schema, enforced by `validate_payload` at the dispatch spine before any handler runs. It is not unbounded.

It is verifier-supplied, and that is its purpose — it is displayed as *the verifier's stated reason*, not as anything the agent vouches for. Render-time hardening (control characters, bidi/homoglyphs, line breaks) is a real concern and belongs in the approver's card, where the text is drawn; it is not something the agent can do on the card's behalf without destroying the content it is meant to convey.

## "Hardcoded Secrets: Generic Api Key" — false positives

The three hits are `did:key:z6MkNeverBound`, `did:key:z6MkPersonaKnows` and a `previewId` literal in test code. They are base58 `did:key` test fixtures, which is what the "generic api key" heuristic is matching on. **None of them is on a line this PR touches** — 501 and 660 predate it; my diff is 18 insertions elsewhere in the file.

## "Breaking wire-format change without versioning/negotiation" — declined, but worth stating

The discriminator **is** the version: `https://openvtc.org/persona/authorization-context/0.1`. A future shape becomes `/0.2` and #190's consumer refuses anything it does not recognise rather than guessing.

There is deliberately no negotiation or compatibility fold. The browser plugin has never been published and has no users outside this workspace; its `CLAUDE.md` records this as a standing rule — *"nothing is deployed — do not write compatibility folds"* — because a dual-accept arm here would be dead code that reads like a live constraint. The two repos cut over together, which is why #1306 and #190 are paired.

## STRIDE-4, namespace collision — addressed in #190

The `type` is namespaced by producer (`openvtc.org/persona/…` vs `openvtc.org/cierge/…`), and as of #190 the consumer refuses a context whose `type` is not the disclosure one. A correctly-signed Cierge share ask from an enrolled agent passes every other guard in that module — proof, issuer, tamper — and only the `type` check separates it. There is a test that drives exactly that.

## STRIDE-2, reason/summary desync — already pinned

`summary` and `reason` are the same string by construction, and the integration test asserts `context["summary"] == approve_request["payload"]["reason"]` so a reword of one without the other fails CI. That was added in this PR.

---

If any of the above is wrong I'd genuinely like to know — particularly the first section. If `payload.ext` were excluded from the proof anywhere in the stack, that would be a serious bug and would change my answer on several of these.

PR #1310, opened 2026-09-08
feat(persona): honour a holder's release override at disclosure

stormer78:
Closes the known limitation stated in #1304: only the claim-type registry's
default `release` was enforced, and a holder's per-attribute override was
ignored.

## The comment that said not to build this yet

`Attribute` carried a deliberate absence:

> There is deliberately no `release` counterpart. The published schema defines
> one, and nothing in this workspace enforces a `stepUp` at disclosure yet; a
> stored override for a gate that does not exist is a promise the holder would
> be entitled to rely on.

#1304 built the gate. The precondition that comment named is satisfied, so the
field can be stored honestly — and this PR is that comment being cashed in
rather than a new idea.

## Copies go down; nothing reads up

The override is set on the **pool** attribute, above the boundary. A preview is
built from the **materialised copy** inside a context, which deliberately
carries no pool identifier and cannot read the pool to ask what the holder
decided.

So the decision travels *with* the projection or not at all:

```
Attribute.release  →  ResolvedClaim.release  →  MaterialisedClaim.release  →  the gate
   (pool)              (profile resolve)          (stored in the binding)
```

Re-materialisation on an attribute edit already exists — "edit once,
everywhere" — so changing the override changes what every bound context
enforces without any of them reading anything.

Inline profile entries and context-local profiles carry `None`: they have no
pool record, so there is nowhere a decision could have been recorded, and the
registry default answers.

## Where the resolution is stored, and why that differs from `sensitivity`

`sensitivity` is derived at every read — *store the override, derive the
default* — so tightening the registry protects attributes already written.
`release` is resolved **once, at preview creation**, onto `Preview.step_up_required`.

The reason is the wire. `persona/disclosure/preview`'s response schema declares
`additionalProperties: false` on its claims, so putting `release` on
`PreviewClaim` would put it on the wire — and `PreviewClaim` *is* serialised
straight into the response. (This is the same trap that caught `elevated` in
the #1304 branch, and the response-conformance layer catches it either way.)
`Preview` itself is not serialised wholesale — the handler picks members out —
so the decision lives there instead.

Freezing it for a preview's lifetime is also the honest reading: a preview is a
snapshot of what the holder was shown, single-use, expiring in minutes. A
holder who changes the override afterwards previews again.

## An override wins in both directions

Following `CLAIM-TYPES.md` §4 rule 1 — *"If the holder set the value explicitly
on the attribute, that wins"* — and matching `sensitivity_of`, which already
argues the case: an agent that kept gating a value its owner had decided needed
no gate would be overruling the person the control exists to serve.

**Loosening is available, and the reason that is not a hole is where the write
happens.** `release` is set through `persona/attribute/put`, which is
holder-scoped — refused to every context-scoped caller and every application
inside a context. A verifier cannot reach it; neither can the site asking for
the disclosure. The only party who can relax the gate is the one it protects.

The audit row records the decision either way, and says so in the trail, because
a holder relaxing their own card from `stepUp` to `consent` is the most
consequential thing this task can do and "attribute updated" would not let them
find the moment the gate came off.

**Deliberately not built:** making the *loosening* itself require a step-up, so
turning a gate off is gated. It is a good idea and it is a new gate, not this
one.

## Tests

`vta-persona` (2 new, 88 total):
- **an override reaches the gate in both directions** — tighten `name.display`
  (which the registry does *not* gate, so a pass cannot come from the default)
  and loosen `payment.card`;
- **a claim not being sent does not gate the one that is** — a profile holding
  a card and a name, previewed for the name only, is a `consent` disclosure.
  The gate is filtered by `requestedClaims` exactly as the preview is.

`vta-service` (1 new, 23 total), end to end through the real dispatch spine:
`attribute/put` accepts `release`, stores it, carries it across the boundary at
bind time, and `present` refuses with `stepUpRequired` — four layers the unit
tests cannot see between. Uses `name.display` so the registry default cannot
account for it, and asserts the same type *without* an override is **not**
gated, so a gate that fired on everything would fail.

`cargo clippy --workspace --all-targets --locked -- -D warnings` clean;
`vta-service --lib` 1045 pass.

## Guide checklist (§9)

- [x] R1.2/R1.3/R1.4/R1.5/R1.6 — no client, lock, retry, loop or ack touched
- [x] R2.1 — the override is written by the existing `put` path; re-materialisation is the existing edit-propagation path
- [x] R3.* — **no wire change**: `MaterialisedClaim` and `Preview` are at-rest types; `release` on `attribute/put` is an already-published optional member
- [x] R5.* — absent override ⇒ registry default; a binding or preview written before these fields deserialises to exactly its previous behaviour (`#[serde(default)]` on both)
- [x] R6.* — the audit row states the holder's decision rather than only that a write happened
- [x] "Process dies on the next line" — no new multi-step mutation
- [x] Deviations — the stored-vs-derived difference from `sensitivity` is explained above and in the field's own docs

### DIFF
--- a/vta-persona/src/binding.rs
+++ b/vta-persona/src/binding.rs
@@ -44,6 +44,19 @@ pub struct MaterialisedClaim {
     /// short, and MUST NOT be disclosed.
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub stale: bool,
+    /// The holder's `release` override, copied down with the value.
+    ///
+    /// **This is the whole mechanism for honouring an override below the
+    /// boundary.** A context cannot read the pool to ask what the holder
+    /// decided, so the decision travels with the projection or it does not
+    /// travel at all — copies go down, nothing reads up. Re-materialisation on
+    /// an attribute edit refreshes it, so changing the override changes what
+    /// every bound context enforces without any of them reading anything.
+    ///
+    /// `None` on a binding written before this field existed, which resolves to
+    /// the registry default — the behaviour those bindings already had.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub release: Option<crate::ReleaseRequirement>,
 }
 
 /// The binding plus the claims it pushed into the context.
@@ -172,6 +185,7 @@ impl PersonaStore {
                 value: c.value,
                 provenance: c.provenance,
                 stale: c.stale,
+                release: c.release,
             })
             .collect())
     }
--- a/vta-persona/src/claim_types.rs
+++ b/vta-persona/src/claim_types.rs
@@ -486,6 +486,45 @@ pub fn sensitivity_of(attribute: &Attribute) -> Sensitivity {
         .unwrap_or_else(|| defaults_for(&attribute.r#type).sensitivity)
 }
 
+/// What it takes to let one attribute leave — §4 per the `release` axis.
+///
+/// Same shape as [`sensitivity_of`], and an override wins in **both**
+/// directions for the same reason: the holder is the principal, and an agent
+/// that kept gating a value its owner had decided needed no gate would be
+/// overruling the person the control exists to serve.
+///
+/// That the loosening direction is available is not a hole, because of where
+/// the write happens. `release` is set through `persona/attribute/put`, which
+/// is **holder-scoped** — above the boundary, refused to every context-scoped
+/// caller and to every application inside a context. A verifier cannot reach
+/// it, and neither can the site asking for the disclosure. The only party who
+/// can relax the gate is the one it protects.
+///
+/// Worth considering separately: whether *loosening* should itself require a
+/// step-up, so the act of turning a gate off is gated. That is a real idea and
+/// deliberately not built here — it is a new gate, not this one.
+#[must_use]
+pub fn release_of(attribute: &Attribute) -> ReleaseRequirement {
+    attribute
+        .release
+        .unwrap_or_else(|| defaults_for(&attribute.r#type).release)
+}
+
+/// The `release` of a claim that has already been pushed into a context.
+///
+/// The pool is not readable from here — that is the boundary — so the holder's
+/// override has to have travelled *down* with the value when the binding was
+/// written. `None` means the projection carries no decision and the registry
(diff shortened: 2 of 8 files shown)


stormer78:
**The red `Feature combos` here is pre-existing on `main` and not from this PR.**

Failing test: `vta-sdk/tests/session_store.rs::ensure_authenticated_runs_full_rotation_flow`, in the `cargo test -p vta-sdk --all-features` step.

```
called `Result::unwrap()` on an `Err` value: "rotate: acl/swap-key failed: not found:
unknown error — has your admin run `vta import-did --did did:key:z6Mkkgmu… --role admin` yet?"
```

This PR touches `vta-persona` and `vta-service`'s persona slice. It does not touch `vta-sdk`.

## Reproduced against unmodified main

Ran the job's exact command in a worktree whose **only** diff from `main` is a shell script and a workflow file — zero Rust:

```
$ git diff origin/main --stat -- '*.rs'      # empty
$ cargo test -p vta-sdk --all-features --test session_store
test ensure_authenticated_runs_full_rotation_flow ... FAILED
test result: FAILED. 23 passed; 1 failed
```

## Where it came from

#1309 (`feat(sdk): one key rotation for every transport, TSP included`), which is this branch's base. **Its own `Feature combos` failed too** — [run 34191224870](https://github.com/OpenVTC/verifiable-trust-infrastructure/actions/runs/34191224870/job/101949526357) — and it merged anyway, so every PR opened since inherits the red.

## The implementation is right; the test is stale

Deliberate, and #1309 says so in the code:

> The swap rides `acl/swap-key` like every other transport's does, so it needs a client — **not a hand-rolled `POST /acl/swap`**.

The test still mocks `POST /acl/swap`, so the request now goes to the Trust-Task dispatch path, finds no mock, and 404s. Nothing is wrong with rotation — it is the fixture that was not carried over.

**Worth not fixing casually.** The replacement mock has to answer a *signed Trust-Task envelope* rather than a plain JSON body, and a mock that is merely permissive enough to go green would leave REST key rotation with a test that passes without exercising it. Right now that path has no passing coverage at all, which is the actual cost here.

I have not touched it — it belongs with #1309. Happy to take it if that is useful; say the word.

PR #1311, opened 2026-09-08
ci: keep authorization-context type constants unique and named

stormer78:
The one recommendation worth keeping from the security review on #1306/#190:
a check that authorization-context `type` constants stay unique.

## Why this is a real property and not tidiness

An approver's device chooses which approval card to render by discriminating on
the context's `type` URI. Two producers sharing one means an approver reads a
prompt describing an act that is not the one they are authorising.

The **consumer** side is already guarded: the browser plugin refuses a context
whose `type` is not the disclosure one, and a test drives a correctly-signed
Cierge share ask — from an enrolled agent, passing proof, issuer and tamper
checks — through the disclosure path and requires it to be refused. This is the
producer side.

It cannot stop a collision *between repositories*, which no single repository
can. It stops the version of the mistake that happens here.

## Two checks

1. **No two constants declare the same URI.**
2. **No production source inlines the URI pattern outside a constant.** This is
   the one that matters: a literal inside a `json!` is invisible to check 1, so
   a second producer added that way would collide silently — precisely the case
   the guard exists for. #1306 added exactly such a literal before it became
   `PERSONA_AUTHZ_CONTEXT_TYPE`.

## Verified non-vacuous, and the first attempt was wrong

Both branches were probed rather than assumed, and the first probe of check 1
**passed when it should have failed** — I appended the colliding constant to the
end of the file, which lands after the `#[cfg(test)]` module the scanner
truncates at. Re-probed with the constant in a realistic position and it fires.
Worth recording, because a guard that looks green against a bad probe is
indistinguishable from one that works.

```
PROBE 1  colliding constant, before the test module  → ::error:: … both declare … EXIT=1
PROBE 2  URI inlined in production code              → ::error:: … outside a constant … EXIT=1
restored                                             → EXIT=0
```

(The exit codes were also checked directly rather than through a pipe, which had
been masking them behind `head`.)

## Test code is exempt, and that had to be handled

Two inline `#[cfg(test)]` fixtures already carry the Cierge type —
`vta-mobile-core/src/task.rs` and `vta-service/src/trust_tasks/step_up.rs` — so
without an exemption the guard would have failed on day one against code that is
correct. A fixture asserting the wire value is the point of a fixture.

The scanner cuts each file at its first module-level `#[cfg(test)]`, which
relies on the Rust convention that the test module comes last. A file that put
one in the middle would have the rest of it skipped, so **the number of lines
actually scanned is printed on success** — a file that suddenly scans almost
nothing is visible rather than silently unguarded.

Comments are stripped before anything is matched, so a doc comment naming
another producer's type as an example is documentation rather than a
declaration. The script's own header would otherwise trip its own guard.

## Current state

```
OK: 1 authorization-context type constant(s) across 869 production file(s)
    (264909 lines scanned), each declared once and named:
  PERSONA_AUTHZ_CONTEXT_TYPE = https://openvtc.org/persona/authorization-context/0.1
```

Runs as its own job — cheap and independent of the build, so a naming mistake is
reported in seconds rather than behind the compile.

## Guide checklist (§9)

- [x] R1.*/R2.1 — CI only; no runtime code, no client, no mutation
- [x] R3.* — no wire change
- [x] R5.* — the guard fails closed: an unparseable or unexpected shape is an error, not a pass
- [x] R6.* — reports what was actually scanned, so a weakened scan is visible
- [x] Deviations — none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -602,6 +602,24 @@ jobs:
       - name: Lockfile self-pin guard
         run: bash scripts/check-lockfile-self-pins.sh
 
+  authz-context-types:
+    # An approver's device chooses which approval card to render by
+    # discriminating on the authorization context's `type` URI. Two producers
+    # sharing one means an approver reads a prompt describing an act that is not
+    # the one they are authorising — so each must be unique, and each must be a
+    # named constant rather than a literal inside a `json!`, because a literal
+    # is invisible to the uniqueness check.
+    #
+    # Cheap and independent of the build, so it runs on its own rather than
+    # behind the compile jobs: a naming mistake should be reported in seconds.
+    name: authorization-context types
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v7
+      - name: Authorization-context type uniqueness guard
+        run: bash scripts/check-authz-context-types.sh
+
   lockfile-resolves:
     # The sibling of the job above, and the same misattribution failure it
     # describes — arrived at from the other direction. There, Cargo.lock pins a
--- a/scripts/check-authz-context-types.sh
+++ b/scripts/check-authz-context-types.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+#
+# Every authorization-context `type` URI this workspace emits must be declared
+# exactly once, as a named constant.
+#
+# ## What this is protecting
+#
+# The context under `payload.ext["org.openvtc.authorization-context"]` is what
+# an approver's device renders as the approval card, and the card is chosen by
+# discriminating on `type`. Two producers sharing a `type` therefore means one
+# of them gets the other's card: the approver reads a prompt describing an act
+# that is not the one they are authorising.
+#
+# The consumer side of that is already guarded — the browser plugin refuses a
+# context whose `type` is not the one it expects, and there is a test that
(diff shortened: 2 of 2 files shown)


stormer78:
**Same pre-existing red as #1310, and this PR is the proof.**

`Feature combos` fails on the identical test — `vta-sdk/tests/session_store.rs::ensure_authenticated_runs_full_rotation_flow`, same `acl/swap-key … not found` panic, same `23 passed; 1 failed`.

**This branch changes no Rust at all:**

```
 .github/workflows/ci.yml             |  18 ++++++
 scripts/check-authz-context-types.sh | 104 +++++++++++++++++++++++++++++++++++
 2 files changed, 122 insertions(+)
```

A workflow file and a shell script cannot break a Rust test. The break is `main`'s, from #1309 — whose own `Feature combos` also failed before it merged. Full analysis in [#1310's comment](https://github.com/OpenVTC/verifiable-trust-infrastructure/pull/1310#issuecomment-5580301672): the implementation is correct and deliberate, the test's mock is the stale half.

Every other check here is green (14 pass), including the new guard job, which reports:

```
OK: 1 authorization-context type constant(s) across 869 production file(s)
    (264909 lines scanned), each declared once and named:
  PERSONA_AUTHZ_CONTEXT_TYPE = https://openvtc.org/persona/authorization-context/0.1
```

PR #1334, opened 2026-09-09
fix(vtc): sign success responses — 265 specs require it and none got one

stormer78:
Found while working out verified reads (#1333). That note described an unsigned put acknowledgement as a rooms problem. It is not a rooms problem.

## The gap

SPEC §7.3 item 7:

> *Where a specification declares no requirement for the **response**, the **request**'s applies to it, so an omission can never weaken a variant.*

**265 published specifications declare a single `proofRequirement: REQUIRED`.** Only 10 use the per-variant form that could relax it. So 265 task types require a proof on their response — and this service attached one to none of them. `success_response` built the document with `respond_with` and serialised it; nothing signed anything on the way out.

## Why nothing caught it

**No consumer verifies a response proof either** — not `vta-sdk`, not the browser client. Producers not signing and consumers not checking is a mutually consistent silence, and it is how this survived:

- the conformance ledger (checks hand-typed example documents),
- `test_support::response_conformance` (validates every observed body against its published schema — but the **payload's shape**; the proof is on the envelope),
- 35 integration binaries.

Each was looking at something adjacent to it.

## What it cost

No answer this community has ever given is evidence of anything. A member cannot show a third party what the host told them, and cannot be contradicted when they misreport it. For rooms specifically that is the whole basis of a write receipt — but it is true of every join decision, every VMC issuance, every credential the VTC has ever acknowledged.

## Signed at the spine

There are twenty-one `success_response` call sites. Signing is the same decision taken twenty-one times, and forgetting once is a task whose response is unattributable with nothing downstream to notice — which is exactly how every response came to be unsigned. The spine is where the framework's inbound checks already live; the outbound one belongs beside them.

**Success responses only.** An error response's `type` resolves to `trust-task-error`, whose own requirement is RECOMMENDED rather than REQUIRED, and whose variant §7.3 makes undeclarable by a task. Signing those is a separate decision with its own rationale — a retained compliance refusal is the case that argues for it — and is not smuggled in here.

**A community with no signer answers unsigned rather than failing.** It has nothing to sign with, and refusing would make an unprovisioned VTC unusable rather than merely unattributable — the wrong direction to err in, since the operation itself succeeded.

## Verification

- `cargo test -p vtc-service --lib` — 956 passed, 0 failed, plus the two new guards.
- **Guard proven non-vacuous**: removing the `sign_success_response` call makes `a_success_response_is_signed` fail, printing the real unsigned challenge response it now catches.
- `cargo clippy --all-targets`, `cargo fmt --all` — clean.
- The 35 integration binaries were still running locally when I opened this; CI is the authoritative check on them and I have not claimed a result I did not see.

## The VTA half is owed, and is not here

`vta-service` has the identical gap and needs a decision this one did not.

`load_vta_issuer_secret` writes an **audit entry per key access**, so signing every response through it would drown the audit trail — and the audit trail is a security control, so that is a real cost rather than a performance note. The resident path (`secrets_resolver` + `signing_vm_id`) exists and avoids it, but is `#[cfg(any(feature = "didcomm", feature = "tsp"))]`, and **conformance must not depend on feature flags**.

So the fix there is a non-gated response signer populated at init. Worth its own PR rather than bolted onto this one.

### DIFF
--- a/vtc-service/src/trust_tasks/mod.rs
+++ b/vtc-service/src/trust_tasks/mod.rs
@@ -134,9 +134,90 @@ pub(crate) async fn dispatch_trust_task_core(
         return reject_with(&doc, reason);
     }
 
-    // 3. Dispatch by type URI.
+    // 3. Dispatch by type URI, then sign what comes back.
     let type_uri = doc.type_uri.to_string();
-    match type_uri.as_str() {
+    let outcome = dispatch_typed(state, ctx, doc, &type_uri).await;
+    sign_success_response(state, outcome).await
+}
+
+/// Attach this community's Data-Integrity proof to a success response.
+///
+/// # Why here and not at each `success_response`
+///
+/// Signing is the same decision taken once per handler, and there are twenty-one
+/// of them. Forgetting once is a task whose response is unattributable, and
+/// nothing downstream would notice — which is exactly how every response came to
+/// be unsigned in the first place. The spine is where the framework's inbound
+/// checks already live; the outbound one belongs beside them.
+///
+/// # Why this was missing, and what it cost
+///
+/// SPEC §7.3 item 7: where a specification declares no separate requirement for
+/// the *response*, the request's applies to it — "an omission can never weaken a
+/// variant". 265 published specifications declare a single
+/// `proofRequirement: REQUIRED`, so their responses require a proof, and this
+/// service attached none to any of them.
+///
+/// Nothing went red because no consumer verifies one either. Producers not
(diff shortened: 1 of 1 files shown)


stormer78:
Local integration results, which I could not see when opening this — the background run's output had not flushed, so the PR body deferred them to CI.

They completed clean: **45 suites, 435 integration tests, 0 failures**, on top of the 956 lib tests. So signing every success response breaks none of the 35 integration binaries, and the deferral in the description can be read as answered rather than outstanding.

PR #1345, opened 2026-09-09
feat(rooms): a `host` feature, so a room's member half reaches wasm

stormer78:
`vti-rooms` already had two halves and no name for them:

| Consumer | Imports |
|---|---|
| `vta-service` (a member) | `mls`, `sealed`, `wire` |
| `vtc-service`, `room-host`, `vti-rooms-dtg` (hosts) | `storage`, `authz`, `audit`, `wire` |

This names the second half `host`, on by default, and makes `vti-common`
optional behind it.

`vti-common` describes itself as server-side infrastructure and pulls axum,
fjall and tokio — it was the whole of what kept a room's member code off
`wasm32-unknown-unknown`. With it optional, `--no-default-features --features
mls` builds there, which is what lets a browser hold a room's keys itself rather
than ask an agent to. Design: #1344.

**The member half needed no source changes.** That is a property of `error.rs`,
which deliberately does not use `vti_common::error::AppError` because key
material is not a service's concern — a decision made for its own reasons that
paid for itself here.

`lifecycle` is deliberately left ungated. Only hosts consume it today, but it
needs nothing from `vti-common`, and a member computing a room's lifecycle state
to explain it on screen should not need a host's dependencies.

## The wasm plumbing, and why it is per-target

Three things, each of which reports as a `compile_error!` that reads like an
unsupported target and is not:

- **OpenMLS 0.9 has a first-class `js` feature** — `web-time` for the
  `SystemTime` wasm lacks, plus getrandom's JS backend. Declared per-target
  rather than as a feature of this crate, so building for wasm just works with
  nothing for a consumer to remember. Verified wasm-only: `web-time` is in the
  wasm dependency tree and absent from the native one.
- **Two getrandom majors are in the graph.** The 0.4 is ours; the 0.2 arrives
  transitively under `openmls_rust_crypto` through RustCrypto's elliptic-curve
  stack, so no feature declared here could reach it. `getrandom_02` is a feature
  shim, not a dependency this crate calls — done here rather than left to each
  artifact crate because the choice is target-specific and on wasm32 there is
  only one right answer.
- **No `RUSTFLAGS` are needed.** With both declared per-target with their
  features, `--cfg getrandom_backend="wasm_js"` is unnecessary.

## Guards

Two steps in CI's `features` job, so neither property rots silently:

- clippy on the member half with no host (`-D warnings`);
- a wasm32 `--lib` check.

## Note for the version bump

Possibly breaking for an external consumer that already sets
`default-features = false` on `vti-rooms 0.1.x`: they previously got
`storage`/`authz`/`audit` and now get the member half. Consumers on default
features are unaffected — `default = ["host"]` gives them exactly what
`default = []` did. cargo-semver-checks builds with default features, so it will
not see this; flagging it rather than touching a `version =` field.

## Verified locally

- `cargo test -p vti-rooms` — 81 passed
- `cargo clippy -p vti-rooms --no-default-features --features mls --all-targets -- -D warnings` — clean
- `cargo check -p vti-rooms --lib --no-default-features --features mls --target wasm32-unknown-unknown` — clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -781,6 +781,22 @@ jobs:
         run: cargo check -p vta-service --no-default-features --features azure-secrets,keyring,rest
       - name: vta-service tsp transport
         run: cargo check -p vta-service --features tsp
+      # A room's member half must not need a server. `vti-common` is optional
+      # behind `host`, and it is what pulls axum, fjall and tokio — so this is
+      # the check that `mls`/`sealed`/`retention` have not quietly acquired a
+      # dependency on one. It is also the *whole* precondition for the browser
+      # member: no store, no `AppError`, no runtime.
+      - name: vti-rooms member half (no host)
+        run: cargo clippy -p vti-rooms --no-default-features --features mls --all-targets -- -D warnings
+      # …and that the member half still reaches wasm, which is the property the
+      # feature exists for. `--lib` because the dev-dependencies (tokio) are
+      # native and do not need to reach wasm for this to hold.
+      - name: vti-rooms member half on wasm32
+        run: |
+          rustup target add wasm32-unknown-unknown
+          cargo check -p vti-rooms --lib --no-default-features --features mls \
+            --target wasm32-unknown-unknown
+
       - name: vta-sdk attest-verify
         run: cargo check -p vta-sdk --features attest-verify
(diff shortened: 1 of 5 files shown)


stormer78:
Folded into #1347. The feature gate is only useful with the crate that needs it, and reviewing them apart meant reviewing a precondition without the thing it is a precondition for.

PR #1356, opened 2026-09-09
fix(rooms)!: a presentation is bound to its presenter — and dtg-credentials 0.6 → 0.9.1

stormer78:
Three commits, one story: `audience` never worked, the fix for it is upstream, and taking
the upstream fix lets two hand-written checks go.

**Absorbs #1358**, which was a separate PR until it turned out to be the last commit of
this one.

## 1 · The CLI could never present anything

`pnm-cli rooms --host-did …` was refused as `WrongAudience` every time; omitting the flag
skipped the check entirely. Broken and unprotected, nothing in between.

`audience` was never the destination. `verify_chain` compared it to the **presenter** —
"the leaf must be presentable by whoever is presenting it" — and `vta-cli-common` passed
the *host's* DID, naming a party no presenter can ever match.

## 2 · So stop sending it, and stop sending `nonce` too

`room_oracle::present` loses both parameters. The leaf grants to the DID this VTA
authenticated, and a host refuses a chain whose leaf grants to anyone but the party that
signed what it received — so a presentation is usable by its caller and nobody else, and a
caller cannot ask for one made out to a third party because it never gets to name the
subject.

Binding a presentation to a *host* is not offered, because it would not mean anything: the
chain is scoped to the room, a room may have more than one host, and moving between them
without reissuing is the point. Binding a **request** to its destination is the `recipient`
member's job on the document carrying it, which its `proof` covers.

`nonce` goes for a sharper reason: a presentation is a bundle of credentials signed by
their *issuers*, never by the presenter, so a challenge placed inside one is
unauthenticated — an attacker replaying a captured presentation copies the challenge along
with everything else. Freshness belongs to the signed request.

## 3 · dtg-credentials 0.6 → 0.8, which makes all of that the library's rule

0.8 **requires** the presenter to be the leaf's subject (`NotThePresenter`) and **removes**
`audience` outright, citing trustoverip/dtgwg-trust-tasks-tf#414 — the issue this thread
raised. 0.7 additionally made a VAC's `validUntil` required.

**Both hand-written copies of the presenter rule are deleted**, which is the payoff.
`vti-rooms-dtg`'s chain verifier and its nomination check had each written it
independently, after hitting the gap separately:

> `verify_chain` uses `claimant` only for the audience check; it does not require the grant
> to name them. Without this a nomination would be a bearer token, and anyone who ever
> observed one could take the room the moment it went dormant.

Two copies of a rule is one place for it to be forgotten. The tests that pinned the
property stay exactly where they were — a chain presented by the wrong party is still
refused, and which layer refuses it is no longer this crate's business. Their assertions
move to the library's wording, which names both parties rather than only the rule.

`rooms/owner/issue-authority` now **refuses** a request with no `validUntil` rather than
defaulting: nothing about a subject's standing is consulted when a chain is verified, so
authority that does not expire is authority nobody can withdraw by waiting — and choosing a
lifetime here would put the room's most consequential number somewhere its owner never
looks.

## Ordering, for anyone else upgrading

OpenVTC/dtg-credentials#22. A **new issuer against an old verifier** reports a broken chain
rather than a version skew — `names parent zQm…, but was presented after urn:uuid:…`, with
nothing to say those are different kinds of identifier. That is the direction a client-first
upgrade hits, and it is how this started: a browser-native room member on 0.7 talking to
`room-host` on 0.6.

## Verified

`vti-rooms` 106+3, `vti-rooms-dtg` 19, `room-host` 23+13, `vtc-service` rooms 26,
`vta-service` rooms 11 — all green. Clippy clean on the touched crates.

`Test (workspace)` failures on this branch are the #1341 fallout inherited from `main`, not
from these changes; #1359 fixes the last of them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3495,9 +3495,9 @@ dependencies = [
 
 [[package]]
 name = "dtg-credentials"
-version = "0.6.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f292eb5ba50d7f69dd9671e87b4d6ceddbeb973cbf0c05715220bee8446ff90"
+checksum = "fa51b03528a22ad07cd0f812a99fae834b7867c205943a40ad90bb8eb62edb08"
 dependencies = [
  "affinidi-data-integrity",
  "affinidi-secrets-resolver",
@@ -3506,7 +3506,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_json_canonicalizer",
- "sha2 0.10.9",
+ "sha2 0.11.0",
  "thiserror 2.0.20",
  "tracing",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -165,7 +165,7 @@ affinidi-messaging-core = "0.1"
 futures-util = "0.3"
 
 # Decentralized Trust Graph Credentials
-dtg-credentials = "0.6"
+dtg-credentials = "0.9.1"
 
 # JSON Schema validation (issue-time credentialSchema checks)
 jsonschema = "0.49"
--- a/pnm-cli/src/cli.rs
+++ b/pnm-cli/src/cli.rs
@@ -338,8 +338,11 @@ pub(crate) enum RoomCommands {
         /// Base URL of the host storing the room.
         #[arg(long)]
         host: String,
-        /// The host's DID. Omit and the minted presentation is not bound to a
-        /// host — usable by anyone who observes it until it expires.
+        /// The host's DID — the recipient the request document names.
+        ///
+        /// It does not affect how the presentation is bound. That is always to
+        /// the DID you authenticate as, so a captured one is worthless to
+        /// whoever captured it whether this is given or not.
         #[arg(long)]
         host_did: Option<String>,
         /// Only records whose key starts with this.
--- a/room-host/examples/data_room.rs
+++ b/room-host/examples/data_room.rs
@@ -592,7 +592,7 @@ async fn act_four(
     )
     .await?;
 
-    let nomination = g.nominate(&g.successor.did, Some(24 * 365)).await;
+    let nomination = g.nominate(&g.successor.did, 24 * 365).await;
     note("the room issued this nomination to its successor long ago, granting `succeed` —");
     note("a word no room task accepts, so it confers nothing at all while Alice is present");
 
@@ -668,7 +668,7 @@ async fn act_four(
         .claim_owner(
             &successor_session(&h),
             // A nomination the room really did issue — naming somebody else.
-            &h.nominate(&h.agent.did, Some(24)).await,
+            &h.nominate(&h.agent.did, 24).await,
             None,
             &h.successor.did,
             &h.successor.secret_multibase,
@@ -687,7 +687,7 @@ async fn act_four(
     let claimed = client
         .claim_owner(
             &successor_session(&h),
-            &h.nominate(&h.successor.did, Some(24 * 365)).await,
+            &h.nominate(&h.successor.did, 24 * 365).await,
             Some("the owner has been unreachable since March"),
             &h.successor.did,
             &h.successor.secret_multibase,
--- a/room-host/src/lib.rs
+++ b/room-host/src/lib.rs
@@ -1684,7 +1684,7 @@ mod tests {
             ROOMS_OWNER_CLAIM_TYPE,
             serde_json::json!({
                 "roomId": f.room.room_id,
-                "nomination": f.nominate(&f.successor.did, Some(24)).await,
+                "nomination": f.nominate(&f.successor.did, 24).await,
                 "presentation": f.as_successor(),
                 "reason": "the owner has been unreachable since March",
             }),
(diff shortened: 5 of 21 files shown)


stormer78:
**Upstream has answered this, and more decisively than either option above.**

OpenVTC/dtg-credentials#21 — `feat!: a VAC is not a bearer credential; remove audience` —
shipped in 0.8.0 today, citing trustoverip/dtgwg-trust-tasks-tf#414. Neither side of the
conflict won:

- `verify_chain` now **requires the leaf to grant to `presenter`**, refusing otherwise with
  a new `AuthorityError::NotThePresenter`. That is dtgwg-cred-spec#41's normative rule.
- With that in place `audience` can only name the subject (adding nothing) or somebody else
  (satisfiable by nobody), so it is **removed** — from `AuthorityGrant`, from `attenuate`,
  and from the verifier along with `WrongAudience`.
- The destination question, which the registry's wording was reaching for, is answered one
  layer up by the trust task document's `recipient` — which its `proof` already covers.

The changelog also notes that both consumers had independently hand-written the presenter
rule after hitting the gap, `vti-rooms-dtg`'s chain verifier and its nomination check, and
that two copies of a check is one place for it to be forgotten.

**This PR is still right for `main` today**, because VTI pins `dtg-credentials = "0.6"` and
under 0.6 the CLI's presentations are refused every time. It becomes a no-op the moment the
workspace reaches 0.8 — `attenuate` loses the parameter entirely, so it will not compile
until the call is updated anyway.

Suggest merging as-is and tracking the bump separately: it is two breaking releases (0.7
parent digests, 0.8 the presenter rule) and it lets `vti-rooms-dtg` delete both hand-written
re-checks. Ordering caveat filed as OpenVTC/dtg-credentials#22.

stormer78:
Rebased on `main` (post-#1359, so the inherited `Test (workspace)` failures should clear) and moved to **0.9.1** rather than 0.8.

0.9.1 applies the same rule to the VDC — `delegation::verify_chain` gains a `presenter` and refuses with `NotTheDelegate`. This workspace does not use delegation, so it is a drop-in with no code change here.

**Pinned `0.9.1`, not `0.9`, deliberately.** That release is API-breaking despite the patch version, and Cargo treats the two as compatible — so `"0.9"` picks it up on a routine `cargo update`. Naming the floor states which behaviour this workspace requires rather than tolerating either.

It also ships an **Upgrading** section saying verifiers move before issuers, which closes OpenVTC/dtg-credentials#22 — the issue this work raised after a browser member on 0.7 was refused by a host on 0.6 with an error that blamed the chain.

stormer78:
Rebased onto `main` with #1347 in it, which **brought `vti-rooms-wasm` into scope**: it takes the workspace pin, so its `present` no longer compiles with an `audience` argument the library has removed.

Two changes there, both small:

- `attenuate` loses the argument.
- its doc block, which said *"there is no `audience` parameter, and that is deliberate"*, becomes *"there is no longer a field either"*. The reasoning it gave held — a browser member could only ever correctly name itself, so offering the choice was offering a way to get it wrong — and 0.8 settled it by removing the field and requiring the presenter to be the leaf's subject. The binding that method relies on is the library's rule now rather than a value it declines to expose.
- its captured-presentation test moves from expecting `WrongAudience` to `NotThePresenter`. Same property, enforced a layer down.

Worth flagging for whoever merges: had this gone in before #1347, that crate would have been written against the new API from the start. Landing them in the other order means this PR carries the fixup — which is fine, but it is the ordering cost the upstream Upgrading section is about, in miniature.

stormer78:
CI failure fixed, and it was a real one rather than fallout.

**Working Draft 02 renamed `credentialSubject.digest` to `digestMultibase` and changed its
encoding** from `sha256:<hex>` to a base58btc multihash. `vtc-service` hand-rolled the old
definition and did not follow — so with 0.9.1 in, a member using a current client
acknowledged a grant and was told it acknowledged a different one.

The tests that caught it were written for exactly this, and say so:

> Two implementations, one definition — this is the assertion that catches either side
> drifting.

## One definition, one implementation

The new form now delegates to `dtg_credentials::digest_multibase_json` rather than being
copied here — the same lesson as the two hand-written presenter checks this PR already
deletes.

Worth flagging the trap: `credentials::ingress` **already has** a `digest_multibase`, and it
is the same *encoding* over different *coverage* — it digests the document whole, `proof`
included. Its own doc comment says "do not substitute one for the other". Reaching for it
here compiles and produces a plausible string that matches nothing.

## Both spellings are read, each checked its own way

A member whose client predates WD02 still sends the old form and their acknowledgement must
keep verifying, so whichever property arrives is compared against the expected value
computed the same way. **This widens what is accepted, never what counts as a match.** A new
`DigestForm` keeps the pairing in one place, because a check comparing a WD02 digest against
a WD01 expectation refuses a *correct* acknowledgement while telling the member they
acknowledged a different grant.

Multibase digests are compared as decoded bytes, not strings — one multihash has more than
one spelling.

## Two tests for a branch that had none

The fixture builds acknowledgements with `dtg-credentials`, which emits only the current
form, so the legacy path was unreachable from it. Both cases are hand-built: an old
acknowledgement of the right grant binds, and an old acknowledgement of a *different* grant
is still refused — the second being the failure a fallback invites, where the property is
accepted and never compared.

`vtc-service --lib`: 960 passed. Workspace clippy clean.

I took the compatible read path rather than a clean break, since it costs two branches and
some tests and does not require knowing whether anyone is running an older client. Say if
you would rather force re-acknowledgement instead — the migration is a decision, not a
mechanical consequence.

PR #1358, opened 2026-09-09
chore(rooms)!: dtg-credentials 0.6 → 0.8

stormer78:
Two breaking releases at once, and the second is why this is worth doing now rather than
eventually.

## 0.8 — a VAC is not a bearer credential

`verify_chain` now **requires the presenter to be the leaf's subject**, refusing otherwise
with `NotThePresenter`. Before, it used `presenter` for one thing — comparing it against
the leaf's *optional* `audience` — so a leaf without one was accepted from **anybody**, and
a captured presentation was a bearer token.

**Both hand-written copies of that rule are deleted.** That is the payoff.
`vti-rooms-dtg`'s chain verifier and its nomination check had each written it
independently, after hitting the gap separately, each with a comment explaining why they
had to:

> `verify_chain` uses `claimant` only for the audience check; it does not require the grant
> to name them. Without this a nomination would be a bearer token, and anyone who ever
> observed one could take the room the moment it went dormant.

Two copies of a rule is one place for it to be forgotten, and the duplicate is the one
nobody updates.

**The tests that pinned the property stay exactly where they were.** A chain presented by
the wrong party is still refused; which layer refuses it is no longer this crate's
business. Their assertions move to the library's wording, which is better — it names both
parties rather than only the rule:

> the chain's leaf grants to `did:key:z6Mkfr2…`, but it was presented by `did:key:z6Mkn5Z…`

`audience` is gone from the credential. `room_oracle::present` still accepts the
`rooms/keys/present/0.1` payload member and ignores it, so every existing caller keeps
working — the CLI names its host there. Removing it from the schema is proposed in
trustoverip/dtgwg-trust-tasks-tf#414. Accepting a request that is not wrong beats refusing
it, and both beat pretending the value still binds something.

## 0.7 — digests, and a required expiry

`validUntil` became required on `new_vac`. `rooms/owner/issue-authority` now **refuses** a
request without one rather than defaulting: nothing about a subject's standing is consulted
when a chain is verified, so authority that does not expire is authority nobody can
withdraw by waiting — and picking a lifetime here would put the room's most consequential
number somewhere its owner never looks.

The same rule made `Option` dead in two test fixtures. Every caller already passed a value,
and "a nomination that never expires" is now a state no test can construct, so both
signatures lose the `Option`.

## Ordering, for anyone else upgrading

Filed as OpenVTC/dtg-credentials#22. A **new issuer against an old verifier** reports a
broken chain rather than a version skew — `names parent zQm…, but was presented after
urn:uuid:…`, with nothing to say those are different kinds of identifier. That is the
direction a client-first upgrade hits, and it is how this whole thread started: a
browser-native room member on 0.7 talking to `room-host` on 0.6.

## Verified

| | |
|---|---|
| `vti-rooms` | 78 + 13 passed |
| `vti-rooms-dtg` | 19 passed |
| `room-host` | 23 passed |
| `vtc-service` rooms | 26 passed |
| `vta-service` rooms | 11 passed |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |

Marked `!`: `rooms/owner/issue-authority` now refuses a request it previously accepted, and
`AuthorityPresentation`'s `audience` no longer binds anything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3495,9 +3495,9 @@ dependencies = [
 
 [[package]]
 name = "dtg-credentials"
-version = "0.6.0"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f292eb5ba50d7f69dd9671e87b4d6ceddbeb973cbf0c05715220bee8446ff90"
+checksum = "4c4a250f81c3939a226b6496dcaf38a5fa90e3591ab901a9421495247adaba10"
 dependencies = [
  "affinidi-data-integrity",
  "affinidi-secrets-resolver",
@@ -3506,7 +3506,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_json_canonicalizer",
- "sha2 0.10.9",
+ "sha2 0.11.0",
  "thiserror 2.0.20",
  "tracing",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,7 +164,7 @@ affinidi-messaging-core = "0.1"
 futures-util = "0.3"
 
 # Decentralized Trust Graph Credentials
-dtg-credentials = "0.6"
+dtg-credentials = "0.8"
 
 # JSON Schema validation (issue-time credentialSchema checks)
 jsonschema = "0.49"
--- a/room-host/examples/data_room.rs
+++ b/room-host/examples/data_room.rs
@@ -592,7 +592,7 @@ async fn act_four(
     )
     .await?;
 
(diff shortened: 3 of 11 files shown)


stormer78:
Folded into #1356 — it turned out to be the last commit of that story rather than a separate one. `audience` never worked, the upstream fix removes the field, and taking 0.8 is what lets the two hand-written presenter checks go.

The rebase also improved it: #1356 had already removed `audience` and `nonce` from `room_oracle::present` entirely, which is better than the accept-and-ignore this PR was carrying.

PR #1360, opened 2026-09-09
feat(deps)!: dtg-credentials 0.6 -> 0.9.1

stormer78:
> **Stacked on #1356.** Base is `fix/presentation-audience-binds-presenter`; merge that first
> and this retargets to `main` on its own.

This workspace was three releases behind, and the one that matters is **0.8.0: a VAC is not
a bearer credential.** `verify_chain` now requires the leaf to grant to the presenter.

We were safe without it only because `vti-rooms-dtg` wrote that check by hand — **twice**,
in `lib.rs` and `nomination.rs`, having hit the gap independently each time. Both copies had
the same comment explaining why they had to exist:

> `verify_chain` takes `presenter` but uses it for **one** thing: the `audience` check… It
> does not require the leaf to grant to the presenter… Which makes it ours, and it is not
> optional: without it a presentation is a bearer token.

That is no longer true, and a stale rationale on a security check is worse than no comment.
Both are removed; the rule lives in `verify_chain`, which refuses with `NotThePresenter`, and
0.9.1 extends it to the VDC via `delegation::verify_chain`.

**The tests that pinned the behaviour are unchanged** — an agent's chain presented by the
agent's human, a nomination claimed by the wrong party — and still expect a refusal. That is
what shows the rule survived the move rather than being quietly dropped.

## Pinned `=0.9.1`, deliberately

0.9.1 is **API-breaking under a patch version** — its own changelog says so, and the version
was chosen knowingly because 0.9.0 had published hours earlier with no consumers. A caret
`"0.9"` would let a routine `cargo update` take that break without anyone deciding to. The
exact pin makes upgrading an act rather than an accident.

## `validUntil`, and why the handler refuses instead of defaulting

`new_vac`'s `valid_until` stopped being an `Option` in 0.7.0, when WD02 made `validUntil`
REQUIRED on a VAC. `rooms/owner/issue-authority` passes `req.valid_until` straight through
and the `0.1` payload types it optional — so this workspace could ask for a **chain root
that never expires**, which no verifier would have accepted.

trustoverip/dtgwg-trust-tasks-tf#418 makes the member REQUIRED in `0.2`, and its generated
payload non-optional, which retires the question entirely. Until this task moves to `0.2`
the check lives in the handler — and it **refuses** rather than substituting a default. How
long a room's authority lasts is the owner's judgement about their own room; a recipient
that picked a lifetime would be making that judgement silently, in the one place the owner
cannot see it.

## Blocked upstream: the 0.2 cut-over is not in this PR

Moving to `rooms/keys/present/0.2` and `rooms/owner/issue-authority/0.2` needs
`trust-tasks-rs 0.19`, and that is blocked by a dependency of ours, not by us:

```
affinidi-messaging-sdk 0.22.0  ->  trust-tasks-rs 0.18.9
```

Pinning this workspace at 0.19 therefore puts **two** versions of `trust-tasks-rs` in the
graph, and `MediatorAcl` — which crosses the `vta-sdk` / `affinidi-messaging-sdk` boundary —
becomes two distinct types. The compiler names it directly: *"there are multiple different
versions of crate `trust_tasks_rs` in the dependency graph"*. Bumping the four sibling
`trust-tasks-*` pins together does not help; the duplicate arrives through messaging-sdk.

This is the public-dependency hazard in its usual shape: nothing in our own manifests is
wrong, and no semver check would have flagged it. The cut-over follows once messaging-sdk
re-pins to 0.19 — worth raising there.

## Verified

- `cargo clippy --workspace --all-targets` — clean
- `cargo fmt --all` — clean
- `cargo test -p vti-rooms-dtg -p vti-rooms -p vta-sdk -p vta-cli-common` — 840 passed, 0 failed

Lockfile moves only `dtg-credentials` (and `sha2` beneath it); nothing else resolves
differently.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3495,9 +3495,9 @@ dependencies = [
 
 [[package]]
 name = "dtg-credentials"
-version = "0.6.0"
+version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f292eb5ba50d7f69dd9671e87b4d6ceddbeb973cbf0c05715220bee8446ff90"
+checksum = "fa51b03528a22ad07cd0f812a99fae834b7867c205943a40ad90bb8eb62edb08"
 dependencies = [
  "affinidi-data-integrity",
  "affinidi-secrets-resolver",
@@ -3506,7 +3506,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_json_canonicalizer",
- "sha2 0.10.9",
+ "sha2 0.11.0",
  "thiserror 2.0.20",
  "tracing",
 ]
@@ -9591,9 +9591,9 @@ dependencies = [
 
 [[package]]
 name = "trust-tasks-capability-client"
-version = "0.18.0"
+version = "0.18.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17b3ff12a929ebae90844cbf38232092a60acbc59591999ad57453d6d19bb183"
+checksum = "df30d9c2a63dc5147b9fee773cc8ccce201b45ffd6a3a4724dd4a8e02dfdbbd2"
 dependencies = [
  "chrono",
  "serde",
@@ -9605,9 +9605,9 @@ dependencies = [
 
 [[package]]
 name = "trust-tasks-didcomm"
-version = "0.18.0"
+version = "0.18.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b338e33658f9e40246f014cd7695331214e529afee8a465f55ac868bf606b978"
+checksum = "16c64a454241025f8a83d7f0ecc39e3b745a25ddf5abcc1cbe2267df5ede7ed5"
 dependencies = [
  "affinidi-messaging-didcomm",
  "base64 0.22.1",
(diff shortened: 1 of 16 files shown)


stormer78:
The upstream blocker is now fixed rather than waiting on: **affinidi/affinidi-tdk-rs#784** moves `trust-tasks-rs` 0.18 → 0.19 across all six messaging crates.

That workspace already documents this exact failure (`expected MediatorAcl, found a different MediatorAcl`) in two `Cargo.toml` comments as the signature of a mixed `trust-tasks-rs` graph, so the diagnosis matches a hazard it had already been bitten by twice — the difference this time is that the mixed graph escaped into a consumer.

Once #784 publishes, the follow-up here is small and already scoped: re-pin the five `trust-tasks-*` deps to 0.19, move the two type URIs to `present/0.2` and `issue-authority/0.2`, and delete the hand-rolled `validUntil` check in `room_owner.rs` — the `0.2` payload types it non-optional, so the handler's guard becomes the schema's job.

This PR stays as it is: the dtg-credentials half stands on its own and does not need any of that.

stormer78:
Superseded by `e4b92fb7` on #1356's own branch, which does the same bump and does it more completely — it also covers `vti-rooms-wasm` (`identity.rs`, `lib.rs`) and `vti-rooms/src/authz.rs`, which this PR missed because `vti-rooms-wasm` is not built by a default `cargo check --workspace`.

We arrived at the same two judgement calls independently, which is reassuring on both: delete both hand-written presenter checks rather than keep them as belt-and-braces, and make `rooms/owner/issue-authority` **refuse** a request with no `validUntil` rather than default one.

One difference worth a look, take it or leave it: that commit pins `dtg-credentials = "0.9.1"` where this one used `"=0.9.1"`. Caret `0.9.1` means `>=0.9.1, <0.10.0`, so a future 0.9.2 is taken on a routine `cargo update`. 0.9.1 itself shipped an **API break under a patch version** — its changelog says so outright — so that range has already been demonstrated to carry breaks. The floor is the important part and it is right either way; the exact pin just makes the next one a decision rather than a surprise.

Nothing else here is lost. What remains unbuilt by either PR is the `0.2` cut-over, now unblocked by affinidi/affinidi-tdk-rs#784.

PR #1433, opened 2026-09-11
ci: full vetting stack against main (DO NOT MERGE)

stormer78:
**Do not merge.** This is a temporary PR that only runs real CI against `main` on the whole peer-identity-vetting stack at once. Stacked PRs don't get CI, because `ci.yml` only runs on pull requests into `main`.

The head is #1432's branch with #1431 merged in (#1431 and #1432 both sit on #1430). Its tree is the combined code of:

#1425 → #1426 → #1427 → #1428 → #1429 → #1430 → #1431 + #1432

**How it's used:**
1. Once this is green, the stack merges bottom-up in one sitting. Each PR is rebased onto `main` and merged straight away, without waiting for its own CI.
2. Once the whole chain is in, `main` holds exactly the code this PR tested, as long as nothing else lands on `main` in between.
3. This PR is closed once the stack has merged, and `ci/vetting-stack-integration` is deleted.

**The merge commit resolves one conflict:** #1431 and #1432 both add a section at the same place in `docs/03-vtc/vetting.md`. The resolution keeps both, with the `cnm` walkthrough steps (7, 8) before "In the admin console". Whichever of the two merges second needs that same resolution when it's rebased.


stormer78:
Closing: this temporary PR only ran CI on the combined vetting stack. The stack (#1425–#1432) has merged into main.

PR #1436, opened 2026-09-11
security(vta-sdk)!: vet DID-advertised VTA endpoints and bind cached tokens to their origin

stormer78:
## Why

A VTA's REST base URL comes from the `#vta-rest` service in its DID document (or is synthesized from the DID's domain). Until now it was used as-is: any scheme, any host, with embedded credentials. Every later authenticated call went there. Separately, `SessionStore::ensure_authenticated` returned the cached access token whatever base URL it was called with. If the advertised endpoint changed, a live bearer token for the original VTA went to the new host.

While fixing that, the existing foreign-fetch guard (`vta_sdk::http::guard_public_url`, which fronts the VTC recognise/present paths and vta-vault status lists) turned out to pass IPv4-mapped IPv6 loopback, carrier-grade NAT addresses (including a cloud metadata address) and `localhost`. It also did no DNS-level check. A unit test against `main` confirmed all three literal cases.

### `vta-sdk::http`
- **One classifier.** `classify_ip` / `classify_host` (with `IpClass` / `HostClass`) cover:
  - IPv4 private, loopback, link-local, CGNAT, `0/8`, broadcast, multicast, documentation, benchmarking and reserved space;
  - IPv6 loopback, unspecified, ULA, link-local, site-local, multicast, documentation, Teredo, discard and local-use NAT64;
  - IPv4 embedded in IPv6 (mapped, compatible, NAT64 `64:ff9b::/96`, 6to4), classified by the embedded address;
  - named cloud-metadata addresses (`fd00:ec2::254`, `100.100.100.200`);
  - local-only names (`localhost`, `*.localhost`, `*.local`, `*.internal`, `*.home.arpa`, single-label names) and metadata hostnames.

  It mirrors `affinidi-did-web`'s classifier. A comment marks it for replacement by the shared `affinidi-net-guard` crate once that is published.
- **`guard_public_url`** uses the classifier. Its signature is unchanged.
- **`foreign_fetch_client`** now installs a DNS resolver that fails closed if any answer for a name is non-public, and connects only to the vetted addresses. It also sets `no_proxy()`, so a proxy cannot resolve the name on its behalf.
- **`guard_vta_endpoint(url, EndpointPolicy)`**:
  - `https` required; `http` only to `localhost`, `127.0.0.0/8` or `::1` (the same predicate as `vta-webvh`'s webvh client).
  - Always refused: userinfo, other schemes, link-local and metadata addresses and names, unspecified, broadcast, multicast, documentation, benchmarking and reserved addresses, and IPv6 forms embedding a non-public IPv4 address.
  - RFC 1918, ULA, CGNAT and local-only names are accepted only when the private-endpoint policy is enabled. The refusal names the opt-in, and URL credentials are never echoed.
- **`EndpointPolicy::process_default()`** enables private endpoints when `VTA_ALLOW_PRIVATE_ENDPOINTS` is truthy or the binary called `set_allow_private_endpoints(true)`.
- **`rest_client()`** follows a redirect only while scheme, host and port stay the same as the original request. Otherwise the `3xx` is returned to the caller.

### `vta-sdk::session`
- Every REST URL from `resolve_vta_endpoint_with_resolver` (and so `resolve_vta_endpoint` and `provision_client::resolve_vta`), `resolve_vta_url` and the forced-REST path of `connect_with_transport` passes `guard_vta_endpoint`. That covers both the advertised `#vta-rest` and the URL synthesized from the DID.
  - A refused URL is an error, never a silent fallback. This includes a document that also advertises TSP/DIDComm, since the REST URL is carried in those variants.
- New `resolve_vta_endpoint_with_resolver_and_policy` takes an explicit policy.
- The persisted session gains `token_origin` (serde-defaulted). It is set wherever a token is stored (`login`, `ensure_authenticated`). A cached token is returned only when the requested base URL has the same origin; otherwise the session re-authenticates. Sessions saved before this change have no origin and re-authenticate once.

### CLIs
- `cnm` and `pnm` gain a global `--allow-private-endpoints` flag (also read from `VTA_ALLOW_PRIVATE_ENDPOINTS`).
- `cnm backup`, `cnm audit verify` and `pnm bootstrap connect` use `vta_sdk::http::rest_client()` instead of a bare `reqwest::Client::new()`, and read responses through `read_body_capped`:
  - backup: 64 MiB, matching the VTC's backup-import request cap;
  - audit verify and bootstrap: 1 MiB.
- `cnm-cli/clippy.toml` and `pnm-cli/clippy.toml` add a `disallowed-methods` entry for `reqwest::Client::new`. Clippy picks config up from each crate's directory, so other crates are unaffected.

## Operator-visible behaviour changes
- A VTA whose DID document advertises a private-network REST endpoint (for example `https://10.0.0.5`) is now refused by default. Pass `--allow-private-endpoints` or set `VTA_ALLOW_PRIVATE_ENDPOINTS=1`. This applies to any SDK consumer resolving VTA endpoints, not only the CLIs.
- Plain `http://` endpoints are refused unless the host is loopback. Local-dev setups using `http://localhost:…` or `http://127.0.0.1:…` keep working.
- A cached token is no longer reused against a different origin; the CLI re-authenticates instead.
- `cnm backup` and `cnm audit verify` now have the SDK's request timeouts (30 s by default, `VTA_REST_TIMEOUT_SECS` to raise).
- Foreign fetches (status lists, recognition) refuse hostnames that resolve to non-public addresses, and no longer go through `HTTP(S)_PROXY`.
- `--url` overrides are unchanged: an operator-supplied URL is not run through the endpoint guard.

## Merge note
A separate change reorders `SessionStore::login` so the session is saved only after authentication succeeds. This PR touches `login` only to add `token_origin: None` to the initial `Session` and to record `session.token_origin` next to the token. Expect a small textual conflict there; keep both.

## Tests
- `vta-sdk/src/http.rs`:
  - classifier block/allow vectors, URL canonicalisation vectors, local-name and scheme vectors;
  - stubbed DNS answers (mixed, mapped, NAT64, empty), and a real `localhost` lookup refused by the resolver;
  - `guard_vta_endpoint` accept, always-refuse and opt-in cases, and credential redaction;
  - a same-origin redirect is followed and a cross-origin redirect is not (the target sees no request).
- `vta-sdk/src/session.rs`: legacy session without `token_origin` loads; origin normalisation; forced-REST URL vetting over a seeded resolver.
- `vta-sdk/tests/session_store.rs`:
  - seeded DID documents through `resolve_vta_endpoint_with_resolver`: the metadata URL is refused even with the opt-in; a userinfo URL is refused without echoing credentials; `https://10.0.0.5` is refused unless opted in; `http://127.0.0.1:8100` is accepted; a bad REST URL beside a DIDComm mediator is refused;
  - a cached token is reused for the same origin (including a `/v1` path) but not for a different origin.

### DIFF
--- a/cnm-cli/clippy.toml
+++ b/cnm-cli/clippy.toml
@@ -1,0 +1,6 @@
+# Every HTTP call from this CLI goes through `vta_sdk::http::rest_client()`,
+# which carries request/connect timeouts and refuses to follow a redirect off
+# the original origin. A bare client has neither.
+disallowed-methods = [
+  { path = "reqwest::Client::new", reason = "use vta_sdk::http::rest_client() (timeouts, same-origin redirects)" },
+]
--- a/cnm-cli/src/audit.rs
+++ b/cnm-cli/src/audit.rs
@@ -19,6 +19,9 @@ use crate::auth;
 /// `vti_common::trust_task::HEADER_NAME`).
 const TRUST_TASK_HEADER: &str = "Trust-Task";
 const VERIFY_TASK: &str = "https://trusttasks.org/spec/audit/verify/0.1";
+/// Largest verification report read into memory. The report is a handful of
+/// counters and identifiers; this is generous headroom.
+const MAX_VERIFY_RESPONSE_BYTES: usize = 1024 * 1024;
 
 /// `cnm audit verify` — walk the community's audit chain and report.
 ///
@@ -33,14 +36,18 @@ pub async fn cmd_verify(
         .ok_or("VTC audit verify requires a REST connection to the VTC")?;
     let token = auth::ensure_authenticated(base, keyring_key).await?;
 
-    let resp = reqwest::Client::new()
+    // The SDK client: finite timeouts, and no redirect off the VTC's origin.
+    let resp = vta_sdk::http::rest_client()
         .get(format!("{base}/audit/verify"))
         .bearer_auth(&token)
         .header(TRUST_TASK_HEADER, VERIFY_TASK)
         .send()
         .await?;
     let status = resp.status();
-    let text = resp.text().await.unwrap_or_default();
+    let bytes = vta_sdk::http::read_body_capped(resp, MAX_VERIFY_RESPONSE_BYTES)
+        .await
+        .map_err(|e| format!("VTC audit verify ({status}): {e}"))?;
+    let text = String::from_utf8_lossy(&bytes);
     if !status.is_success() {
         return Err(format!("VTC audit verify failed ({status}): {text}").into());
     }
--- a/cnm-cli/src/backup.rs
+++ b/cnm-cli/src/backup.rs
@@ -26,6 +26,10 @@ use crate::auth;
 const TRUST_TASK_HEADER: &str = "Trust-Task";
 const EXPORT_TASK: &str = "https://trusttasks.org/spec/vtc/backup/export/0.1";
 const IMPORT_TASK: &str = "https://trusttasks.org/spec/vtc/backup/import/0.1";
+/// Largest VTC backup response read into memory. Matches the VTC's cap on a
+/// backup import request body, so an export larger than this could not be
+/// restored anyway.
+const MAX_BACKUP_RESPONSE_BYTES: usize = 64 * 1024 * 1024;
 
 /// Authenticated REST POST to a VTC `/v1/backup/*` route. `client.rest_url()`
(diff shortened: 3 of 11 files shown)


stormer78:
Heads-up: this and #1437 both change `SessionStore::login` in `vta-sdk/src/session.rs` and will conflict there. #1436 records the token's origin; #1437 saves the session only after authentication succeeds. Whichever merges second needs a rebase, and I'll do it once the first one lands.

PR #1437, opened 2026-09-11
security(sealed-transfer): check who sealed an admin credential before installing it

stormer78:
## Summary

A sealed bundle is HPKE-encrypted to the consumer's request key. That keeps it
confidential, but it does not show who produced it: anyone who has seen the
bootstrap request can seal a bundle to it. Before this change, `cnm` installed an
admin credential from a bundle with no check tying it to the operator's VTA:

- the setup wizard treated an empty digest as "skip verification";
- a `DidSigned` assertion passed `open_bundle` regardless of which DID signed it,
  and no install path verified the assertion;
- the bundle's `vta_did` was never compared with the DID the operator typed;
- `SessionStore::login` stored the bundle as the session **before**
  challenge-response, so a login that failed still left that `vta_did` as the
  session later commands used.

## What producer assertion do admin credential bundles carry today?

Every path that seals an `AdminCredential` or `ContextProvision` for a consumer to
install from a file uses **`PinnedOnly` with a throwaway producer `did:key`**
(`vta_cli_common::sealed_producer::seal_for_recipient`; the private half is dropped
immediately):

| Producer | Payload | Assertion |
|---|---|---|
| `pnm auth-credential create`, `cnm auth-credential create` | `AdminCredential` | `PinnedOnly`, ephemeral did:key |
| `pnm contexts bootstrap`, `cnm contexts bootstrap` | `AdminCredential` | `PinnedOnly`, ephemeral did:key |
| `pnm contexts provision`, `pnm contexts reprovision` | `ContextProvision` | `PinnedOnly`, ephemeral did:key |
| `vta` offline context reprovision (`bootstrap_cli::run_context_reprovision`) | `ContextProvision` | `PinnedOnly`, ephemeral did:key |
| `vta bootstrap seal` (Mode C) | any payload file | `PinnedOnly`, ephemeral did:key |
| TEE `POST /bootstrap/request` | `AdminCredential` | `Attested`; consumed only by `pnm bootstrap connect`, which verifies the quote |
| provision-integration (`#sealed-transfer-0`) | `TemplateBootstrap` / `AdminRotation` | `DidSigned`, producer = VTA DID; not an admin-credential install path |

So for admin credentials the out-of-band digest is the only anchor that exists
today, and an expected VTA DID is enough for the `DidSigned` case. No
`--expect-producer-did` flag is added.

### `vta-cli-common::sealed_consumer`

- `open_admin_credential(bundle_path, config_dir, expect_digest, expected_vta_did)`
  opens the bundle and calls `verify_admin_bundle`, which matches the verified
  assertion exhaustively (the same pattern as `verify_template_bundle` in
  `vta-service`):
  - `PinnedOnly`: accepted only with `expect_digest`;
  - `DidSigned`: accepted only if the producer DID equals `expected_vta_did` and
    the Ed25519 signature verifies. Any other signer is refused, with or without
    a digest. Only `did:key` producers can be verified here;
  - `Attested`: refused;
  - when `expected_vta_did` is given, `credential.vta_did` must equal it.
- The request secret is removed only after the bundle is accepted, so a rejected
  bundle does not use up the pending request. `open_armored_bundle` keeps its
  existing behaviour for inspection.
- `normalize_expected_digest`: 64 hex characters, trimmed, lower-cased.

### `cnm-cli`

- `setup` / `community add`: the digest prompt requires 64 hex characters (the
  "empty = skip" option is gone), and the credential is checked against the VTA
  DID typed in the same wizard step.
- `auth login --credential-bundle`: flags go through `validate_digest_flags`, and
  the bundle is checked against the community's configured `vta_did`.
  `--no-verify-digest` stays, but only a bundle signed by that DID is then
  accepted.
- `bootstrap open` (inspection) is unchanged, including `--no-verify-digest`.

### `pnm-cli`

- There is no `pnm auth login --credential-bundle` subcommand in the code (the
  README and `phase5-cutover.md` still mention one). The install-shaped path is
  `pnm bootstrap open --out`, which writes the credential for file-based
  consumers. It now goes through `verify_admin_bundle`, checked against
  `--expect-vta-did` when given. Without `--out`, `bootstrap open` only inspects
  and is unchanged.

### `vta-sdk`

- `SessionStore::login` runs challenge-response first and saves the session only
  on success. An existing session under the same key is left untouched by a
  failed login.

## Behaviour changes

- The `cnm` wizard no longer accepts an empty digest.
- `cnm auth login` refuses a credential whose `vta_did` differs from the
  community's configured VTA DID, and refuses `--expect-digest` combined with
  `--no-verify-digest`.
- `cnm auth login --no-verify-digest` fails for every bundle `pnm`/`cnm` produce
  today, because they are `PinnedOnly`.
- `pnm bootstrap open --out --no-verify-digest` no longer writes a credential
  file for a `PinnedOnly` bundle.
- If `cnm auth login` opens a bundle but authentication then fails (e.g. the VTA
  is unreachable), no session is stored. The request secret has already been
  used by then, so a retry needs a new bootstrap request and bundle.

## Testing

New tests in `vta-cli-common::sealed_consumer`. Each seals a real bundle to a real
bootstrap request:

- `admin_credential_signed_by_an_unexpected_did_key_is_rejected`: a correctly
  signed `DidSigned` assertion from a freshly generated did:key, with the
  credential pointing at that key's own DID. Refused against
  `did:webvh:victim` without a digest, with the bundle's own digest, and with no
  expected DID at all. The request secret is still present afterwards.
- `admin_credential_signed_by_the_expected_vta_is_accepted`: producer DID ==
  expected VTA DID, valid signature, no digest. Accepted, and the secret is
  consumed.
- `did_signed_by_another_key_in_the_vta_name_is_rejected`: the assertion names the
  expected VTA but is signed by a different key.
- `pinned_only_admin_credential_needs_the_digest`: refused without a digest (secret
  kept); accepted with the digest as an operator might type it (upper-case,
  surrounding whitespace).
- `admin_credential_for_a_different_vta_is_rejected`: valid digest, but
  `credential.vta_did` differs from the expected DID.
- `attested_admin_credential_is_rejected`.
- `expected_digest_must_be_64_hex_characters`.

`cnm-cli::setup`: `digest_prompt_rejects_empty_input`,
`digest_prompt_accepts_only_a_sha256_hex_digest`.

`vta-sdk/tests/session_store.rs`:

- `login_against_an_unreachable_vta_persists_no_session`: `login` against
  `http://127.0.0.1:1` errors, and `has_session` / `loaded_session` show nothing
  was stored.
- `failed_login_keeps_the_existing_session`: an earlier session under the same
  key survives a failed login.
- `login_propagates_challenge_failure` now also asserts no session is stored.

Run locally (macOS, rustc 1.98.1):

- `cargo fmt --check`: clean.
- `cargo clippy -p vta-sdk -p vta-cli-common -p cnm-cli -p pnm-cli --all-targets -- -D warnings`: clean.
- `cargo clippy -p vta-sdk --features session,test-support --lib --test session_store -- -D warnings`: clean.
- `cargo test -p vta-cli-common -p cnm-cli -p pnm-cli`: all pass (vta-cli-common
  107, cnm-cli 28, pnm-cli 62).
- `cargo test -p vta-sdk --features session,test-support`: 686 passed, 0 failed,
  10 ignored, including `tests/session_store.rs`.

`cargo clippy -p vta-sdk --features session,test-support --all-targets -- -D warnings`
fails with `items_after_test_module` at
`vta-sdk/src/protocols/acl_management/swap.rs:118`. This change doesn't touch that
file, and the same command fails the same way on `main` (5a6d4923).

Not exercised end to end: an interactive `cnm setup` run, and a `DidSigned` bundle
from a `did:webvh` producer. Those can't be verified without DID resolution and
are refused.

### DIFF
--- a/cnm-cli/README.md
+++ b/cnm-cli/README.md
@@ -149,8 +149,10 @@ tokens:
 
 ```sh
 # Apply a sealed admin credential bundle (e.g. a backup-restore handoff
-# or a sealed transfer from another operator)
-cnm auth login --credential-bundle <file>
+# or a sealed transfer from another operator). The digest comes from the
+# operator who sealed it, over a separate channel you trust; the credential
+# must be for the community's configured VTA DID.
+cnm auth login --credential-bundle <file> --expect-digest <sha256>
 
 # Check auth status
 cnm auth status
--- a/cnm-cli/src/main.rs
+++ b/cnm-cli/src/main.rs
@@ -416,7 +416,10 @@ enum AuthCommands {
         /// Expected SHA-256 digest, communicated out-of-band by the producer.
         #[arg(long)]
         expect_digest: Option<String>,
-        /// Skip out-of-band digest verification (testing only — prints a warning).
+        /// Skip out-of-band digest verification. The bundle is then accepted
+        /// only if it is signed by the community's configured VTA DID; the
+        /// admin credentials pnm and cnm seal carry no such signature and
+        /// need --expect-digest.
         #[arg(long)]
         no_verify_digest: bool,
     },
@@ -761,31 +764,27 @@ fn print_banner() {
 /// Implementation of `cnm auth login --credential-bundle <file>`.
 ///
 /// Opens an armored sealed bundle (matching a secret persisted earlier by
-/// `cnm bootstrap request`), extracts the admin credential from the payload,
-/// and installs it via `auth::login`.
+/// `cnm bootstrap request`), verifies its producer and VTA DID against
+/// `expected_vta_did` (the community's configured DID, when there is one),
+/// and installs the admin credential via `auth::login`.
 async fn auth_login_sealed(
     client: &VtaClient,
     keyring_key: &str,
     credential_bundle: &std::path::Path,
     expect_digest: Option<&str>,
     no_verify_digest: bool,
+    expected_vta_did: Option<&str>,
 ) -> Result<(), Box<dyn std::error::Error>> {
     let config_dir =
         config::config_dir().map_err(|e| format!("could not resolve config dir: {e}"))?;
-    if no_verify_digest {
-        vta_cli_common::sealed_consumer::warn_no_verify_digest();
-    }
-    let opened = vta_cli_common::sealed_consumer::open_armored_bundle(
+    vta_cli_common::sealed_consumer::validate_digest_flags(expect_digest, no_verify_digest)?;
+    let bundle = vta_cli_common::sealed_consumer::open_admin_credential(
         credential_bundle,
         &config_dir,
         expect_digest,
-        no_verify_digest,
+        expected_vta_did,
     )?;
-    eprintln!(
-        "Sealed bundle opened ({} — digest {}).",
-        opened.bundle_id_hex, opened.digest
-    );
-    let bundle = vta_cli_common::sealed_consumer::extract_admin_credential(opened.payload)?;
+    eprintln!("Sealed bundle opened and verified.");
     let base = client
         .rest_url()
         .ok_or("login requires a REST connection to the VTC")?;
@@ -1079,7 +1078,7 @@ async fn main() {
                 "Error: could not bootstrap session from personal VTA: {e}\n\n\
                          To fix this, either:\n  \
                          1. Import a sealed credential bundle from your VTA admin:\n     \
-                            cnm auth login --credential-bundle <bundle.armored> [--expect-digest <sha256>]\n  \
+                            cnm auth login --credential-bundle <bundle.armored> --expect-digest <sha256>\n  \
                          2. Re-run setup: cnm setup"
             );
             std::process::exit(1);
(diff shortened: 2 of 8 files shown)


stormer78:
Heads-up: this and #1436 both change `SessionStore::login` in `vta-sdk/src/session.rs` and will conflict there. Whichever merges second needs a rebase, and I'll do it once the first one lands.

stormer78:
Rebased onto `main` (now including #1436). The only conflict was `SessionStore::login` in `vta-sdk/src/session.rs`, where both changes apply, so both are kept: the session is built **after** `challenge_response` succeeds (this PR) and carries `token_origin: url_origin(base_url)` (#1436). The pre-authentication `save_session` is gone, so a bundle that fails to authenticate still leaves no session behind.

Checked after the rebase: `cargo fmt --all --check` clean, `cargo clippy -p vta-sdk -p vta-cli-common -p cnm-cli -p pnm-cli --all-targets -- -D warnings` clean, and tests pass for those crates (vta-sdk 695).

PR #1454, opened 2026-09-14
fix(pnm-cli): anchor TEE bootstrap connect by DID and PCR0

oleksiipiliugin:
Add mutually exclusive --vta-did and --vta-url targets. Resolve WebVH locally without guessing a URL on failure, preserve the advertised REST endpoint, and reject credentials for a different VTA DID.

Accept a pinned PCR0 as the online connect trust anchor without requiring the server-generated digest or an opt-out warning. Preserve explicit digest opt-out warnings, conflicting-flag errors, and mandatory offline digest verification.

Add regression tests for target selection, strict endpoint resolution, credential identity checks, and anchor combinations. Update the PNM quick start and TEE bootstrap guide.


marat-affinidi (approved):

stormer78:
Reviewed this and pushed four commits to the branch rather than leaving a list of asks — the merge conflict needed resolving anyway and the rest was small. Everything below is on top of your work; revert any of it freely.

The core change is right. Anchoring on `--expect-pcr0` instead of a digest that is minted *during* the call is correct, and the codebase already agreed with you — `bootstrap.rs` has said "Attestation + TLS give the primary integrity anchor; this is a belt-and-suspenders check" since before this PR. Mandatory `--expect-digest` on connect was in practice forcing every operator onto `--no-verify-digest`, which is strictly worse than what you've replaced it with. `check_pcrs` genuinely enforces the pin, the anchor check runs before any network I/O, and `check_connect_vta_did` closes a real gap.

## What I changed

**1. Dropped the `orchestrator.rs` hunk (this was the conflict).** #1456 landed on `main` a few hours before I looked, solving the same problem as `vta_support::next_version_time`. It's a superset of the clamp here: it truncates to whole seconds (the only precision that survives serialisation — comparing at full precision lets two entries microseconds apart pass the clamp and then serialise identically), caps the index offset at 1439 minutes, and reaches a strictly-increasing timestamp without a `tokio::time::sleep` on the request path. Your version's safety rested on the comment *"the chain was validated above, so its head is not in the future"* — an assumption about didwebvh-rs's **resolve-time** check that nothing enforces locally at write time, so an imported log or enclave clock skew would park a request handler. Resolved in favour of main's.

Worth noting: `tee_existing_identity_rest_repair_appends_signed_update` — the test whose comment reads *"No sleep: even a repair immediately after TEE genesis must validate"* — passes unchanged against main's implementation. #1456 also backdates TEE genesis, so the scenario you were protecting against is covered from the other side.

**2. `--expect-pcr8` wasn't validated pre-flight, and the carve-out is single-use.** This is the one I'd most want a second opinion on, because it's the only finding here that can cost someone a VTA.

Your own test is named `connect_rejects_malformed_pcr0_before_resolving_or_posting`, and the reason that name matters is `routes/bootstrap.rs` — the carve-out sentinel is written **before** the bundle is returned, behind the durability barrier. So any client-side pin rejected only at `check_pcrs` is rejected *after* the VTA's one and only first boot has been spent. `validate_connect_anchor` checked PCR0 and ignored `expect_pcr8` entirely, so a 95-character or stray-non-hex PCR8 survived to comparison time. The operator is then holding a closed carve-out, no installed credential, and no second attempt — a typo in an optional defence-in-depth flag becomes an unbootstrappable VTA.

Validated on the same rule as PCR0. The regression now runs both pins, holding the one not under test well-formed so a failure can only come from the pin named in `which`.

**3. Guard rejections no longer suggest `--vta-url`.** `resolve_connect_url` wrapped *every* error, `guard_vta_endpoint` rejections included, with "explicitly use `--vta-url <URL>`". But `--vta-url` isn't guarded (`(None, Some(url)) => url`), so for a DID advertising a private address, cloud metadata or a non-HTTP scheme the message told the operator to route around the control. The guard already carries its own remedy (`--allow-private-endpoints` / `VTA_ALLOW_PRIVATE_ENDPOINTS`, and `process_default()` correctly picks up pnm's global flag), so guard failures are now a separate arm that surfaces that and nothing else. The test asserts it still names the DID, still redacts URL userinfo, still names the guard's remedy, and does *not* mention `--vta-url`. Your `file:///tmp/vta` case moved into the guard test, where it was always a guard rejection rather than a resolution failure.

**4. The error now says resolution is local-only.** Ignoring `PNM_RESOLVER_URL` / `[resolver_url]` here is the right call and I kept it — the resolved document picks the endpoint that mints a super-admin credential against a carve-out that can only be spent once, so that choice shouldn't be delegated to a remote resolver. But `main.rs` documents that setting as reaching *every* `DIDCacheClient` in PNM, and this is now the one that doesn't. The error said "Retry once its DID log is published", which is the wrong diagnosis for an operator whose egress reaches only a resolver sidecar and whose log is already published — they'd wait for something that had already happened. The exception is now named in the error and in the README.

**5. Repointed the `bootstrap open` operator hint** (`bootstrap.rs`), which still printed `--vta-url <url>` plus an `--expect-digest` this PR establishes cannot be known before the call.

**6. `public_url` is read once in `did_autogen.rs`.** The new `VTARest` block trimmed whitespace; the `TeeAttestation` block beside it didn't — and `trim_end_matches('/')` strips nothing from a string ending in whitespace. So `public_url = "  https://host/  "` produced a usable REST endpoint and, in the same signed genesis, a TeeAttestation endpoint of `"  https://host/  /attestation/report"`. Genesis isn't rewritten on reboot, so that's permanent. The discovery test gains the padded-URL case with `embed_in_did` **on** — it previously exercised padding only with the attestation service off, which is exactly the combination that hid this — and now asserts no advertised `serviceEndpoint` carries surrounding whitespace, so a third service can't reintroduce it.

**7. CLAUDE.md.** This PR relaxes something CLAUDE.md lists under the sealed-transfer invariants marked *do not relax* ("digest pinning is mandatory at the CLI"). The relaxation is sound, but an invariant with an exception needs the exception written next to it, or the next reader applies the rule as stated and the one after that treats the divergence as precedent. Recorded as one named exception with its reason and its boundary. Also updated the TEE Mode B entry point, which still read `--vta-url <url>`, and `sealed-bootstrap.md`.

## Verification

Green locally on what CI actually runs, not an approximation:

```
cargo test --workspace --exclude vtc-service --exclude vtc-client --no-fail-fast   4276 passed, 0 failed
cargo test -p vtc-service -p vtc-client                                           1611 passed, 0 failed
cargo test -p vta-service --no-default-features --features tee,rest,didcomm       1523 passed, 0 failed
cargo clippy --workspace --all-targets -- -D warnings                             clean
cargo clippy -p vta-service --no-default-features --features tee,rest,didcomm --all-targets -- -D warnings   clean
cargo fmt --all -- --check                                                        clean
```

The tee-feature run matters here: `cargo test --workspace` never compiles `tee_webvh`'s tests, so the crate-level run alone says nothing about the two tests this PR adds there.

One process note, not a criticism of the change: the previous all-green run was against base `76eae3f8`, five commits behind — it never saw #1456. Worth re-reading the new run rather than the old rollup.

## Two things I left alone

- **`--vta-url` is still unguarded.** That's defensible — the guard is for remotely-controlled endpoints, and an operator-supplied URL isn't one. It's also not a regression, since `--vta-url` was the only path before this PR. Flagging it only so the asymmetry is a decision rather than an accident.
- **`check_connect_vta_did` is necessarily post-hoc**, so the mismatch it catches is caught after the target VTA's carve-out has closed. Inherent to the flow rather than something this PR introduced, and the check is still worth having — it stops the credential being installed.

stormer78 (approved):
LGTM

PR #1456, opened 2026-09-14
fix(webvh): clamp versionTime against the previous log entry

yaroslava-kurash:
## Summary

Fixes TEE VTA `did:webvh` identities becoming unresolvable after their first update.

## Problem

PR #600 introduced `backdated_version_time` for the standard VTA genesis and update paths. The TEE genesis path was not updated and continued relying on `didwebvh-rs`'s default `Utc::now()` timestamp.

As a result:

1. TEE genesis received a real current `versionTime`.
2. The first runtime update received a timestamp backdated by one day.
3. The update timestamp was earlier than genesis.
4. `didwebvh-rs` rejected the chain because `versionTime` must be strictly increasing.

This affected mediator connection, DIDComm operations, backup export, and other flows that resolve the VTA's DID.

## Solution

Reuse the PR #600 backdating policy for TEE genesis:

```rust
.version_time(backdated_version_time(0))
```

The helper is now owned by `vta-support::version_time` so both `vta-service` and `vta-tee` use the same implementation without creating a dependency cycle.

The policy remains:

```text
versionTime = now - 1 day + entry_index * 1 minute
```

This keeps timestamps safely in the past and strictly increasing at the whole-second precision required by `did:webvh`.

## Testing

- Shared `vta-support` version-time regression test passes.
- `vta-service` did:webvh test suite passes.
- `vta-tee` test suite passes: 35 tests.
- Clippy passes for `vta-support` and `vta-tee`.
- `cargo fmt --all -- --check` passes.


stormer78:
Reviewed this — the diagnosis is right and the TEE genesis path was genuinely the only mint site left un-backdated (I checked: there are exactly three production log-entry writes, the two `CreateDIDConfig` genesis sites and the update orchestrator). I've pushed a follow-up commit to the branch rather than asking you to round-trip it, since one of the findings is more urgent than the PR as it stood.

## The fix only helped DIDs minted after it

`backdated_version_time` derives its timestamp from the entry index alone, which silently assumes every earlier entry in the chain was stamped by the same policy. For a TEE VTA **provisioned before this PR**, that assumption is false — its genesis carries a real `Utc::now()`, so its first runtime update is still computed as `now - 1 day + 1 min`, i.e. earlier than genesis. Exactly the bug you're fixing, on the VTAs that already exist.

What makes it bite rather than just fail: `didwebvh-rs` checks monotonicity only at **resolve** time (`log_entry/read.rs:378`), never at write time. `create_log_entry` signs and appends whatever timestamp it is handed. So the bad entry is signed, appended and published to the host before anything complains, and because the log is append-only, no later update can repair it. An operator upgrading an existing Nitro VTA runs `pnm services didcomm enable`, is told it succeeded, and permanently loses mediator connection, DIDComm and backup export.

## What the follow-up commit changes

`backdated_version_time(n)` becomes `next_version_time(existing_entry_count, previous)`, taking the previous entry's timestamp and clamping to `previous + 1s` whenever the backdated value would not be strictly later:

- **Chains whose genesis predates this fix recover on their next update** instead of bricking on it.
- **A backwards clock step larger than the one-minute index spacing** no longer produces a non-increasing entry. Relevant here specifically: a Nitro enclave takes its clock from the parent at boot, so an NTP correction afterwards is a real path to the same broken chain on a freshly provisioned VTA.
- **The index offset is capped at 1439 minutes.** Uncapped, `now - 1 day + N minutes` overtakes `now` at N = 1440 and the entry is rejected as future-dated — the same unresolvable DID reached from the other side, and just as silent at write time. Past the cap, strict increase is carried by the clamp rather than the index.

Comparisons are made on whole seconds, which is the only precision that survives serialisation. Worth flagging because I got this wrong first time: comparing at full precision let two saturated entries a few microseconds apart pass the clamp and then serialise to the identical timestamp — the exact collision the helper exists to prevent. The saturation test caught it.

The update path in `operations/did_webvh/update/orchestrator.rs` now snapshots `last_state.log_entry.get_version_time()` alongside the prior version-id and document it already keeps, and passes it through.

## Regression coverage

The helper test in the original diff is the pre-existing one relocated from `vta-service` — it passed before this PR and would keep passing if the new `.version_time(...)` call were deleted again, because the defect was a missing call at a builder site, not a wrong helper.

So the TEE genesis builder is extracted as `build_genesis_create_config` (testable without a seed store, KMS or enclave), and two tests now assert that the config it produces carries a backdated `versionTime`, that `create_did` stamps that value onto the entry, and that a first update after such a genesis is strictly later. Both fail if the `.version_time(...)` call is removed; each of the five helper tests fails with its own fix reverted. I checked that rather than assuming it.

## Verification

Run locally against the same command set CI uses, all clean:

```
cargo test --workspace --exclude vtc-service --exclude vtc-client --no-fail-fast
cargo clippy --workspace --all-targets -- -D warnings
cargo fmt --all -- --check
```

All 18 CI checks are green on the branch.

## One thing left

The PR title is now narrower than the change — squash-merge makes it the changelog entry, and this isn't TEE-scoped any more. Suggest `fix(webvh): clamp versionTime against the previous log entry`, but it's your PR, so I'll leave that to you.

marat-affinidi (approved):

PR #1480, opened 2026-09-15
fix: route backup post_trust_task through the shared signed builder

anusha-kommu:
- The backup descriptor flow hand-built a bare { id, type, issuedAt, payload } envelope with no recipient and no proof which VTA rejects as malformedRequest. 
- Delegate to the shared dispatch_trust_task path so the envelope carries recipient + issuer + proof from the client's ClientIdentity, matching every other trust-task surface.


marat-affinidi (approved):

stormer78:
Reviewed. **The change itself is correct** — I verified the things that could have gone wrong and none of them did:

- The transport gate is semantically identical to the one it replaced (REST-only), and the match stays exhaustive with both features off.
- `dispatch_trust_task`'s REST arm posts to the same `{base}/trust-tasks`, and `send_authed` subsumes the removed `ensure_token_valid` + `with_auth_token`. Its 401/403 replay resends the *same signed document* (same `id`) after a request that never executed, so it cannot double-mint a bundle.
- No success response is misclassified: `FinalizeImportResultBody` has `message` but no `code`, so `trust_task_error` will not misread it.
- All five backup bodies conform to the published `additionalProperties: false` schemas, so the newly-reachable `check_payload_conforms` does not break the happy path.
- `verify_reply`'s signer check is satisfiable — `finalize_import` does not hot-swap the signing key.

## Blocking: CI has not run on this PR

`DCO` is the only check that has executed. Both workflows are parked at `action_required` (fork PR), zero jobs, `0s`:

```
34939254109  action_required  PR title  0s
34939254238  action_required  CI        0s
```

Someone needs to click **"Approve and run workflows"** before this merges. Per `CLAUDE.md`, DCO is the only *required* check on `main`, so a red suite merges exactly as readily as a green one — this is the case that section was written for.

I ran CI's jobs locally against `ea4abeb0` in the meantime. Two pass, one fails:

| Job | Command | Result |
|---|---|---|
| Test | `cargo test --workspace --exclude vtc-service --exclude vtc-client --no-fail-fast` | 118 targets, **4255 passed, 0 failed** |
| Clippy | `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| Format | `cargo fmt --all -- --check` | **FAILS** |

### 1. `cargo fmt --all -- --check` is red (would fail the Format job)

Two diffs, both in code this PR adds — `backup_descriptors.rs:228` (the new `dispatch_trust_task` call) and `:322` (the new test module's import). Same `stable` toolchain CI installs.

### 2. The test helper's doc claim is inaccurate, and the tests step over the new gate

> `signed_task_document` is exactly what `post_trust_task` now delegates to, so what it produces is what goes on the wire.

It delegates to `dispatch_trust_task`, which runs `check_payload_conforms` **first**. That gate is the one new client-side failure mode this change introduces, and `assert_conforming_request` skips it — so an `InitiateExportBody` with a 10-character password passes the test and fails at runtime.

Patch for both, verified (`fmt` clean, both tests still pass under `--features client`, which is how the workspace job reaches them via feature unification):

<details>
<summary><code>git apply</code>-able diff</summary>

```diff
diff --git a/vta-sdk/src/client/backup_descriptors.rs b/vta-sdk/src/client/backup_descriptors.rs
index e1dd177c..e665d822 100644
--- a/vta-sdk/src/client/backup_descriptors.rs
+++ b/vta-sdk/src/client/backup_descriptors.rs
@@ -228,7 +228,9 @@ impl VtaClient {
         // REST arm (it paces only the DIDComm/TSP wait), so its value is
         // immaterial here.
         let payload_value = serde_json::to_value(&payload)?;
-        let response_payload = self.dispatch_trust_task(type_uri, payload_value, 30).await?;
+        let response_payload = self
+            .dispatch_trust_task(type_uri, payload_value, 30)
+            .await?;
         Ok(serde_json::from_value(response_payload)?)
     }
 
@@ -322,7 +324,9 @@ fn sha256_hex(bytes: &[u8]) -> String {
 #[cfg(test)]
 mod descriptor_envelope_tests {
     use super::super::{ClientIdentity, VtaClient};
-    use crate::protocols::backup_management::descriptors::{InitiateExportBody, InitiateImportBody};
+    use crate::protocols::backup_management::descriptors::{
+        InitiateExportBody, InitiateImportBody,
+    };
     use crate::trust_tasks;
 
     const VTA_DID: &str = "did:key:z6MkVtaBackupTarget";
@@ -348,13 +352,23 @@ mod descriptor_envelope_tests {
 
     /// Build the document the descriptor flow would emit for `type_uri` and
     /// assert it is addressed to the VTA, issued by the caller, and carries a
-    /// cryptographically valid proof by the caller's key. `signed_task_document`
-    /// is exactly what `post_trust_task` now delegates to, so what it produces
-    /// is what goes on the wire.
+    /// cryptographically valid proof by the caller's key.
+    ///
+    /// This walks the two steps `post_trust_task`'s delegate
+    /// [`VtaClient::dispatch_trust_task`] takes before the transport, in the
+    /// order it takes them: the payload-conformance gate, then
+    /// `signed_task_document`. Running the gate here is what keeps the helper
+    /// honest about being "what goes on the wire" — it is the one client-side
+    /// refusal the delegation introduced, and a test that only built the
+    /// envelope would pass on a payload the real call rejects before it ever
+    /// builds one.
     async fn assert_conforming_request(type_uri: &'static str, payload: serde_json::Value) {
         let id = caller_identity();
         let client = VtaClient::new("http://vta.invalid").with_identity(id.clone());
 
+        VtaClient::check_payload_conforms(type_uri, &payload)
+            .unwrap_or_else(|e| panic!("{type_uri}: the payload does not conform: {e}"));
+
         let doc = client
             .signed_task_document(type_uri, payload)
             .await
```
</details>

## Not blocking, fixed separately

The schema gate makes `finalize-import/1.0`'s `minLength: 15` on `password` the first thing a short restore password hits, and no `backup import` prompt validated locally — so an operator would read a schema-conformance error naming a task URI instead of "wrong password". That gap predates this PR (the spine already validated payloads, so the rejection is not new), so I fixed it in #1484 rather than expanding scope here.

PR #1490, opened 2026-09-15
fix(enclave-proxy): re-lock affinidi-crypto version skew + CI coverage

marat-affinidi:
affinidi-did-resolver-cache-sdk 0.8.37 pulls didwebvh-rs 0.7.0, whose own manifest mixes affinidi-crypto 0.1.5 (via affinidi-secrets-resolver 0.5.5) and 0.2.6 (via affinidi-did-common 0.4.2) — E0308 in did_key.rs. Bumps affinidi-secrets-resolver to 0.5.11 and affinidi-did-resolver-traits to 0.1.3 (both already published, both within existing semver floors) to collapse onto one affinidi-crypto major. Lockfile-only.

Also adds a CI job for this standalone crate (excluded from the root workspace), since no existing job builds it — which is why this sat on main undetected until a downstream pipeline hit it.

Verified: `cargo build --release --locked` and the downstream Docker build both pass locally.


stormer78:
Reviewed, and pushed four commits addressing what it turned up. The lockfile change itself is correct and tightly scoped — I diffed the full package-version sets against `017e773d` and only the five `affinidi-*` / dalek families move, exactly as described; `cargo metadata --locked` resolves clean, and the `windows-sys` churn in the diff is cosmetic (all three versions were already in the lock before and after).

Everything below is in the new CI job, plus one thing that job immediately found.

### `13da383a` — two live advisories in the standalone lock

Adding cargo-deny coverage (below) meant running it on this crate for the first time ever, and it was not clean:

| | | |
|---|---|---|
| `h2` | 0.4.14 → 0.4.19 | **RUSTSEC-2026-0258** — unbounded empty DATA frames; hyper queues them without limit, so an un-drained stream is unbounded memory or a length-overflow panic |
| `crossbeam-epoch` | 0.9.18 → 0.9.21 | **RUSTSEC-2026-0204** — invalid pointer dereference in `fmt::Pointer` |
| `chacha20` | 0.10.0 → 0.10.2 | yanked |
| `spin` | 0.9.8 → 0.9.9 | yanked |

All four were already patched in the root workspace's lock — this crate was just never scanned. `h2` is the one that matters: this binary terminates TLS on the parent instance and reaches `h2` through `reqwest` on every DID resolution. Exactly those four packages move; the other 137 stay put and the affinidi-crypto collapse is untouched.

### `00e41a0c` — CI corrections

**The cache key collided with the `enclave` job.** restore-keys are *prefix* matches, so `Linux-cargo-enclave-proxy-<hash>` is a hit for the `enclave` job's `Linux-cargo-enclave-`. That job keys on `hashFiles('**/Cargo.lock')`, which now includes this crate's lock, so it falls through to its restore-key routinely — and GitHub returns the most recently created match, frequently this job's ~400-package registry instead of the root workspace's, after which `enclave` silently re-downloads the difference. Same shape as #891, which is documented at the top of the file. Renamed the stem to `encproxy`.

**`cargo test --locked` compiled the whole graph a second time.** `target/` isn't cached, so the dev-profile test build shared nothing with the release build above it — 399 packages twice, `aws-lc-sys` (cmake/C, the slowest item) included, inside `timeout-minutes: 20` on a cold runner. Every step is `--release` now, which is also the profile the downstream image ships. clippy builds *dependencies* with plain rustc, so it warms the cache for the steps after it rather than competing with them.

**Widened the coverage** the job's own comment identifies: added `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and a second cargo-deny step against this manifest. `--config deny.toml` is required there — cargo-deny looks for a config beside the manifest, finds none (own `[workspace]`), and would otherwise fall back to defaults enforcing none of the justified ignores.

**Test step is `--test-threads=1`.** `detect.rs` mutates `PATH` / `NITRO_CLI` with `unsafe { set_var }` while 22 sibling tests run; a `setenv` racing another thread's `getenv` — libc's own included — is the unsoundness edition 2024 made that call `unsafe` for. 23 tests, so serialising costs nothing.

### `3ac606e4`, `c200f4de` — making those two checks pass

Neither had ever run, so both were red. Mechanical, but they're the bulk of the line count, hence separate commits:

- **rustfmt**: `cargo fmt` output only, no edits of mine.
- **clippy** (21 lines): three `collapsible_if` → let-chains; `ptr_arg` (`write_did_log_file` took `&PathBuf`, so its own `unwrap_or(data_dir.as_path())` existed only to undo the borrow); `type_complexity` → a `KvPair` alias; `trim_split_whitespace` (`split_whitespace` already skips what the `trim()` removed, the CONNECT line's CRLF included). No behaviour change in any of them.

Ran the full job locally, all green: fmt ✓ · clippy `-D warnings` ✓ · `build --release --locked` ✓ · 23 tests ✓ · `deny check advisories bans sources licenses` ✓.

Happy to drop the two mechanical commits if you'd rather keep this PR to the lockfile — the `fmt`/`clippy` steps would need to come out with them.

PR #1515, opened 2026-09-16
Honour configured DID cache when verifying Trust-Task replies

marat-affinidi:


stormer78:
Closing in favour of #1517, which now includes this fix (commit 7a7ae675). #1517 moves every vta-sdk resolver onto one shared, cached resolver per process, including the reply-verification resolver this PR changes. Both PRs edited the same lines and disagreed there: #1517 had kept reply verification in local mode. It now uses `shared_did_resolver_from_env()`, so reply verification honours `PNM_RESOLVER_URL` as this PR intended, and also reuses the process-wide cache. Thanks for catching it.

PR #1562, opened 2026-09-18
fix(rate-limit)!: key the per-IP limiter on trusted-proxy CIDRs, not a global XFF flag

marat-affinidi:
Replaces the boolean `trust_xff` flag with `trust_xff_cidrs: Vec<CIDR>` (VTA + VTC). The per-IP rate limiter now reads `X-Forwarded-For` only when the request's peer address falls inside an explicit trusted-proxy CIDR allowlist, keying on the rightmost entry.

Fixes two issues with the old flag: an untrusted peer could forge a leading XFF entry to evade its own limit, and every request behind a trusted proxy shared one bucket — one client's burst could 429 unrelated clients.

Breaking config change: replace `trust_xff = true/false` with `trust_xff_cidrs = ["<cidr>", ...]`.

## Test plan
- [x] `cargo test --workspace --exclude vtc-service --exclude vtc-client --no-fail-fast` (4600 passed)
- [x] `cargo test -p vtc-service -p vtc-client --no-fail-fast` (63/63 passed)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] Local RED/GREEN gate-test harness against real Docker containers: spoof-rejection, direct-bypass, per-client attribution, and backward-compat all pass on the patched build; attribution correctly fails on unpatched `main`


stormer78:
Reviewed this in a clean worktree off `origin/main` and reproduced the test plan — everything it claims is accurate (workspace 4610 passed, vtc 1677 passed, clippy and fmt clean, `cargo deny` ok on both manifests).

The core change is right and the unit tests cover the right cases, including forged XFF from an untrusted peer. Refusing the retired key rather than ignoring it is also the right call given `vta-config` only *warns* on unknown keys — a silent ignore would have reverted the fix on every upgraded deployment.

Two things block it as written, and I have a fix ready for both plus the smaller points.

## 1. `trust_xff_cidrs = ["127.0.0.1/32"]` in the Nitro configs is a live bypass

The whole Nitro inbound path is a layer-4 byte bridge with no HTTP awareness:

- parent: `socat TCP-LISTEN:8443,fork VSOCK-CONNECT:<cid>:5100` (`deploy/nitro/parent-proxy.sh:196`), or the Rust `enclave-proxy`'s `run_inbound` → `copy_bidirectional` (`deploy/nitro/enclave-proxy/src/channels.rs:22`)
- enclave: `socat VSOCK-LISTEN:5100,fork TCP-CONNECT:127.0.0.1:8100` (`deploy/nitro/enclave-entrypoint.sh:76`)

Nothing there sets, rewrites or strips `X-Forwarded-For`. So the VTA's socket peer is always `127.0.0.1` — which this PR adds to the trusted list — and the client's own header arrives verbatim. `TrustedProxyKeyExtractor::extract` then keys on the attacker-chosen value. Rotate the header per request and `/auth/challenge`, `/bootstrap/request` and `did.jsonl` are unmetered. The PR's own `trusted_peer_keys_on_rightmost_xff` test is exactly this behaviour.

Before this PR `trust_xff` defaulted `false`, so the enclave was coarse (one shared bucket) but not spoofable — so this is a regression, and `deploy/nitro/config.toml` is committed to PCR0 under `BAKE_CONFIG=true`, so it ships in every enclave image.

## 2. The docs still document the retired key, and following them now bricks the VTA

`docs/02-vta/rate-limiting.md:43-59` still describes `trust_xff = true/false`, and line 97 has `# trust_xff = true` in the sample config. Uncommenting that is now a hard config-load failure. Same stale references at `docs/README.md:98` and `docs/02-vta/non-interactive-setup.md:173`.

## Smaller points

3. **A test-only shim moved into the VTA's production router.** `routes/mod.rs:259,301,559` layer `insert_default_connect_info_if_missing` onto the unauth, did-log and backup-blob branches. That retires the guard `no_attributable_client_is_refused_not_unlimited` documents — an unattributable request used to be refused, now it's attributed to `127.0.0.1`, which finding 1 also makes *trusted*. The test still passes only because it calls `apply()` directly and so misses the layer the real router carries.

4. **Rightmost-only is wrong for a chained proxy, which is what a CIDR *list* implies.** With one hop the rightmost entry is the client (nginx `$proxy_add_x_forwarded_for` and ALB both append the peer) — correct. With `ALB → nginx → VTA` the rightmost is the ALB's address, so every client collapses into one bucket again, one hop out from the bug being fixed. Fails closed, so not a bypass — but the config already takes a list, so it may as well mean what it says.

5. **The Dockerfile changes are unrelated scope and buy nothing on CI.** `.github/workflows/docker.yml:85-90` deliberately configures no build cache, with a comment about the 10 GB Actions-cache eviction incident, so BuildKit cache mounts are per-job ephemeral there. `Dockerfile.nitro` (the other matrix entry) didn't get the same treatment, so the two now diverge. Worth splitting out so a revert of the security fix doesn't drag the build change with it.

6. Nits: `vtc-service/src/routes/mod.rs:1080` has a vestigial `let _ = trust_xff_cidrs;` now the value is used; and at `:1183` the branch picks between `PeerIpKeyExtractor` and `TrustedProxyKeyExtractor`, which already coincide on an empty list (your `empty_cidr_list_never_trusts_xff` proves it) — one arm removes the chance of getting them the wrong way round.

## Fix ready

I have four commits addressing 1–4 and 6, on `fix/xff-trust-boundary-at-proxy`. The approach for 1 is to make the claim true rather than withdraw it: `enclave-proxy`'s inbound channel terminates HTTP/1.1, strips the client-supplied identity headers, sets `X-Forwarded-For` from the address it actually accepted, and forwards over vsock. It's the last hop that knows the client's address, so it's the only place this can be done. The other five channels stay byte bridges. `parent-proxy.sh` genuinely can't do this, so it now refuses to start against a config that sets `trust_xff_cidrs`.

hyper on both ends deliberately — it's also the VTA's own server, and a hand-rolled rewriter that disagrees with the upstream parser about request boundaries is how smuggling happens. Bodies stream through unbuffered.

For 4, the extractor walks the chain right-to-left and takes the first entry that isn't itself a declared proxy. Also fixes three robustness gaps found on the way: IPv4-mapped peers (`::ffff:a.b.c.d` from a dual-stack listener never matched an IPv4 CIDR, so a correct proxy would silently stop being trusted), `addr:port` and `[v6]` entry forms, and an empty entry from a trailing comma ending the walk. Drops the `client-ip` dependency, which can only do rightmost-of-last-header.

Same CI set is green on it: 4610 workspace, 1677 vtc, clippy `-D warnings`, fmt, `cargo deny` on both manifests. The forwarding module has end-to-end tests against a hyper upstream that echoes back the header it was handed, including across keep-alive; both fail if the sanitising call is removed.

One caveat: `enclave-proxy` is Linux-only (`tokio-vsock`), so I couldn't build it on macOS. I verified the module in a scratch crate with TCP standing in for vsock — which did catch a real clippy failure — but its CI job will be the first full build. Happy to push the branch here or open it as a PR against `feat/trust-xff-cidrs`, whichever you prefer.

stormer78:
Pushed the four commits to this branch (fast-forward off `e00e8f6f`, nothing of yours rewritten) so they're here to review rather than described in the abstract:

- `2805bffb` `fix(rate-limit): walk the forwarded chain past every declared proxy` — finding 4, plus IPv4-mapped peers, `addr:port`/`[v6]` entry forms, and empty entries. Drops `client-ip`.
- `6f0f5de3` `fix(rate-limit): keep the synthetic-peer shim out of the production routers` — finding 3, and nit 6.
- `989094b1` `fix(nitro): make the parent proxy the trust boundary for X-Forwarded-For` — finding 1.
- `82e6960c` `docs(rate-limit): document trust_xff_cidrs and what naming a proxy asserts` — finding 2.

Drop any of them if you'd rather take a different approach — particularly on finding 1, where making `enclave-proxy` HTTP-aware is a real change in what that binary is, and "remove the key from the Nitro configs and keep peer-IP keying there" is a legitimate smaller answer.

Finding 5 (the Dockerfile BuildKit change) I deliberately left alone — it's yours and it isn't harmful, it just wants its own PR so a revert here doesn't drag it along.

Green locally on the full CI set: 4610 workspace, 1677 vtc, `clippy --workspace --all-targets -- -D warnings`, `fmt --check`, and `cargo deny` on both the root and the `enclave-proxy` manifests.

One caveat worth repeating now it's on the branch: `enclave-proxy` is Linux-only (`tokio-vsock`), so none of it compiled on my machine. I verified `http_forward` in a scratch crate with TCP standing in for vsock — which did catch a real clippy failure — but the `run_inbound` call site has only been read, not built. The `Enclave proxy (standalone crate)` job is the first real build of it, so worth watching that one specifically.

PR #1567, opened 2026-09-18
fix(trust-tasks): let the spine decide what an inbound document is

stormer78:
## The defect

Authorization happened in **each transport**, before the spine saw the document. That defeated the ordering the spine already documents and relies on (`dispatch_trust_task_inner`):

> *"Checked here, **before any authorization or dispatch**, because that is what it is: a reply carries no authority and asks for nothing, and running it through the request pipeline would at best refuse it and at worst execute it."*

The intent was right. It just never got a chance to apply, because the transport had already refused the document at the door.

Under DIDComm the transport's own `thid` correlation hid it. Under TSP — whose binding has no request/response of its own — a **DID hosting server's answer to a task the VTA had itself sent** came back and was ACL-refused as though unsolicited:

```
refusing trust task: DID not in ACL: did:webvh:QmUcyd…
  type_uri=.../did-management/did/problem-report/0.1     → 403
```

From the outside that reads as a missing ACL entry. **It isn't** — and granting one would hand a peer standing to send *requests* when all it ever needed was to answer.

A second live failure was the same missing idea: an error document dispatched as a request, refused, and answered with another error — which the peer then did too. One failure became a permanent exchange that stopped only when the mediator began rate-limiting.

## The model

A transport now hands over two things — **the document, and the VID it proved** — and makes no policy decision at all. `accept_from_proven_sender` asks what the document *is* and only then decides what may be done with it:

| kind | what happens |
|---|---|
| **error** | terminal: logged *with its reason*, answered with nothing |
| **threaded, and someone is waiting** | delivered to its waiter; never authorized, never answered |
| **everything else** | reaches the ACL as a request |

Every intrinsic-sender transport gets that at once — the property a per-transport copy kept failing to have. Both TSP and DIDComm shrink; REST keeps bearer auth and cannot receive replies anyway.

## Threading deliberately does not make a reply

This is the subtle part, and getting it wrong would have been worse than the bug. A step-up `approve-response` and a `task-consent/decision` **thread to the request that provoked them and are still requests** — they carry the approval. Treating threading as proof of reply-ness would have silently stranded every ceremony waiting on a human.

So the *waiter* decides: if one holds that thread, the document is its answer; if not, it falls through to the normal pipeline. `a_threaded_document_with_no_waiter_is_still_dispatched` pins it.

## Why the classification is in `vta-sdk`

It is a fact about the *document*, so it is the same fact for every peer that reads one — the VTA, a DID hosting service, anyone else. Put it in one end and the other end's copy drifts; the TSP binding taught that already. Same reasoning, same home as `tsp_binding` and `budget`.

## Also fixed

An empty reply was being sealed as `{"type":…,"document":}` — a malformed frame in place of silence. Only reachable now that answering nothing is a real outcome; caught by the new tests.

## Testing

12 TSP-inbound tests (3 new) + 5 classifier tests. The two that would have failed before: `a_reply_reaches_its_waiter_without_the_sender_needing_acl_standing` and `an_inbound_error_is_terminal_and_is_never_answered`.

`cargo fmt --all --check` ✅ · `cargo clippy --workspace --all-targets` ✅ **0 issues** · targeted suites ✅ — full `cargo test --workspace` was still running when this was opened; I'll confirm it on the PR.

## Context

Part of a chain found while diagnosing a persona DID mint that failed with nothing but a timeout. affinidi-webvh-service#201 (binding envelope) and #202 (the peer half of the loop) are merged; #1566 fixes the timeout budget that hid the diagnosis.


stormer78:
Full-sweep confirmation, as promised in the description:

`cargo test --workspace` — one failure, `vtc-service --test trust_task_manifest::every_bound_canonical_task_exists_in_the_registry`. **Pre-existing on `origin/main`**: it fails identically with this branch's changes reverted, and `vtc-service` is not touched here. Everything else passes.

`cargo clippy --workspace --all-targets` — 0 issues.

stormer78:
**Correcting my earlier comment on this PR — that failure was mine, not pre-existing.**

I claimed `every_bound_canonical_task_exists_in_the_registry` failed on `origin/main` too. My baseline check was invalid: I used `git checkout origin/main -- .`, which restores tracked files but does not remove a *new* one, so `vta-sdk/src/inbound.rs` stayed in the tree and the test still saw it. Checking out the commit properly, `origin/main` passes 8/8.

**Cause:** the test scans the workspace for `trusttasks.org/spec/` strings ending in a `MAJOR.MINOR` segment and asserts the registry publishes each one. My `a_future_error_version_is_still_an_error` test spelled out a made-up `trust-task-error/9.9`, which read as a real binding on an authority nobody serves — taking that family from 1 unpublished URI to 2.

**Fix:** compose the URI from `TRUST_TASK_ERROR_PREFIX` instead of writing it out. The case is worth keeping (a receiver that only knew `0.5` would start answering the next version — the loop again), it just must not claim a spec that does not exist. The prefix constant itself was never implicated: it ends in a slash, so the scan's `MAJOR.MINOR` filter already excluded it.

Now 8/8, matching `origin/main`.

PR #1581, opened 2026-09-20
fix(resolver): take the shared DID resolver instead of building one

stormer78:
Closes KR-19 (their **VTI-19**) — one command fetching `/.well-known/did.jsonl` forty times in five minutes, earning a 429 from the DID host's own rate limiter.

## Two problems, not one

**Twelve call sites each constructed their own `DIDCacheClient`.** A client owns its cache, so the same DID was fetched once per *construction* rather than once per process. `vta-service/src/webvh_cli.rs` alone had five.

**And by calling `DIDCacheClient::new` directly, none of them read `PNM_RESOLVER_URL`.** So an operator who had configured a resolver sidecar — the documented mitigation for exactly this load — did not get it on any of these paths. That env var exists precisely so a single setting reaches every resolver construction without breaking the SDK's public API (see the module doc on `vta-sdk/src/resolver.rs`), and these sites went around it.

The second is the more consequential of the two: the first multiplies fetches, the second removes the thing an operator would reach for to absorb them.

## Why the conversion is behaviour-preserving

Every converted site built

```rust
DIDCacheConfigBuilder::default()
    .with_host_policy(vta_sdk::resolver::webvh_host_policy())
    .build()
```

which is byte-for-byte what `build_did_cache_config(None)` produces. So `shared_did_resolver_from_env()` is a drop-in that additionally honours the env var, and the twelve caches collapse into the one the SDK already keeps per (runtime, sidecar URL, host policy).

## Three sites deliberately left alone

Each looks convertible, and none is:

- **`vtc-service/src/server.rs`** and **`room-host/src/main.rs`** build a plain default with **no host policy**. Converting them would add `webvh_host_policy()` and change which hosts are permitted — a security-relevant change, not a caching one.
- **`pnm-cli/src/bootstrap.rs`** passes `None` **on purpose**. `bootstrap connect` resolves the DID locally so the operator verifies the SCID and signed log on their own machine rather than trusting a sidecar; CLAUDE.md records this explicitly. That `None` is the trust boundary, not an oversight. It is the single most convertible-looking construction in the tree.

This is why each site's config was read rather than the constructor pattern-matched.

## The test that was missing

All of the above now rests on `shared_did_resolver` actually sharing, and nothing pinned that. If sharing broke, every converted site would quietly revert to a private cache with no test failing — the same shape as the bug being fixed.

`two_callers_on_one_runtime_share_one_resolver` asserts on the registry rather than on cache hits: counting fetches against a loopback DID would need the private-host policy and make the test as much about host policy as about sharing.

It is feature-gated (`resolver` sits behind `session`/`didcomm`/`tsp`/…), so I checked it runs in CI's workspace job rather than assuming: it does, via feature unification — 876 tests compile under `--workspace` versus 380 for `-p vta-sdk` alone, which is the discrepancy CLAUDE.md warns about.

## Verification

- `cargo test --workspace --exclude vtc-service --exclude vtc-client --no-fail-fast` — 399 passed, 0 failed
- `cargo test -p vtc-service -p vtc-client --no-fail-fast` — 1110 passed, 0 failed
- `cargo clippy --workspace --all-targets -- -D warnings` — clean

No version edits: release-plz owns those here, and a red semver report on this PR would be the report working.


stormer78:
Fixed in 92f671a7 — **the test was wrong, not the code.**

`two_callers_on_one_runtime_share_one_resolver` asserted `after == before + 1` over the whole registry. That registry is process-global and every other `#[tokio::test]` in the binary registers under its own runtime key, so a total is noisy — and, the part that actually broke it, **not monotonic**. `shared_did_resolver` calls `prune_dead` *before* it inserts, so one call can remove several entries and add one. CI hit exactly that: a dead entry pruned in the same call that registered ours, leaving the count unchanged and the assertion reading `left: 1, right: 2`.

It now counts only entries keyed to the current runtime, which is immune to what any other test registered or left behind, and is the property the test is actually about: this runtime asks twice and gets one resolver.

The `0 → 1 → still 1` sequence is what stops it passing vacuously — if registration were broken the count would stay 0, and if sharing were broken it would reach 2.

One thing I tried and backed out: proving the counter can move by registering a second key with a different sidecar URL. That puts the client in network mode and made the test spend **five seconds** waiting on a socket that isn't the subject. The 0 → 1 step already proves it. Back to 0.02s.

**Verified against the failure mode rather than in isolation**, since a filtered run wouldn't reproduce it: `cargo test -p vta-sdk --all-features --lib` puts 948 tests in one process — a superset of the 874 CI ran — and it passes there. `clippy --all-features --all-targets -- -D warnings` and `fmt --check` clean.

stormer78:
`Test (workspace)` is red on a **pre-existing flake**, not on this branch. Evidence, since I could not assume it:

The failing test is `test_support::transport_harness_tests::d6_recovers_a_reply_from_an_answering_peer` (`test_support.rs:2868`) — not the test this PR touches, which passes.

- **main run [35465791169](https://github.com/OpenVTC/verifiable-trust-infrastructure/actions/runs/35465791169)** (commit `73509892`, 19 Sep 19:54, before this branch existed): **failed on this exact test at this exact line.**
- **main run 35466562292** (commit `89ebd895`, 19 Sep 20:10 — my base commit): green.

Same test, opposite outcomes on `main` sixteen minutes apart. It is a five-second TSP round trip, so it is timing-sensitive by construction.

What I checked before concluding that, because "probably flaky" is the easy answer and main *was* green on my base commit:

- Ran the full harness locally on this branch: **9/9 pass**, including `d6_recovers_a_reply_from_an_answering_peer`.
- Re-ran the failed CI jobs with no code change: `Feature combos` went green, `Test (workspace)` failed again on the same test.
- Checked whether the mechanism could be mine. The harness builds a dedicated `DIDCacheClient` and preseeds mock DID documents into it (`test_support.rs:1288`), so a path switched to the process-shared resolver would resolve against one with no preseeded documents — a real way this change could break exactly this test. It is not what happened: the only converted site outside a CLI entry point is in `did_webvh.rs::run_create_did_webvh`, which the recovery path does not reach. The harness's own resolver still goes into `AppState.did_resolver` untouched.

So: not this branch. Worth its own issue though — it is failing in `Test (workspace)`, which this repo's CI comments call "the gate that matters", and a gate that goes red on its own schedule teaches people to merge past it.

stormer78:
Rebased onto `main` now that #1583 has merged, which fixes the `d6_recovers_a_reply_from_an_answering_peer` flake this PR kept hitting. Clean rebase, no conflicts — the flake fix is in `test_support.rs` and this branch touches resolver construction, so they never overlapped.

Expecting `Test (workspace)` green on this run. The `semver report` red is this repo's normal behaviour and never blocks.

PR #1590, opened 2026-09-20
fix(vtc): return the endorsement-type delete body the OpenAPI document declares

stormer78:
**`Test (VTC)` is red on `main`.** This is the one-handler fix.

## What broke

#1587 replaced the hand-written `{ typeUri }` with the generated
`vtc/endorsement-types/delete/0.1` response and documented the route with
`EndorsementTypeDelete01Response` — the `vta_sdk::openapi` wrapper that renders
the specification's own schema, because `utoipa::ToSchema` cannot be derived on
a foreign type. The handler kept returning the generated type *inside* that
wrapper, so `openapi_response_census` failed:

```
endorsement_types.rs::delete
    declares: EndorsementTypeDelete01Response
    returns:  Json<DeleteTaskResponse>
```

Nothing was wrong on the wire. The wrapper is `#[serde(transparent)]`, so the
bytes are identical either way — which is the whole reason that census compares
*names*. `admin-ui/openapi.json` generates `admin-ui/src/lib/wire.ts`, so an
annotation that drifts from its handler ships a console typed against the wrong
shape and nothing says so: the generated TypeScript typechecks against itself.

## The fix

The handler builds the generated type — the builder's `try_into` is what checks
the value against the schema — and returns the wrapper. Hoisted into a `let`,
because `.try_into()` inside `From::from(...)` has nothing to infer from.

The explanation sits in a plain comment, not the doc comment: utoipa copies a
handler's doc comment into the API document, so putting it there would have
changed `openapi.json` (and required regenerating `wire.ts`) in order to explain
one of our own censuses to API consumers. **The API document is unchanged** —
the diff is one file.

## Verification

- `cargo test -p vtc-service -p vtc-client` (the `Test (VTC)` job's command) —
  63 targets, 0 failures
- `cargo test -p vtc-service --test openapi_response_census` — 6 pass
- `cargo test -p vtc-service --test openapi_snapshot` — 4 pass, document
  unchanged
- `cargo clippy -p vtc-service --all-targets`, `cargo fmt --all -- --check` —
  clean

## Why this sat unnoticed

`cargo test` stops at the first failing target. Until #1587 landed, the first
failure in the tree was the generated-wire-types census — in the *other* job —
so this one never got to run. Worth remembering that the two VTI test jobs
exclude each other (`Test (workspace)` excludes `vtc-service` and `vtc-client`;
`Test (VTC)` is only those two), so a change that touches shared crates needs
both run locally, not one.


stormer78:
Superseded by #1589, which landed the same fix — the handler returns `EndorsementTypeDelete01Response` built from the generated value, so the `body =` annotation and the return type agree and `openapi_response_census` passes.

The two differ only in where the explanation sits: #1589 puts it in the handler's doc comment (which utoipa copies into `openapi.json`, hence the regenerated document and `wire.ts`), this one in a plain comment to leave the API document untouched. Not worth churning a merged fix over. Closing.

